PR #1, opened 2017-06-12
Entry point, plugin system, daemon, existing tools

dirk-thomas:
The PR adds some command line tools and their infrastructure. The pieces are structured in separate commits to make it easier to review:

* The entry-point based plugin system, interfaces for commands and verbs, the cli `ros2`
* Some "not advertised" commands to introspect the plugin system: `extension_points`, `extensions`
* Using `argcomplete` for completion in `bash`
* The interface to use `rclpy` nodes (either directly or via a daemon using xml-rpc for communicating)
* Linter tests for copyrights and `flake8`
* Add the command `ros2 pkg` with the verbs `list` and `prefix`
* Add the command `ros2 node` with the verb `list` (similar to previous prototypes)
* Add the command `ros2 topic` with the verbs `echo`, `list`, and `pub` (similar to previous prototypes)

The CI builds are not really saying much (beside that the linter tests for the new code are passing):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2709/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/251/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2166/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2837/)

Since this covers all the features provided in https://github.com/ros2/cli_tools I suggest to delete that repo after this PR has been merged and is being referenced in the `ros2.repos` file.


dirk-thomas:
Ready for review.

ros2cli/ros2cli/cli.py
```diff
@@ -0,0 +1,69 @@
+# Copyright 2016-2017 Dirk Thomas
```
dirk-thomas: Because the code was written by me in my spare time for a different project.

ros2cli/ros2cli/cli.py
```diff
@@ -0,0 +1,69 @@
...
+import argparse
+import signal
+
+from ros2cli.command import add_subparsers
+from ros2cli.command import get_command_extensions
+
+
+def main(*, script_name='ros2', argv=None, description=None, extension=None):
+    if description is None:
+        description = 'ros2 is an extensible command-line tool for ROS 2.'
```
dirk-thomas: The argument allows other packages to pass a different `script_name` to this function (e.g. if you want an executable called `ros2topic`).

> and not use it here

I am not sure what you mean with this. The variable is used later.

ros2cli/ros2cli/node/daemon.py
```diff
@@ -0,0 +1,66 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import platform
+import socket
+import subprocess
+
+from ros2cli.daemon import PORT
+
+import xmlrpc.client
```
dirk-thomas: Using `XML-RPC` gets us going quickly since it is very easy to use in Python.

In the future it might be replaced with any other protocol. E.g. even with DDS once we are able to connect to a service without the overhead of the discovery.

ros2cli/ros2cli/node/daemon.py
```diff
@@ -0,0 +1,66 @@
...
+        if e.errno in [98, 10048]:
+            return True
+    else:
+        s.close()
+    return False
+
+
+def spawn_daemon(args):
+    subprocess.Popen(
+        ['_ros2_daemon'], stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
```
dirk-thomas: From my understanding a daemon was the idea to speed up the response time.

ros2cli/test/test_copyright.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2015 Open Source Robotics Foundation, Inc.
```
dirk-thomas: Updated.

ros2node/ros2node/api/__init__.py
```diff
@@ -0,0 +1,41 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from ros2cli.node.strategy import NodeStrategy
+
+# TODO(mikaelarguedas) revisit this once it's specified
+HIDDEN_NODE_PREFIX = '_'
```
dirk-thomas: Both usages in this patch are kind of unrelated. Therefore I didn't introduce a dependency for this. This constant should be defined in a lower level package like `rclpy` (hence the TODO).

ros2node/ros2node/command/node.py
```diff
@@ -0,0 +1,35 @@
...
+    def add_arguments(self, parser, cli_name):
+        # get verb extensions and let them add their arguments
+        verb_extensions = get_verb_extensions('ros2node.verb')
+        add_subparsers(parser, cli_name, '_verb', verb_extensions)
+
+    def main(self, *, parser, args):
+        try:
+            extension = getattr(args, '_verb')
+        except AttributeError:
+            assert False, 'The previous line should have exited'
```
dirk-thomas: Imo yes. It is the case which was explicitly considered during development which shouldn't happen if the code before works as designed.

ros2node/ros2node/verb/__init__.py
```diff
@@ -0,0 +1,44 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from ros2cli.plugin_system import PLUGIN_SYSTEM_VERSION
+from ros2cli.plugin_system import satisfies_version
+
+
+class VerbExtension(object):
```
dirk-thomas: Since the class is defined in the namespace `ros2node` I didn't embed that into the class name. Imo the information is already part of the context in which the class is defined in.

ros2node/test/test_copyright.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2015 Open Source Robotics Foundation, Inc.
```
dirk-thomas: Updated.

ros2topic/ros2topic/verb/echo.py
```diff
@@ -0,0 +1,148 @@
...
+def register_yaml_representer():
+    # Register our custom representer for YAML output
+    yaml.add_representer(OrderedDict, represent_ordereddict)
+
+
+# Custom representer for getting clean YAML output that preserves the order in
+# an OrderedDict.
+# Inspired by:
+# http://stackoverflow.com/a/16782282/7169408
+def represent_ordereddict(dumper, data):
```
dirk-thomas: I don't see this function more than once. Can you please point to the duplicate locations.

ros2topic/ros2topic/verb/echo.py
```diff
@@ -0,0 +1,148 @@
...
+    return result
+
+
+# Convert a msg to an OrderedDict. We do this instead of implementing a generic
+# __dict__() method in the msg because we want to preserve order of fields from
+# the .msg file(s).
+def msg_to_ordereddict(msg):
+    d = OrderedDict()
+    # We rely on __slots__ retaining the order of the fields in the .msg file.
+    types = [bool, bytes, dict, float, int, list, str, tuple, OrderedDict]
```
dirk-thomas: I think `OrderedDict` is being used here for recursion. This functionality is again coming 1:1 from the `cli_tools` repo.

ros2topic/ros2topic/api/__init__.py
```diff
@@ -0,0 +1,46 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from ros2cli.node.strategy import NodeStrategy
+
+HIDDEN_TOPIC_PREFIX = '_'
+
+
+def get_topic_names_and_types(*, node, include_hidden_topics=False):
```
dirk-thomas: Since this is public API intended to also be used from other code it ensures that callers pass the arguments by keyword which ensures that if we add additional arguments in the future existing code is guaranteed to work correctly.

ros2topic/ros2topic/verb/__init__.py
```diff
@@ -0,0 +1,44 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from ros2cli.plugin_system import PLUGIN_SYSTEM_VERSION
+from ros2cli.plugin_system import satisfies_version
+
+
+class VerbExtension(object):
```
dirk-thomas: See previous comment about the namespace of the module clarifying the context / name.

ros2topic/ros2topic/verb/pub.py
```diff
@@ -0,0 +1,75 @@
...
+    publisher(args.message_type, args.topic_name, args.values)
+
+
+def publisher(message_type, topic_name, values):
+    try:
+        package_name, message_name = message_type.split('/', 2)
+    except ValueError:
+        raise RuntimeError('The passed message type is invalid')
+    module = importlib.import_module(package_name + '.msg')
+    msg_module = getattr(module, message_name)
```
dirk-thomas: This could indeed be provided by a different Python package. I don't think `rclpy` is a good place for this. Maybe some `rosidl` related package. But since we currently don't have a better place the code exists here. It can be moved somewhere else in the future.

ros2cli/ros2cli/daemon/__init__.py
```diff
@@ -0,0 +1,60 @@
...
+                _print_invoked_function_name(node.get_topic_names_and_types))
+
+            # function to shutdown daemon remotely
+            shutdown = False
+
+            def shutdown_handler():
+                nonlocal shutdown
+                print('Remote shutdown requested')
+                shutdown = True
+            server.register_function(shutdown_handler, 'system.shutdown')
```
dirk-thomas: The rpc `shutdown` is currently not being called. But it allows any client to trigger a shutdown of an already running daemon (for whatever reason).

ros2cli/ros2cli/node/daemon.py
dirk-thomas: Do you imagine the daemon to have a "max-time-to-live"?

ros2cli/ros2cli/cli.py
wjwwood: The description string has `ros2` hard coded rather than using `script_name`.

ros2cli/ros2cli/node/daemon.py
wjwwood: > Do you imagine the daemon to have a "max-time-to-live"?

Do the tools spawn a daemon if one is not available or does the user have to spawn one?

If it is the latter case, then I wouldn't expect this function to be necessary (no need to fork).

If the tool forks one itself, then I would expect a max-time-to-live, like how it works for rti's code gen or bazel's build engine.

ros2node/ros2node/command/node.py
wjwwood: But, from a debugging perspective, now I have to go look at the code to see what is "the previous line". The AttributeError is more precise than the assert isn't it?

ros2node/ros2node/verb/__init__.py
wjwwood: Class names are almost always not used with their modules. For example, I wouldn't be able to import this `VerbExtension` and the `VerbExtension` from `ros2topic` in the same file without aliasing. Seems like a draw back to me.

ros2topic/ros2topic/verb/echo.py
wjwwood: Sorry, I guess I passed over it twice while reviewing. I don't see one.

ros2topic/ros2topic/verb/echo.py
wjwwood: @mikaelarguedas then, do you know? I would have thought that the recursion would only pass a message (submessage) type down?

ros2topic/ros2topic/api/__init__.py
wjwwood: Only new parameters need to be kwargs to avoid any issues. `node` should always be first, even if new parameters are added. There's no reason to make it kwarg like this imo.

This seems like something we should do everywhere or only when required. As it stands it's in contrast to all of our existing code.

ros2topic/ros2topic/verb/pub.py
wjwwood: Please put a TODO then to deduplicate this code.

ros2cli/ros2cli/daemon/__init__.py
wjwwood: So again, when does the daemon shutdown, if it is forked off ("spawned")? Or how does the user shut it down?

ros2cli/ros2cli/node/strategy.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class NodeStrategy(object):
+
+    def __init__(self, args):
+        if is_daemon_running(args):
+            self.node = DaemonNode(args)
+        else:
+            spawn_daemon(args)
+            self.node = DirectNode(args)
```
wjwwood: Why spawn the daemon if you're not going to use it?

ros2cli/ros2cli/node/daemon.py
wjwwood: Also, what happens if I do?:

```
$ ROS_DOMAIN_ID=42 ros2 topic list
$ ROS_DOMAIN_ID=43 ros2 topic list
```

dirk-thomas:
> Also there are some design choices that I wish there had been consensus on before doing this pr. I don't think this should block this pr at all (because of the timeline for the beta), but I think it's a poor approach to development on a team (program first, discuss later). These items could have easily been settled with a design doc, no matter how minimal:

I absolutely agree that writing a design document before would have been the better choice. But the decision to skip the design document for now (because of the goal to integrate these changes for beta 2) was made in the ROS 2 team meeting. So the "poor approach" (to cite your words) have been chosen there - not by me.

> why use xmlrpc for the daemon?

As mentioned inline XML-RPC was chosen for now since it is directly supported in Python and enabled the communication between the command line tool and the daemon with very little code and effort. In the future the protocol can be replaced with any alternative suitable. Even DDS if we have the ability to avoid the overhead of the distributed discovery phase.

> why use long running daemon?
> when does it shutdown?

The goal of the daemon is to provide faster results to queries which require to be part of the DDS network. Hence the daemon needs to be running before the command line script is invoked in order to have these information already. Currently the daemon run indefinitely. We could also make the daemon shutdown automatically, e.g. after a certain period of inactivity (obviously with the downside that another command line invocation after that timeout with be slower again)?

> why not use (or extend) osrf_pycommon's verb pattern (which is used by ament_tools)?

I think the code in this repo is much shorter, easier to read and maintain, and doesn't need some of the "extra" features implemented in `osrf_pycommon`. Therefore I didn't see the benefit of depending on that package. Maybe you can point to any significant code block in `osrf_pycommon` which could be used here which would reduce the code size or complexity in this repo.

> since you're not using osrf_pycommon's verb patter, why extension points?

I am not sure I understand the question. The extension points are being used to enable arbitrary packages to contribute commands and/or verbs to the command line tools.

> they may seem like the obvious solution, but
> they're known to be slow and make issues for tab completion: catkin/catkin_tools#297

I haven't experience any unacceptable slowdown and tab completion seemed snappy enough when I tried it. Since the command line tools are designed to be extensible I chose the default mechanism in Python to realize that: enty points. Alternatives like "fast entry points" can always be considered if necessary and sufficiently mature / supported.

> we could have considered using ament_index, which I think ought to be faster

I don't think that the ament index should be used instead of Python entry points. While the ament index is certainly very valuable for various cases I consider Python entry points to be the "native" way of doing stuff in Python. Replacing a commonly known think with a custom concept imo increases the bar for external users / developers. Also when installing these Python package I would rather avoid to have the need for an amend index (and e.g. an `AMENT_PREFIX_PATH` environment variable).

> why not iterate on the existing cli tools repo and rename it?

The current tools are standalone scripts. I didn't see any benefit (like saving time) by reusing them. I did reuse the code snippets containing the logic though when creating this PR.

> we're loosing the history from there

The repo doesn't have any valuable history imo. The Python scripts have just been created from what was C++ code before (maybe plus a single commit for some parts).

> There seems to be a pattern of starting over, which I think is going to mean more technical debt in the long run, as we've punted the consolidating of the verb pattern further down the road. If you think there are improvements to be made, then that's fine, but imo we should be building on existing implementations where appropriate rather than continuously restarting from scratch just because it's an option.
> For example, you could have iterated on osrf_pycommon.verb_pattern either on a branch or by duplicating the module as osrf_pycommon.verb_pattern_v2 or even by just copying the code into this repo. With the intention of reintegrating in the future, but without incurring the cost of doing so now (having to update ament_tools, et al). Perhaps there is a reason not to start with osrf_pycommon, but again that's lost here because there was no discussion in the first place.

I have reused existing code (like the plugin system, the logic if the cli tools etc.) for exactly that reason. I don't see the value of depending on `osrf_pycommon` though. An regarding `ament_tools` we have already concluded that we don't want to spend time into developing that package any further but at some point in time focus on a build tool which can satisfy all current use cases (ROS 2, ROS 1, non-ROS packages).

> This pattern of starting over is echoed by completely avoiding the existing cli_tools repo, for which I cannot come up with a technical reason and has the consequence of discarding contributions from others on the team unnecessarily. Personally, I really feel that this pr should be applied to that repository, even if the result is just deleting those packages in that repo. At least the contributor stats would have been preserved. But again, I won't make my review of this pr contingent on that.

Imo reuse is beneficial if you gain efficiency from it. Just reusing an existing repo for the sake of reusing it (since you mentioned removing all content before adding the new code).

I am not sure how the contributor stats play in here. Are you worried that users who have committed to `cli_tools` are worried that their contributions "diappear" when the repo is being deleted? If that is the only concern I am happy to make a new PR to `cli_tools` which first wipes the repo clean and then applies the commits from this repo.

ros2cli/ros2cli/node/strategy.py
dirk-thomas: In the case where the daemon is not running yet using the spawned daemon doesn't provide any benefit. Using the direct node is at least as fast as using the just spawned daemon and it allows the user to control e.g. the timeouts using command line arguments.

ros2cli/ros2cli/cli.py
dirk-thomas: Updated to use `script_name` in the description.

ros2cli/ros2cli/node/daemon.py
dirk-thomas: > Do the tools spawn a daemon if one is not available or does the user have to spawn one?

The tools spawn a daemon on demand.

> I would expect a max-time-to-live, like how it works for rti's code gen or bazel's build engine.

This can certainly be implemented. What default timeout of inactivity are you proposing?

wjwwood: > This can certainly be implemented. What default timeout of inactivity are you proposing?

I don't know. I think bazel's long lived daemon runs until something like two hours after no activity.

At the moment the user has no way to shut it down without `kill` (or the task list on windows), or to even know if it is running. So I think it should be visible and easy to shutdown portably or it should have a ttl.

dirk-thomas: > Also, what happens if I do?:

```
$ ROS_DOMAIN_ID=42 ros2 topic list
$ ROS_DOMAIN_ID=43 ros2 topic list
```

The daemon and the command line tools currently don't know anything about `ROS_DOMAIN_ID`. So this use case is currently not covered. I can certainly add this though - different ways are possible:

* E.g. whenever a RPC is being made to the daemon the current domain id is being passed. The daemon only answers the request if the domain id matches. Otherwise it should probably shut itself down.

* The daemon chooses a port number which depends on the domain id. That way multiple daemon would happily coexist when a user uses different domain ids from different shells.

ros2node/ros2node/verb/__init__.py
dirk-thomas: > Seems like a draw back to me.

Let's agree to disagree on this :wink:

ros2topic/ros2topic/api/__init__.py
dirk-thomas: Imo a function call like this:

`func('one', True, 23)`

is much less readable than:

`func(package_name='foo', show_hidden=True, timeout=23)`

ros2topic/ros2topic/verb/pub.py
dirk-thomas: Sure. Done.

ros2node/ros2node/command/node.py
dirk-thomas: I replaced the exception handling with a comment: https://github.com/ros2/ros2cli/pull/1/files#diff-071aae77e7328b857aed121f5f268f5dR29

dirk-thomas:
I made the port of the forked daemon dependent on the `ROS_DOMAIN_ID`: https://github.com/ros2/ros2cli/pull/1/files#diff-0db5b41e1e3495a1cb622aa6f10bdf78R54

I will add a timeout to the daemon to self destruct after a certain amount of inactivity... Suggestions for a default value are welcome.

wjwwood:
> So the "poor approach" (to cite your words) have been chosen there - not by me.

You don't have to do a full design doc to communicate your intentions to the rest of the team. Take my rviz plan I posted to the mailing list. It's not a full design doc, but it hits the high points on the plan for doing the work as well as some key decisions. The purpose of which is to coordinate with others, inform them so they can collaborate if they want to, or raise concerns if they see them.

My understanding of the meeting was to implement something that was the bare minimum for the requirements since we didn't have time to do a design first, not to make key design decisions without any discussion or design doc.

> As mentioned inline XML-RPC was chosen for now since it is directly supported in Python and enabled the communication between the command line tool and the daemon with very little code and effort. In the future the protocol can be replaced with any alternative suitable. Even DDS if we have the ability to avoid the overhead of the distributed discovery phase.

:+1:

> The goal of the daemon is to provide faster results to queries which require to be part of the DDS network. Hence the daemon needs to be running before the command line script is invoked in order to have these information already. Currently the daemon run indefinitely. We could also make the daemon shutdown automatically, e.g. after a certain period of inactivity (obviously with the downside that another command line invocation after that timeout with be slower again)?

That all sounds fine. I don't think it should live for ever, but also it should live long enough to make the issue of having to spin it up again not that common.

> I think the code in this repo is much shorter, easier to read and maintain, and doesn't need some of the "extra" features implemented in osrf_pycommon. Therefore I didn't see the benefit of depending on that package.

Perhaps instead you could list the "extra" features that `osrf_pycommon` has that you don't need and why that matters? The point is that it implements exactly what you've reimplemented here. And since you _were_ starting from scratch, all of the features were useful until you rewrote them here.

I agree that this code is easy to read and has some advantages over the `cli_utils` in `osrf_pycommon`, but these are all improvements that could have been put into `osrf_pycommon` as well. I don't see a reason not to do that.

> Maybe you can point to any significant code block in osrf_pycommon which could be used here which would reduce the code size or complexity in this repo.

Well it handles the registration, listing, adding of arguments to parsers for verbs (all of which is done in this repo and would not need to be). It also provides utilities for sub parsers and for extracting subsets of arguments using `--` as a delimiter.

Your code does more, it has a cleaner interface for registering extensions, and goes beyond verbs to commands. It also integrates with argcomplete, which is awesome. I just wish you would push that into osrf_pycommon so it could be used by catkin_tools, ament_tools, or other python command line tools we might need to make, rather than just plop it here.

Despite the fact that we want to replace catkin_tools and ament_tools with something better, there is no plan for that and I think it's a weak reason to avoid improving osrf_pycommon.

> I am not sure I understand the question. The extension points are being used to enable arbitrary packages to contribute commands and/or verbs to the command line tools.

Sorry I meant entry_points.

> I haven't experience any unacceptable slowdown and tab completion seemed snappy enough when I tried it. Since the command line tools are designed to be extensible I chose the default mechanism in Python to realize that: enty points. Alternatives like "fast entry points" can always be considered if necessary and sufficiently mature / supported.

I believe it is a matter of scale. As you increase the number of entry_points and as the code which each entry_point imports becomes more complex, it gets a lot slower. It's also possible that later versions of Python have since improved the performance.

I'm fine with the decision, but again it's just one of the things that ideally would have been discussed first.

> I don't think that the ament index should be used instead of Python entry points. While the ament index is certainly very valuable for various cases I consider Python entry points to be the "native" way of doing stuff in Python. Replacing a commonly known think with a custom concept imo increases the bar for external users / developers. Also when installing these Python package I would rather avoid to have the need for an amend index (and e.g. an AMENT_PREFIX_PATH environment variable).

That's fine. I agree adding the dependency is unnecessary (even though ament_flake8 is already a dependency).

This is just another thing that should be in the "alternatives" section of a design doc.

> The repo doesn't have any valuable history imo.

> Are you worried that users who have committed to cli_tools are worried that their contributions "diappear" when the repo is being deleted?

Well, perhaps @mikaelarguedas and @Karsten1987 don't mind, but I think it would be the considerate thing to do to try and preserve their commits in the history. They spent time on these things and deserve credit for them.

> I have reused existing code (like the plugin system, the logic if the cli tools etc.) for exactly that reason.

:+1:, but it would be better with commit history to show that.

> I don't see the value of depending on osrf_pycommon though.

The value is in collaborating, and improving things for all of our command line tools, not just this set of tools in this one place. The benefits of collaborating is shared time on documentation and testing (for which `osrf_pycommon` has much more already) and for fixing common bugs (rather than running into the same issues in N places).

> An regarding ament_tools we have already concluded that we don't want to spend time into developing that package any further but at some point in time focus on a build tool which can satisfy all current use cases (ROS 2, ROS 1, non-ROS packages).

That really doesn't have anything to do with it. catkin_tools also uses it, and this other build tool you're talking about should also, ideally, make use of it (with any necessary improvements or changes to make that acceptable). This is all part of trying to arrive a common solutions for similar problems. Instead we're constantly forking away from each other. This is not the ideal collaboration we should be having.

----

For me, the default case should be to try and use what is already there, improving it if necessary. And only with justification should we opt to build something new that does the same thing.

Please @ros2/team agree or disagree with me here, because I think this is a core creative decision that we should all have a consensus on.

wjwwood:
> I will add a timeout to the daemon to self destruct after a certain amount of inactivity... Suggestions for a default value are welcome.

Thanks, two hours?

ros2topic/ros2topic/api/__init__.py
wjwwood: Those are literals, there is no way that `get_topic_names_and_types(node, ...)` is ambiguous, unless the node parameter name is bad, but that's it's own problem.

I could be convinced that this would be a good thing to do for literals, but it is still the case that this code is in direct contradiction to most of our existing code. If this is the "new way forward" it needs to be documented in our developer guide after we discuss it as a team.

ros2cli/ros2cli/node/daemon.py
wjwwood: I was thinking more like, every request had the domain id embedded and the daemon could create a node for each domain id it encounters. And the function that detects if a daemon is available would additionally see that the daemon had a node running for the current domain id already.

wjwwood: I just saw your other commit, port is fine too.

dirk-thomas:
> You don't have to do a full design doc to communicate your intentions to the rest of the team. Take my rviz plan I posted to the mailing list. It's not a full design doc, but it hits the high points on the plan for doing the work as well as some key decisions. The purpose of which is to coordinate with others, inform them so they can collaborate if they want to, or raise concerns if they see them.
> My understanding of the meeting was to implement something that was the bare minimum for the requirements since we didn't have time to do a design first, not to make key design decisions without any discussion or design doc.

As mentioned before that decision was made not by me. And I also want to mention that sharing more information before the implementation phase is valuable - especially for larger development efforts. But in this case the work only took a couple of days so it is a balance of efficiency and overhead. Also nothing is set in stone and can be iterated on.

> Perhaps instead you could list the "extra" features that osrf_pycommon has that you don't need and why that matters?

I tend to approach this the other way around. Instead of arguing what a possible dependency has which I don't need I ask myself how much does it safe me by using a dependency (tested code, reduction of code, etc.). Since I don't "see" the benefit I decided not to use it. We might have a different perspective how much benefit the extra dependency would provide. Without a patch showing the potential difference it is difficult to objectively judge the pros and cons. And as mentioned before I already had a code base which provided the functionality and therefore I reused that.

> I believe it is a matter of scale. As you increase the number of entry_points and as the code which each entry_point imports becomes more complex, it gets a lot slower. It's also possible that later versions of Python have since improved the performance.
> I'm fine with the decision, but again it's just one of the things that ideally would have been discussed first.

Something like using entry points is in my opinion an implementation detail. I didn't even consider that it would result in any discussion to use the one mechanism in Python for implementing a plugin system. If there is evidence that the entry points are not suitable for this use case they can easily be replaced with something better.

dirk-thomas:
> > I will add a timeout to the daemon to self destruct after a certain amount of inactivity... Suggestions for a default value are welcome.

> Thanks, two hours?

I added a default timeout which shuts down the daemon after 2h of inactivity, see https://github.com/ros2/ros2cli/pull/1/files#diff-0db5b41e1e3495a1cb622aa6f10bdf78R23

ros2cli/ros2cli/node/strategy.py
wjwwood: Does the `DirectNode` always wait? If so, this is fine. The only case I was worried about was `ros2 topic list` returning the wrong data on the first shot because it didn't wait for enough input before returning.

See my other comment about the timeout being pretty short by default.

ros2cli/ros2cli/node/direct.py
```diff
@@ -0,0 +1,59 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import rclpy
+
+# TODO(mikaelarguedas) revisit this once it's specified
+HIDDEN_NODE_PREFIX = '_'
+
+DEFAULT_TIMEOUT = 0.1
```
wjwwood: This doesn't seem very long, why did you arrive at this value?

ros2topic/ros2topic/verb/echo.py
mikaelarguedas: It looks that was needed by @gerkey in https://github.com/ros2/cli_tools/pull/5 but I don't know why. It doesn't look to be used for recursion because it's listed as part of the `types` list that I understood as "primitive types that don't need recursion". Both the slots field and every array field in the python messages are represented as list so I don't see what the OrderedDict refers to

wjwwood: Ok, well, we can leave it. I was just hoping you knew why it was there.

dirk-thomas: When reading yaml the tool uses a custom representation to maintain the order: https://github.com/ros2/ros2cli/pull/1/files#diff-e3db6166c228ad905748ccc95b7e20c4R61

So this needs to stay.

ros2cli/ros2cli/node/strategy.py
dirk-thomas: The command line has a `timeout` option to specify how long the user wants to wait: https://github.com/ros2/ros2cli/pull/1/files#diff-0db5b41e1e3495a1cb622aa6f10bdf78R25

ros2cli/ros2cli/node/direct.py
dirk-thomas: The previous incarnation of the tool used 0.01s (see https://github.com/ros2/cli_tools/blob/de404fdf6fc18deedf247c2ab2cabff4f532731f/cli_tools_py/rosnode_list.py#L27). When I tested the command line tools I sometimes didn't get the complete result with that short timeout and therefore increased it. With 0.1s I got complete results for all of my tests. That certainly depends in the network topology and size of the graph. If you think we should use a different default value please propose one.

wjwwood:
`argcomplete` needs to be added to the installation instructions and CI I think. When I first tried this out I got:

```
% ros2 --help
  File "/Users/william/ros2_ws/install/bin/ros2", line 6, in <module>
    from pkg_resources import load_entry_point
  File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3036, in <module>
    @_call_aside
  File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3020, in _call_aside
    f(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3049, in _initialize_master_working_set
    working_set = WorkingSet._build_master()
  File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 654, in _build_master
    ws.require(__requires__)
  File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 968, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 854, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'argcomplete' distribution was not found and is required by ros2cli
```

ros2cli/ros2cli/node/direct.py
wjwwood: I don't have any less arbitrary number in mind. 0.5, 1.0 seconds? I guess we can just wait and see. Someone should run for a while without the daemon to see what number may be most appropriate with a realistic system, like the turtlebot demo.

dirk-thomas:
> `argcomplete` needs to be added to the installation instructions and CI I think.

Should I make it a "hard" dependency or remove it from the `setup.py` and make sure that it is used only when available (like here https://github.com/ros2/ros2cli/pull/1/files#diff-67470fc4e48a44a62467192fc8be262bR47)?

ros2cli/ros2cli/node/direct.py
mikaelarguedas: I didn't benchmark it but I often used more than a second on the non very beefy arm boards to be sure to have all topics. Then it will vary a lot according to the nature of the system so I'm fine with whatever value as long as it's easy to override it

dirk-thomas: I updated the default timeout to 0.5s.

wjwwood:
I don't mind either way. I don't have a problem making it a hard dependency so long as there is a new enough apt package, or you're willing to make it optional just for the deb.

dirk-thomas:
I will go with making it optional than. If the user doesn't have `argcomplete` there is simply no completion (gracefully falling back of course). Once `argcomplete` is available the tools will provide completion. This will take a little bit...

dirk-thomas:
@wjwwood There is something wrong with the last commit: 917fca363005d97c9972127fbceda4b4a93387d3

wjwwood:
@dirk-thomas sorry, I meant for just the gitignore.

Want me to force push, or make a new commit.

wjwwood:
I'm seeing some strange behavior on my machine with respect to the daemon. (that's the debugging stuff I accidentally committed to the `.gitignore`)

dirk-thomas:
Make a new commit for now since I already committed additional stuff. I will squash / rebase before the merge.

wjwwood:
I think the check for daemon is not working on my machine. I ran the daemon in a separate terminal, added a suffix to the node name when the daemon creates it, and added a print message when the tool trys to spawn a daemon. I get this:

```
william@farl ~/ros2_ws (git:master:f5c69a4)
% time ros2 node list -a
spawning daemon
_ros2cli_node
_ros2cli_node_daemon
_ros2cli_node_daemon
ros2 node list -a  0.40s user 0.06s system 47% cpu 0.973 total

william@farl ~/ros2_ws (git:master:f5c69a4)
% time ros2 node list -a
spawning daemon
_ros2cli_node
_ros2cli_node_daemon
_ros2cli_node_daemon
ros2 node list -a  0.39s user 0.05s system 46% cpu 0.953 total

william@farl ~/ros2_ws (git:master:f5c69a4)
% time ros2 node list -a
spawning daemon
_ros2cli_node
_ros2cli_node_daemon
_ros2cli_node_daemon
ros2 node list -a  0.40s user 0.06s system 47% cpu 0.968 total

william@farl ~/ros2_ws (git:master:f5c69a4)
% time ros2 node list -a
spawning daemon
_ros2cli_node
_ros2cli_node_daemon
_ros2cli_node_daemon
ros2 node list -a  0.38s user 0.06s system 46% cpu 0.948 total
```

I haven't looked into why the daemon is not properly detected.

wjwwood:
That seems to have fixed it.

dirk-thomas:
I passed the domain id explicitly to the daemon (efb834de7aca25462d0c263fa4ba9ea6f29c154e). That makes it more visible in e.g. `ps` which domain id the daemon belongs to.

After merging in #2 which enables passing a node name suffix I appended the pid to the direct node name / the domain id to the node name in the daemon (43028ff4441b325229a1cd1f09d979fc1922d803).

dirk-thomas:
This should be ready for re-re-review :wink:

Once I have a :+1: for merging I will rebase some of the commits to keep a clean history.

PR #3, opened 2017-06-14
add more commands

dirk-thomas:
This PR adds some more commands:

* Add the command `ros2 msg` with the verbs `list`, `package`, `packages`, and `show`
* Add the command `ros2 srv` with the verbs `list`, `package`, `packages`, and `show`
* Add the command `ros2 run`

The CI builds are again not really saying much (beside that the linter tests for the new code are passing):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2732/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/268/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2194/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2862/)

Ready for review.

### DIFF
--- a/ros2msg/package.xml
+++ b/ros2msg/package.xml
@@ -1,0 +1,23 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="2">
+  <name>ros2msg</name>
+  <version>0.0.0</version>
+  <description>
+    The msg command for ROS 2 command line tools.
+  </description>
+  <maintainer email="dthomas@osrfoundation.org">Dirk Thomas</maintainer>
+  <license>Apache License 2.0</license>
+
+  <depend>ros2cli</depend>
+
+  <exec_depend>ament_index_python</exec_depend>
+
+  <test_depend>ament_copyright</test_depend>
+  <test_depend>ament_flake8</test_depend>
+  <test_depend>ament_pep257</test_depend>
+
+  <export>
+    <build_type>ament_python</build_type>
+  </export>
+</package>
--- a/ros2run/ros2run/api/__init__.py
+++ b/ros2run/ros2run/api/__init__.py
@@ -1,0 +1,83 @@
+# Copyright 2017 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import subprocess
+
+from ros2cli.node.strategy import NodeStrategy
+from ros2pkg.api import get_prefix_path
+
+
+class MultipleExecutables(Exception):
+
+    def __init__(self, paths):
+        self.paths = paths
+
+
+class PackageNotFound(Exception):
+
+    def __init__(self, package_name):
+        self.package_name = package_name
+
+
+def get_executable_paths(*, node, package_name):
+    prefix_path = get_prefix_path(package_name)
+    if prefix_path is None:
+        raise PackageNotFound(package_name)
+    base_path = os.path.join(prefix_path, 'lib', package_name)
+    executable_paths = []
+    for dirpath, dirnames, filenames in os.walk(base_path):
+        # ignore folder starting with .
+        dirnames[:] = [d for d in dirnames if d[0] not in ['.']]
+        dirnames.sort()
+        # select executable files
+        for filename in sorted(filenames):
+            path = os.path.join(dirpath, filename)
+            if os.access(path, os.X_OK):
+                executable_paths.append(path)
+    return executable_paths
+
+
+def get_executable_path(*, node, package_name, executable_name):
+    paths = get_executable_paths(node=node, package_name=package_name)
+    paths2base = {
+        p: os.path.basename(p) for p in paths
+        if os.path.basename(p) == executable_name}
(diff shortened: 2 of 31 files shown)


ros2msg/package.xml
```diff
@@ -0,0 +1,22 @@
...
+  </description>
+  <maintainer email="dthomas@osrfoundation.org">Dirk Thomas</maintainer>
+  <license>Apache License 2.0</license>
+
+  <depend>ros2cli</depend>
+
+  <exec_depend>ament_index_python</exec_depend>
+
+  <test_depend>ament_copyright</test_depend>
+  <test_depend>ament_flake8</test_depend>
```
mikaelarguedas: is it on purpose that none of the new packages run the pep257 test ?

dirk-thomas: Yes, pep257 linting is provided by a flake8 plugin.

mikaelarguedas: I don't think we ever installed this plugin because not available from debian packages (only via pip like `import-order`). Can't find any trace of it in the ci scripts either... Is there a PR I missed that was adding it ?

dirk-thomas: Should we install `flake8-docstrings` in the CI jobs then (as we do for `flake8-import-order`)? Or should we add `pep257` tests explicitly?

mikaelarguedas: flake8-docstrings is not available via apt so I'd lean towards using pep257 explicitely as we do everywhere else. But I don't feel very strongly about it.

dirk-thomas: Sounds good to me. I will add them for all packages in this repo.

dirk-thomas: I updated this PR and created #5 for the already merged packages.

mikaelarguedas:
That's pretty cool!

While testing a few things surprised me:
- ros2 msg:
  - `ros2 msg show` expects "PackageName MessageName" but `ros2 msg list` gives me "PackageName/MessageName". So that means we cant copy paste the result of `list` to use `show`. It both differ from rosmsg in ros1 and is less convenient
- ros2 run
  - run doesn't complete executable name so I need to know the package and the name  to be able to run a node.
  - if I pick an executable it cannot find it:
```
$ros2 run examples_rclcpp_minimal_publisher examples_rclcpp_minimal_publisher_lambda
No executable found
```
Same terminal
```
$ examples_rclcpp_minimal_publisher_lambda
Publishing: [Hello, world! 0]
Publishing: [Hello, world! 1]
```

dirk-thomas:
`ros2 run` looks in the package specific `libexec` folder (`<prefix>/share/lib/<pkgname>`) for executables. Since non of our current packages installs any executables there you won't get any completion result.

Regarding `ros2 msg`: I usually prefer to pass information atomically. Otherwise e.g. your completion list for the combined argument would be extremely long. So I am not sure if we should change it to be the same as in ROS 1 or update the `ros2 msg list` output to print both parts (the package name and message name) separately.

dirk-thomas:
Updated. See https://github.com/ros2/ros2cli/pull/3/files#diff-c2d4477e31b61377c91386997085d179R24 and https://github.com/ros2/ros2cli/pull/3/files#diff-c351f7ddc76367ae3cf77f70feede99cR24

mikaelarguedas:
Note: I tried to install my executables in `lib` but couldn't get them to run `No executable found`. Also not sure how we should install ament_python packages now, but all this can be discussed/addressed after beta2

dirk-thomas:
Did you install the executables in `lib` (which is the wrong location or `lib/<pkgname>`? When I tried the code it worked for the later location.

wjwwood:
I got it to work with this patch to `demo_nodes_cpp`:

```diff
diff --git a/demo_nodes_cpp/CMakeLists.txt b/demo_nodes_cpp/CMakeLists.txt
index 0e4f484..f6b0034 100644
--- a/demo_nodes_cpp/CMakeLists.txt
+++ b/demo_nodes_cpp/CMakeLists.txt
@@ -22,7 +22,7 @@ function(custom_executable subfolder target)
     "rcutils"
     "std_msgs")
   install(TARGETS ${target}
-    DESTINATION bin)
+    DESTINATION lib/${PROJECT_NAME})
 endfunction()

 # Tutorials of Publish/Subscribe with Topics
```

ros2run/ros2run/api/__init__.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+class MultipleExecutables(Exception):
+
+    def __init__(self, paths):
+        self.paths = paths
+
+
+def get_executable_paths(*, node, package_name):
+    prefix_path = get_prefix_path(package_name)
```
wjwwood: What if this is `None`?

wjwwood: I get this:

```
% ros2 run taco talker
Failed to load entry point 'service': No module named 'ros2service.command.service'
  File "/Users/william/ros2_ws/install/bin/ros2", line 11, in <module>
    load_entry_point('ros2cli', 'console_scripts', 'ros2')()
  File "/Users/william/ros2_ws/build/ros2cli/ros2cli/cli.py", line 64, in main
  File "/Users/william/ros2_ws/build/ros2run/ros2run/command/run.py", line 50, in main
    executable_name=args.executable_name)
  File "/Users/william/ros2_ws/build/ros2run/ros2run/api/__init__.py", line 45, in get_executable_path
    paths = get_executable_paths(node=node, package_name=package_name)
  File "/Users/william/ros2_ws/build/ros2run/ros2run/api/__init__.py", line 30, in get_executable_paths
    base_path = os.path.join(prefix_path, 'lib', package_name)
  File "/usr/local/Cellar/python3/3.6.0/Frameworks/Python.framework/Versions/3.6/lib/python3.6/posixpath.py", line 78, in join
    a = os.fspath(a)
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

dirk-thomas: Fixed: 1232edf609ceaef883221a503ec35a34b6343731

wjwwood: Now it says `No executable found`, which to be more accurate, the package was not found. It would be better if it said that.

```
% ros2 run taco talker
No executable found
```

PR #4, opened 2017-06-15
Refactor get topic names and types

wjwwood:
Connects to ros2/ros2#361


wjwwood:
Example usage (with `talker` and `ros2 topic pub /chatter std_msgs/Int32 '{"data": "1"}'`):

```
% ros2 topic list -t
/chatter [std_msgs/Int32, std_msgs/String]
/parameter_events [rcl_interfaces/ParameterEvent]

% ros2 topic list -t --no-demangle
rt/chatter [std_msgs::msg::dds_::Int32_, std_msgs::msg::dds_::String_]
rt/parameter_events [rcl_interfaces::msg::dds_::ParameterEvent_]
```

dirk-thomas:
The new package should have a pep257 test (as in #5).

wjwwood:
> The new package should have a pep257 test (as in #5).

1c3d5b6

wjwwood:
e62767d handles the multiple types like so:

```
% ros2 topic echo /chatter
Cannot echo topic '/chatter', as it contains more than one type: [std_msgs/Int32, std_msgs/String]

% echo $?
1
```

I'll do an audit of all uses of the topic names and types to ensure it is being dealt with everywhere before putting these into review.

wjwwood:
rebased

ros2service/ros2service/verb/call.py
```diff
@@ -0,0 +1,97 @@
...
+
+        def float_or_int(input_string):
+            try:
+                value = float(input_string)
+            except ValueError:
+                raise argparse.ArgumentTypeError("'%s' is not an integer or float" % input_string)
+            return value
+
+        parser.add_argument(
+            '-l', '--loop', metavar='N', default=None, type=float_or_int,
```
dirk-thomas: Maybe rename to something like "repeat" or "rate"?

Nitpick `default=None` is the implicit default so it can be omitted.

ros2service/ros2service/verb/call.py
```diff
@@ -0,0 +1,97 @@
...
+            help='Repeat the call every N seconds')
+
+    def main(self, *, args):
+        return main(args)
```
dirk-thomas: I would suggest to call `requester(...)` here and get rid of the `main` function.

ros2service/setup.py
```diff
@@ -0,0 +1,39 @@
...
+
+setup(
+    name='ros2service',
+    version='0.0.0',
+    packages=find_packages(exclude=['test']),
+    install_requires=['ros2cli'],
+    author='Dirk Thomas',
+    author_email='dthomas@osrfoundation.org',
+    maintainer='Dirk Thomas',
+    maintainer_email='dthomas@osrfoundation.org',
```
dirk-thomas: Your name?

ros2topic/ros2topic/verb/echo.py
```diff
@@ -77,7 +77,14 @@ def subscriber(node, topic_name, message_type, callback):
...
         for n, t in topic_names_and_types:
             if n == topic_name:
-                message_type = t
+                if len(t) > 1:
+                    print(
+                        "Cannot echo topic '%s', as it contains more than one type: [%s]" %
+                        (topic_name, ', '.join(t)),
+                        file=sys.stderr
+                    )
+                    sys.exit(1)
```
dirk-thomas: You should use `raise RuntimeError('Cannot echo topic '%s', as it contains more ...')` here instead. Same as a few lines below.

PR #12, opened 2017-06-21
Improve executable selection

dirk-thomas:
The first commit removes the unused `node` argument from the signatures.

The second commit considers `PATHEXT` on Windows to allow calling e.g. `ros2 run demo_nodes_py talker` (rather than needing to type `ros2 run demo_nodes_py talker.exe`).

### DIFF
--- a/ros2run/ros2run/api/__init__.py
+++ b/ros2run/ros2run/api/__init__.py
@@ -15,7 +15,6 @@
 import os
 import subprocess
 
-from ros2cli.node.strategy import NodeStrategy
 from ros2pkg.api import get_prefix_path
 
 
@@ -31,7 +30,7 @@ class PackageNotFound(Exception):
         self.package_name = package_name
 
 
-def get_executable_paths(*, node, package_name):
+def get_executable_paths(*, package_name):
     prefix_path = get_prefix_path(package_name)
     if prefix_path is None:
         raise PackageNotFound(package_name)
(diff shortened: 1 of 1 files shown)


ros2run/ros2run/api/__init__.py
```diff
@@ -49,11 +49,21 @@ def get_executable_paths(*, node, package_name):
...
-        if os.path.basename(p) == executable_name}
+def get_executable_path(*, package_name, executable_name):
+    paths = get_executable_paths(package_name=package_name)
+    paths2base = {}
+    for p in paths:
+        basename = os.path.basename(p)
+        if basename == executable_name:
+            # pick exact match
+            paths2base[p] = basename
+        elif sys.platform == 'win32':
```
mikaelarguedas: out of curiosity if I have a `talker` and a `talker.exe` this will select `talker.exe` first and discard `talker`, is that correct?

dirk-thomas: No, if there is an exact match it will select that one (line 57-59). Only if not filenames with additional extensions according to `PATHEXT` are being considered.

So in your case if you call `<pkgname> talker` it will execute the file named `talker`.

mikaelarguedas: oh right, yeah I saw it the other way around, if I passed `talker.exe` it would pick that rather than trying to remove the extension and look for generic extension points without extension. But it does make a lot more sense as you described it :confounded:

PR #13, opened 2017-06-21
Removed node from call to get_executable path in ros2 run command,

allenh1:
see ros2/ros2cli#12.

### DIFF
--- a/ros2run/ros2run/command/run.py
+++ b/ros2run/ros2run/command/run.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 from ros2cli.command import CommandExtension
-from ros2cli.node.strategy import NodeStrategy
 from ros2pkg.api import package_name_completer
 from ros2run.api import ExecutableNameCompleter
 from ros2run.api import get_executable_path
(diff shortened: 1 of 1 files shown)


ros2run/ros2run/command/run.py
```diff
@@ -45,7 +45,7 @@ def main(self, *, parser, args):
         with NodeStrategy(args) as node:
             try:
                 path = get_executable_path(
-                    node=node, package_name=args.package_name,
+                    package_name=args.package_name,
```
wjwwood: If you're removing node here, doesn't the `with NodeStrategy(args) as node:` above this become unused?

dirk-thomas:
The whole `with` block and the variable `node` need to go away.

allenh1:
@wjwwood good point, fixed to reflect the changes.

ros2run/ros2run/command/run.py
```diff
@@ -42,19 +42,18 @@ def add_arguments(self, parser, cli_name):
...
+                executable_name=args.executable_name)
+        except PackageNotFound:
+            raise RuntimeError(
+                "Package '{args.package_name}' not found"
+                .format_map(locals()))
+        except MultipleExecutables as e:
+            msg = 'Multiple executables found:'
+            for p in e.paths:
+                msg += '\n- {p}'.format_map(locals())
                 raise RuntimeError(msg)
```
allenh1: forgot to indent this line

dirk-thomas: Now this line is indented wrongly. It was on the same level as `for` before.

mikaelarguedas:
* Linux packaging [Build Status](http://ci.ros2.org/job/ci_packaging_linux/24/)

allenh1:
@dirk-thomas fixed -- should I rerun the CI?

PR #14, opened 2017-06-22
add unit tests

dirk-thomas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2802/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/321/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2269/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2940/)

Ready for review.

### DIFF
--- a/ros2msg/test/test_cli.py
+++ b/ros2msg/test/test_cli.py
@@ -1,0 +1,58 @@
+# Copyright 2017 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import subprocess
+
+
+def test_cli():
+    packages_cmd = ['ros2', 'msg', 'packages']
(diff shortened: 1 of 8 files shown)


ros2msg/test/test_cli.py
```diff
@@ -0,0 +1,58 @@
...
+    packages_cmd = ['ros2', 'msg', 'packages']
+    packages_result = subprocess.run(packages_cmd, stdout=subprocess.PIPE, check=True)
+    package_names = packages_result.stdout.decode().splitlines()
+
+    # explicit dependencies of this package will for sure be available
+    assert 'std_msgs' in package_names
+
+    count = 0
+    for package_name in package_names:
+        package_cmd = ['ros2', 'msg', 'package', package_name]
```
wjwwood: The same applies to `ros2 srv package[s]` verbs.

@ros2/team can you guys weigh in on this? Do you prefer `ros2 msg package <pkg>` or `ros2 msg list --package <pkg>` (also `ros2 msg packages` vs. `ros2 msg list --packages`)?

If the consensus is that the alternative verb names I proposed would be better, then I'll open a pr to update the commands later.

dirk-thomas:
I will go ahead and merge these tests. Please continue the conversation / reply to @wjwwood question.

ros2msg/test/test_cli.py
dirk-thomas: The advantage I saw with having separate "verbs" is that each of them is self contained. Additional ones can easily be added (without fiddling with existing code, e.g. by external Python packages). Also with a single "verb" handling multiple different cases you have to "read" the usage to figure out which option can be used together with others and how it might affect the output (e.g. the `--packages` and `--package` options would be mutually exclusive).

PR #15, opened 2017-06-26
various fixes and improvements

### LINKED ISSUE #17: pyyaml dependency in ros2 topic echo
I didn't have `pyyaml` installed and got this error:

```
$ ros2 run demos_nodes_cpp listener
Failed to load entry point 'echo': No module named 'yaml'
```

I am adding `python3 -m pip install pyyaml` to the installation instructions, but I noticed it's not listed in the setup.py for `ros2topic` where I think it should be. Is that correct?

dirk-thomas:
Beside other changes fixes #17.

### DIFF
--- a/ros2run/ros2run/command/run.py
+++ b/ros2run/ros2run/command/run.py
@@ -22,7 +22,7 @@ from ros2run.api import run_executable
 
 
 class RunCommand(CommandExtension):
-    """Various run related sub-commands."""
+    """Run a package specific executable."""
 
     def add_arguments(self, parser, cli_name):
         arg = parser.add_argument(
--- a/ros2cli/ros2cli/daemon/__init__.py
+++ b/ros2cli/ros2cli/daemon/__init__.py
@@ -57,6 +57,8 @@ def main(*, script_name='_ros2_daemon', argv=None):
                 _print_invoked_function_name(node.get_node_names))
             server.register_function(
                 _print_invoked_function_name(node.get_topic_names_and_types))
+            server.register_function(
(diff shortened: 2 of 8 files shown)


dirk-thomas:
Ready for review.

dhood:
I opened https://github.com/ros/rosdistro/pull/15377 because after speaking with @nuclearsandwich I understood that that would be necessary upstream in order for this package to get built on the buildfarm since 41d411e93ed09b6742e60e5420fbfd86d66d2bbb

dirk-thomas:
To determine if a package contains executables required to crawl the file system. Since that is a significant overhead it isn't being done for the package name completion.

ros2run/ros2run/command/run.py
```diff
@@ -22,7 +22,7 @@
 
 
 class RunCommand(CommandExtension):
-    """Various run related sub-commands."""
+    """Run a package specific executable."""
```
mikaelarguedas: Maybe that's my english, "Run a package **'s** specific executable." ?

wjwwood: It looks ok as-is to me. I read it as `Run a "package specific" executable.`  Where the quoted string is an adjective and is optional, i.e. it would read as well (though less specific) as `Run an executable.`.

wjwwood: A totally unambiguous way to say it is `Run an executable of a specific package.`

mikaelarguedas: ok thanks for clarifying!

PR #19, opened 2017-06-27
use yaml for parsing msg and srv values

wjwwood:
This should be backwards compatible with the JSON style inputs. Also, these packages already needed yaml, so it's not a new dependency. It also makes using this feature on Windows easier since you don't have to so quote escaping in a lot of cases.

@Kukanani FYI you should be able to do `ros2 topic pub /chatter std_msgs/String "data: Hello world"` with this pr.

### DIFF
--- a/ros2service/package.xml
+++ b/ros2service/package.xml
@@ -12,6 +12,8 @@
   <depend>rclpy</depend>
   <depend>ros2cli</depend>
 
+  <exec_depend>python3-yaml</exec_depend>
+
   <test_depend>ament_copyright</test_depend>
   <test_depend>ament_flake8</test_depend>
   <test_depend>ament_pep257</test_depend>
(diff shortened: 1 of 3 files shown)


dhood:
I am in favour of this because I was annoyed I couldn't copy-paste stuff from ROS 1

wjwwood:
I changed the base to be `master` now that https://github.com/ros2/ros2cli/pull/15 is merged.

wjwwood:
We still need to make a decision on whether or not this should get merged today. CI:

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/2825/)
- Linux-aarch64:
  - [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/330/)
- macOS:
  - [Build Status](http://ci.ros2.org/job/ci_osx/2282/)
- Windows:
  - [Build Status](http://ci.ros2.org/job/ci_windows/2962/)

wjwwood:
I tested the behavior manually. I'll go ahead an merge since we've got a few pending changes that must go in anyways.

PR #24, opened 2017-06-27
fix population of recursive message fields

dirk-thomas:
Fixes #22.

Ready for review.

PR #27, opened 2017-06-27
also handle ValueError nicely

dirk-thomas:
Follow up of #24.

This handles `ValueErrors` as well. To reproduce try:

```
ros2 topic pub initialpose geometry_msgs/PoseWithCovarianceStamped "header:
  stamp:
    sec: 'sec'
    nanosec: 0.5
  frame_id: 'map'   
pose:
  pose:
    position: {x: 5.5, y: 4.5, z: 0.0}
    orientation: {x: 0.0, y: 0.0, z: 0.0, w: 1.0}
  covariance: [0.5, 0.5, 0.0, 0.0, 0.0, 0.0, 0.0, 0.5, 0.5, 0.0, 0.0, 0.0, 0.0, 0.0,    0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0,
    0.0, 0.0, 0.0, 0.0, 0.0, 0.0]"
```

### DIFF
--- a/ros2topic/ros2topic/verb/pub.py
+++ b/ros2topic/ros2topic/verb/pub.py
@@ -99,6 +99,8 @@ def set_msg_fields(msg, values):
                 raise SetFieldError(
                     '{field_name}.{e.field_name}'.format_map(locals()),
                     e.exception)
+        except ValueError as e:
+            raise SetFieldError(field_name, e)
         try:
             setattr(msg, field_name, value)
         except Exception as e:


ros2topic/ros2topic/verb/pub.py
```diff
@@ -99,6 +99,9 @@ def set_msg_fields(msg, values):
                 raise SetFieldError(
                     '{field_name}.{e.field_name}'.format_map(locals()),
                     e.exception)
+        except ValueError as e:
+            raise SetFieldError(
+                '{field_name}'.format_map(locals()), e)
```
wjwwood: Do you not have to do `**locals()`?

dirk-thomas: No, not with `format_map`. There is already plenty of other code using this snippet.

wjwwood: Ah sorry, thought it was just format, lgtm.

dirk-thomas: Looking at the line again the whole use of `format_map` is pretty unnecessary. I used it for the field name concatenation in another line but here it only contains `field_name`. I will update it to use the literal variable instead...

PR #30, opened 2017-06-28
add exec_depend on python3-pkg-resources

mikaelarguedas:
the rosdep key has been added in https://github.com/ros/rosdistro/pull/15406

### DIFF
--- a/ros2cli/package.xml
+++ b/ros2cli/package.xml
@@ -9,6 +9,7 @@
   <maintainer email="dthomas@osrfoundation.org">Dirk Thomas</maintainer>
   <license>Apache License 2.0</license>
 
+  <exec_depend>python3-pkg-resources</exec_depend>
   <exec_depend>rclpy</exec_depend>
 
   <test_depend>ament_copyright</test_depend>


ros2cli/package.xml
```diff
@@ -9,6 +9,7 @@
   <maintainer email="dthomas@osrfoundation.org">Dirk Thomas</maintainer>
   <license>Apache License 2.0</license>
 
+  <exec_depend>python3-pkg-resources</exec_depend>
```
dirk-thomas: Is this only a run dependency? Is it not already used at buildtime?

mikaelarguedas: I thought that `ament_python` packages rely on setuptools at buildtime that provide this dependency. So I assumed that runtime would be enough, isn't it ?

dirk-thomas: Ok, I think that is fine for now with the `setuptools` dependency being injected by bloom.

mikaelarguedas: Sounds good merging as is then

PR #31, opened 2017-06-28
truncate arrays, bytes, and strings by default, add option to show in full or use custom threshold

dirk-thomas:
This will allow to print every image message with: `ros2 topic echo /image`

### DIFF
--- a/ros2topic/ros2topic/verb/echo.py
+++ b/ros2topic/ros2topic/verb/echo.py
@@ -55,12 +55,12 @@ class EchoVerb(VerbExtension):
             help='Output all recursive fields separated by commas (e.g. for '
                  'plotting)')
         parser.add_argument(
-            '--full-length', action='store_true',
+            '--full-length', '-f', action='store_true',
             help="Output all elements for arrays, bytes, and string with a "
                  "length > '--truncate-length', by default they are truncated "
                  "after '--truncate-length' elements with '...''")
         parser.add_argument(
-            '--truncate-length', type=unsigned_int, default=DEFAULT_TRUNCATE_LENGTH,
+            '--truncate-length', '-l', type=unsigned_int, default=DEFAULT_TRUNCATE_LENGTH,
             help='The length to truncate arrays, bytes, and string to '
                  '(default: %d)' % DEFAULT_TRUNCATE_LENGTH)
 
--- a/ros2topic/ros2topic/verb/echo.py
+++ b/ros2topic/ros2topic/verb/echo.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from argparse import ArgumentTypeError
 from collections import OrderedDict
 import importlib
 import sys
@@ -23,6 +24,18 @@ from ros2topic.api import get_topic_names_and_types
 from ros2topic.api import TopicNameCompleter
 from ros2topic.verb import VerbExtension
 import yaml
+
+DEFAULT_TRUNCATE_LENGTH = 128
+
+
+def unsigned_int(string):
+    try:
+        value = int(string)
+    except ValueError:
+        value = -1
+    if value < 0:
+        raise ArgumentTypeError('value must be non-negative integer')
+    return value
 
 
 class EchoVerb(VerbExtension):
@@ -41,6 +54,15 @@ class EchoVerb(VerbExtension):
             '--csv', action='store_true',
             help='Output all recursive fields separated by commas (e.g. for '
                  'plotting)')
+        parser.add_argument(
+            '--full-length', action='store_true',
+            help="Output all elements for arrays, bytes, and string with a "
+                 "length > '--truncate-length', by default they are truncated "
+                 "after '--truncate-length' elements with '...''")
+        parser.add_argument(
+            '--truncate-length', type=unsigned_int, default=DEFAULT_TRUNCATE_LENGTH,
+            help='The length to truncate arrays, bytes, and string to '
+                 '(default: %d)' % DEFAULT_TRUNCATE_LENGTH)
 
     def main(self, *, args):
         return main(args)
@@ -49,9 +71,9 @@ class EchoVerb(VerbExtension):
 def main(args):
     if not args.csv:
         register_yaml_representer()
-        callback = subscriber_cb
+        callback = subscriber_cb(args)
     else:
-        callback = subscriber_cb_csv
+        callback = subscriber_cb_csv(args)
     with DirectNode(args) as node:
         subscriber(node, args.topic_name, args.message_type, callback)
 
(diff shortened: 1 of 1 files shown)


dirk-thomas:
My rational for using truncation by default is the following:

* For large messages I want "non-broken" behavior by default, if long messages are printed in full by default that basically prevents all messages to be outputted.
* For small message I want to see all array values, and all character in bytes and strings.
* ROS 1 has options to truncate arrays / strings (actually to show nothing but the length). Since these options have been added later in order to not previous behavior they don't achieve the goal of the first bullet. Therefore I don't think we should mimic ROS 1 behavior here.

Additional improvements could be:

* Show the number of truncated values / characters or the full length for information purposes.
* Allow to pass a custom threshold for the truncation.

mikaelarguedas:
I'm not sure if the default behavior should be "not show the entire message". While making sense for large messages it seems odd that if I ask to print the content of a message I get only part of it.
I like the `--no-arr` option in ROS1 allowing to print messages without printing the arrays. It doesnt allow you to know that the content of your image changes in that case
`rostopic hz` is another useful way of knowing of often you are receiveing messages.
An alternative could be to print a warning before the message if the size exceed an arbitrary threshold. This way users would be warned that displaying it would screw up their performance but not making an arbitrary decision on the size of the message they can display.

Maybe if you can clarify what is the use case we try to solve with this approach we can discuss alternatives and decide which one is the best suited default behaviour

mikaelarguedas:
Well I guess I should have sent this message when I wrote it after the offline discussion :confounded: and not 40 min late... thanks for providing a rationale.

I see your reasoning, I agree that defaulting to something having bad performances is not ideal. I think that if this is to be merged we should at least allow users to chose the threshold.

I'd like to have more input from @ros2/team before merging this because that's a change from ROS1 behavior (and general expected behavior) that should be discussed and considered carefully before imposing it to users. Then I agree we can iterate on it later so it could be rolled out as is in beta2 and modified later.

dirk-thomas:
> we should at least allow users to chose the threshold.

I will add an additional option for this. I guess one threshold for all three types is sufficient?

> I'd like to have more input from ros2/team before merging this because that's a change from ROS1 behavior (and general expected behavior) that should be discussed and considered carefully before imposing it to users.

Before discussing what we should do I would like to have a consensus on the goals. Do we agree about the two goals described in the first two bullets (non-broken for "large" msgs, see all values for "small" msgs)? If each of you could answer that question first that would be helpful.

mikaelarguedas:
> I will add an additional option for this. I guess one threshold for all three types is sufficient?

Yes I think that's sufficient

> non-broken for "large" msgs, see all values for "small" msgs

Yeah I think that's a good goal to have if we agree that someone calling `ros2 topic echo` should not expect a full message if it's a long one. 

Looking at the patch, this truncature also applies to fixed length arrays and not only the dynamic ones right ?

dirk-thomas:
> Looking at the patch, this truncature also applies to fixed length arrays and not only the dynamic ones right?

Yes, good point. I will rename the constant to `TRUNCATE_LENGTH`.

clalancette:
> My rational for using truncation by default is the following:
> 
>     For large messages I want "non-broken" behavior by default, if long messages are printed in full by default that basically prevents all messages to be outputted.

Yeah, I definitely agree with this.  Otherwise the user has no idea what is wrong with their invocation of echo.

>     For small message I want to see all array values, and all character in bytes and strings.

Seems sane to me.

>     ROS 1 has options to truncate arrays / strings (actually to show nothing but the length). Since these options have been added later in order to not previous behavior they don't achieve the goal of the first bullet. Therefore I don't think we should mimic ROS 1 behavior here.

I can see that option being useful here as well, but I'm happy to defer that until later.

wjwwood:
I'm ok with having automatic truncating (or automatic summarizing, however you want to call it). +1 for the proposed changes.

Though I'm dubious about the rationale that large messages are broken by default in this case. Whether or not it can print all the messages depends on the speed of your computer and the size of message. Not whether or not it is truncated. Presumably a large enough message (or high enough frequency messages) would also result in not every message being printed.

I like the proposed changes because it's a better user experience, i.e. the summary is more useful than the raw data in most cases, imo.

dirk-thomas:
I updated the patch. It has an option to pass a custom threshold (which can also be zero).

Ready for re-review.

dirk-thomas:
I could remove `--full-length` and use any negative value for `--truncate-length` to select that behavior?

ros2topic/ros2topic/verb/echo.py
```diff
@@ -41,6 +54,15 @@ def add_arguments(self, parser, cli_name):
             '--csv', action='store_true',
             help='Output all recursive fields separated by commas (e.g. for '
                  'plotting)')
+        parser.add_argument(
+            '--full-length', action='store_true',
```
wjwwood: I think it would be good to have a short flag for this, maybe `-f` or `-a`. @ros2/team what does everyone else think?

ros2topic/ros2topic/verb/echo.py
```diff
@@ -41,6 +54,15 @@ def add_arguments(self, parser, cli_name):
...
+            help="Output all elements for arrays, bytes, and string with a "
+                 "length > '--truncate-length', by default they are truncated "
+                 "after '--truncate-length' elements with '...''")
+        parser.add_argument(
+            '--truncate-length', type=unsigned_int, default=DEFAULT_TRUNCATE_LENGTH,
```
wjwwood: Same here, maybe `-l`.

wjwwood:
Personally I like the explicit `--full-length` argument. It is more visible/obvious when someone reads it (imagine it's in a tutorial). With the implicit full length on `-1` you have to assume or read the docs to understand it. You could also do both, but I don't see a reason for it.

clalancette:
> I could remove --full-length and use any negative value for --truncate-length to select that behavior?

I'd be OK with that, though if you did that, I would suggest changing the name of the flag to "max-print-length" or something like that.

ros2topic/ros2topic/verb/echo.py
clalancette: Yeah, I also think it would be good to have short options.

dirk-thomas: Done: 0777ef20bf0de97e3c8c4efd24ccf244e1ac7407

ros2topic/ros2topic/verb/echo.py
dirk-thomas: Done: 0777ef20bf0de97e3c8c4efd24ccf244e1ac7407

dirk-thomas:
I will go ahead soon and merge this as-is if there are no further comments.

mikaelarguedas:
that sounds good to me we can iterate on a summarize option or an alternative after the release

dirk-thomas:
Updated test accordingly in 1156c0d968ec9db082e775292d7df3b6f944ccc0.

PR #33, opened 2017-06-29
Improve windows daemon

### LINKED ISSUE #10: daemon prevents cmd.exe from closing on Windows
After running a `ros2 topic ...` command in a command prompt on Windows, the command prompt will hang when you type `exit`. This is because the daemon is somehow still attached to the command prompt I think.

This isn't a great user experience, and I wouldn't have know without deep knowledge about what's going on. I'm not sure if we can run the daemon in a way that avoids this or not, but it would be good if we could.

dirk-thomas:
The first commit fixes #10.

The second commit makes sure that the folder which the daemon was launched from can be deleted even while the daemon is still running.


Kukanani:
Confirmed that both features are working on Win 10 from source.

Issue #34, opened 2017-06-30
daemon prevents rebuilding after running on Windows due to having dll open


tfoote:
After using the ros2 cli the daemon prevents a rebuild from completing successfully on windows. 

A way to have introspection as to the cause of this failure would be very helpful. Or else a workaround that would allow the build to request the daemon to shutdown or other technique to not block the build would provide a better user experience. 

```
  CMake Error at cmake_install.cmake:46 (file):
    file INSTALL cannot copy file "C:/dev/ros2/build/rcl/Release/rcl.dll" to
    "C:/dev/ros2/install/bin/rcl.dll".


C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\V140\Microsoft.CppCommon.targets(133,5): error MSB3073: The command "
setlocal\r [C:\dev\ros2\build\rcl\INSTALL.vcxproj]
C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\V140\Microsoft.CppCommon.targets(133,5): error MSB3073: "C:\Program F
iles\CMake\bin\cmake.exe" -DBUILD_TYPE=Release -P cmake_install.cmake\r [C:\dev\ros2\build\rcl\INSTALL.vcxproj]
C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\V140\Microsoft.CppCommon.targets(133,5): error MSB3073: if %errorleve
l% neq 0 goto :cmEnd\r [C:\dev\ros2\build\rcl\INSTALL.vcxproj]
C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\V140\Microsoft.CppCommon.targets(133,5): error MSB3073: :cmEnd\r [C:\
dev\ros2\build\rcl\INSTALL.vcxproj]
C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\V140\Microsoft.CppCommon.targets(133,5): error MSB3073: endlocal & ca
ll :cmErrorLevel %errorlevel% & goto :cmDone\r [C:\dev\ros2\build\rcl\INSTALL.vcxproj]
C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\V140\Microsoft.CppCommon.targets(133,5): error MSB3073: :cmErrorLevel
\r [C:\dev\ros2\build\rcl\INSTALL.vcxproj]
[...]

PR #38, opened 2017-06-30
add daemon command with verbs status, start, stop

dirk-thomas:
Addresses #34.

### DIFF
--- a/ros2cli/ros2cli/command/daemon.py
+++ b/ros2cli/ros2cli/command/daemon.py
@@ -1,0 +1,35 @@
+# Copyright 2016-2017 Dirk Thomas
+# Copyright 2017 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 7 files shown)


Kukanani:
LGTM, confirmed working on Win 10.

Would it be useful to add some context to the help output, explaining what the daemon is/does? Maybe just a pointer to documentation, if such documentation exists.

clalancette:
This also works for me on Linux.  There's a couple of overall things that would make this nicer:

1.  In all of the sub-commands, it would be nice to tell the user which ros-domain-id we are starting/stopping/statusing the daemon.
1.  Starting and stopping the daemon for a particular domain id is controlled through the environment variable ROS_DOMAIN_ID.  How do we control the rmw_implementation?  It would be nice to document how to control both of these somewhere.

dirk-thomas:
Thanks for testing. More documentation is always nice. I am not sure I will add it right now though. If you have specific ideas what to add please consider making a PR.

The rmw impl. is selected by the environment variable `RMW_IMPLEMENTATION`. If not set a default is being selected. So "normal" users will only use one rmw impl at a time I don't think it needs exposure in the tool atm.

mikaelarguedas:
> Starting and stopping the daemon for a particular domain id is controlled through the environment variable ROS_DOMAIN_ID. How do we control the rmw_implementation? It would be nice to document how to control both of these somewhere.

I think both of them (`ROS_DOMAIN_ID` and `RMW_IMPLEMENTATION`) should be documented globally because they apply to every ROS2 executable and not only this tool. For now the rmw information is documented in the [Working-with-multiple-RMW-implementations](https://github.com/ros2/ros2/wiki/Working-with-multiple-RMW-implementations) wiki page

dirk-thomas:
Merging this for now as is. With potential improvements coming in the future.

dirk-thomas:
Follow up fix in 45bbfab5f717388b89a9cf85c3220b66c5c523fc

PR #41, opened 2017-07-25
added --prefix argument to ros2 run

volkandre:
According to the proposal by @wjwwood on my [question](http://answers.ros.org/question/267261/how-can-i-run-ros2-nodes-in-a-debugger-eg-gdb/), I am providing a pull request to add the '--prefix' argument from 'rosrun' to 'ros2 run'.

### DIFF
--- a/ros2run/ros2run/command/run.py
+++ b/ros2run/ros2run/command/run.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import shlex
+
 from ros2cli.command import CommandExtension
 from ros2pkg.api import package_name_completer
 from ros2pkg.api import PackageNotFound
@@ -26,9 +28,10 @@ class RunCommand(CommandExtension):
 
     def add_arguments(self, parser, cli_name):
         arg = parser.add_argument(
-            '--prefix', nargs=1,
+            '--prefix',
             help="Prefix command, which should go before the executable, "
-                 "like gdb or valgrind (e.g. --prefix \'gdb -ex run --args\').")
+                 "like gdb or valgrind. Command must be wrapped in quotes "
+                 "(e.g. --prefix 'gdb -ex run --args').")
         arg = parser.add_argument(
             'package_name',
             help="Name of the ROS package")
@@ -60,5 +63,5 @@ class RunCommand(CommandExtension):
             raise RuntimeError(msg)
         if path is None:
             return 'No executable found'
-        prefix=args.prefix[0] if args.prefix is not None else ''
-        return run_executable(prefix=prefix, path=path, argv=args.argv)
+        prefix = shlex.split(args.prefix) if args.prefix is not None else None
+        return run_executable(path=path, argv=args.argv, prefix=prefix)
--- a/ros2run/ros2run/command/run.py
+++ b/ros2run/ros2run/command/run.py
@@ -29,8 +29,8 @@ class RunCommand(CommandExtension):
     def add_arguments(self, parser, cli_name):
         arg = parser.add_argument(
             '--prefix',
-            help="Prefix command, which should go before the executable, "
-                 "like gdb or valgrind. Command must be wrapped in quotes "
+            help="Prefix command, which should go before the executable. "
+                 "Command must be wrapped in quotes if it contains spaces"
                  "(e.g. --prefix 'gdb -ex run --args').")
         arg = parser.add_argument(
             'package_name',
--- a/ros2run/ros2run/command/run.py
+++ b/ros2run/ros2run/command/run.py
@@ -26,6 +26,10 @@ class RunCommand(CommandExtension):
 
     def add_arguments(self, parser, cli_name):
         arg = parser.add_argument(
+            '--prefix', nargs=1,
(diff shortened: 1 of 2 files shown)


ros2run/ros2run/command/run.py
```diff
@@ -26,6 +26,10 @@ class RunCommand(CommandExtension):
 
     def add_arguments(self, parser, cli_name):
         arg = parser.add_argument(
+            '--prefix', nargs=1,
+            help="Prefix command, which should go before the executable, "
+                 "like gdb or valgrind (e.g. --prefix \'gdb -ex run --args\').")
```
wjwwood: Why do the `'` need to be escaped here?

ros2run/ros2run/command/run.py
```diff
@@ -56,4 +60,5 @@ def main(self, *, parser, args):
             raise RuntimeError(msg)
         if path is None:
             return 'No executable found'
-        return run_executable(path=path, argv=args.argv)
+        prefix=args.prefix[0] if args.prefix is not None else ''
```
wjwwood: nitpick: PEP8 needs whitespace around the `=` (this will also cause our tests to fail)

wjwwood: Also, why does `--prefix` need to be a list of size 1? Could you not just drop the `nargs=1` and access it as `args.prefix`?

wjwwood: You could also use the `nargs=*` and require it to be at the end or to be terminated with a `--` (otherwise it will greedily consume the pkg and exec args too).

dirk-thomas: The argument should either be a single argument wrapped in quotes or accept multiple arguments separated by spaces. In the first case calling `str.split` isn't enough since the argument might contained quotes substrings. Therefore you will need to use `shlex.split` instead.

dhood:
not sure if others feel this way, but the use of "prefix" here is quite close to the use of prefix in the sense of ament package prefix (e.g. [get_prefix_path in ros2pkg](https://github.com/ros2/ros2cli/blob/5584754399cb388b479814328dbf8840483ff985/ros2pkg/ros2pkg/api/__init__.py#L32))

This could mislead users, especially if we ever wanted to add an option to `ros2 run` which allows for specifying the package prefix of an executable (probably we'd just want people to reorder their `AMENT_PREFIX_PATH`, but still)

What if we rename this argument to "prefix-command" or something?

wjwwood:
I could see where the confusion could come from, but it should be pretty obvious when they give a path, and the shell tries to execute it, that it doesn't work that way.

I'd actually prefer to see the other option be more specific when/if it is implemented, something like `--ament-prefix-path`. I'd vote for this one being the `--prefix` option since that's what it is in ROS 1 with `rosrun --prefix ...`.

dhood:
OK, happy to concede this one for easier ROS 1 compatibility: `--prefix` is fine

ros2run/ros2run/command/run.py
```diff
@@ -26,6 +28,11 @@ class RunCommand(CommandExtension):
 
     def add_arguments(self, parser, cli_name):
         arg = parser.add_argument(
+            '--prefix',
+            help="Prefix command, which should go before the executable, "
+                 "like gdb or valgrind. Command must be wrapped in quotes "
```
dirk-thomas: Maybe rephrase: `Command must be wrapped in quotes if it contains spaces`?

ros2run/ros2run/command/run.py
```diff
@@ -26,6 +28,11 @@ class RunCommand(CommandExtension):
 
     def add_arguments(self, parser, cli_name):
         arg = parser.add_argument(
+            '--prefix',
+            help="Prefix command, which should go before the executable. "
+                 "Command must be wrapped in quotes if it contains spaces"
```
dirk-thomas: This needs a space before the closing quote (to have spacing before the opening parenthesis).

volkandre:
I hope I got it right now. Thanks for your feedback and your patience. I'm still learning... ;-)

wjwwood:
@volkandre it lgtm, I started some CI for you (mostly just to check for linters errors and stuff like that):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2920/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/394/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2363/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3046/)

mikaelarguedas:
I don't know if it was an expected side effect but adding this argument prevents autocomplete for packages that have only one executable (because now there are 2 choices: `<EXECUTABLE_NAME>` or `--prefix`). Not sure if/what we want to do about it but I think it was the reason we "hid" the `-h` and `--help` originally to be close to the behavior of ROS1's rosrun (`rosrun rv<TAB><TAB>` => `rosrun rviz rviz`)

dirk-thomas:
I don't think this was considered when the `--prefix` options was added. I agree that for completion it would be beneficial to not show completion for `--prefix` in order to make the completion of the executable name more convenient. Please see #43.

PR #43, opened 2017-08-15
run: suppress prefix completion

dirk-thomas:
Only "works" once kislyuk/argcomplete#224 becomes available.

Otherwise the code will gracefully fall back and keep the behavior as-is.

### DIFF
--- a/ros2run/ros2run/command/run.py
+++ b/ros2run/ros2run/command/run.py
@@ -32,6 +32,12 @@ class RunCommand(CommandExtension):
             help="Prefix command, which should go before the executable. "
                  "Command must be wrapped in quotes if it contains spaces "
                  "(e.g. --prefix 'gdb -ex run --args').")
+        try:
+            from argcomplete.completers import SuppressCompleter
+        except ImportError:
+            pass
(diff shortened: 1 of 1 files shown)


ros2run/ros2run/command/run.py
```diff
@@ -32,6 +32,12 @@ def add_arguments(self, parser, cli_name):
             help="Prefix command, which should go before the executable. "
                  "Command must be wrapped in quotes if it contains spaces "
                  "(e.g. --prefix 'gdb -ex run --args').")
+        try:
+            from argcomplete.completers import SuppressCompleter
+        except ImportError:
+            raise
```
mikaelarguedas: Maybe I'm missing something here, but if we raise here, how do "gracefully fall back"

dirk-thomas: Good catch. That was a left over from testing. Fixed in 69ba53a5fd8613ed2a709362ba4f6bb63adf8d0e.

dirk-thomas:
The feature is now available in the latest release of `argcomplete`: 1.9.2

PR #46, opened 2017-08-30
update style to satisfy new flake8 plugins

dirk-thomas:
Connect to ros2/ci#98.

### DIFF
--- a/ros2run/ros2run/command/run.py
+++ b/ros2run/ros2run/command/run.py
@@ -29,8 +29,8 @@ class RunCommand(CommandExtension):
     def add_arguments(self, parser, cli_name):
         arg = parser.add_argument(
             '--prefix',
-            help="Prefix command, which should go before the executable. "
-                 "Command must be wrapped in quotes if it contains spaces "
+            help='Prefix command, which should go before the executable. '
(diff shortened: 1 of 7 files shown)


ros2run/ros2run/command/run.py
```diff
@@ -29,8 +29,8 @@ class RunCommand(CommandExtension):
     def add_arguments(self, parser, cli_name):
         arg = parser.add_argument(
             '--prefix',
-            help="Prefix command, which should go before the executable. "
-                 "Command must be wrapped in quotes if it contains spaces "
+            help='Prefix command, which should go before the executable. '
+                 'Command must be wrapped in quotes if it contains spaces '
```
mikaelarguedas: looks a bit weird to have part of the multiline string in simple quotes and parts in double quotes. Is there an alternative allowing us to stay consistent regarding quoting within the same string?

dirk-thomas: > Is there an alternative allowing us to stay consistent regarding quoting within the same string?

No, each string is checked on its own.

mikaelarguedas: too bad, lgtm then

PR #48, opened 2017-09-11
better error message when passing an invalid value to ros2 topic pub

dirk-thomas:
Instead of showing a stacktrace with the following exception:

```
AttributeError: 'str' object has no attribute 'items'
```

this patch prints an error message ~~which includes an example value for the specified message type~~.

### DIFF
--- a/ros2topic/ros2topic/verb/pub.py
+++ b/ros2topic/ros2topic/verb/pub.py
@@ -66,6 +66,8 @@ def publisher(message_type, topic_name, values):
     module = importlib.import_module(package_name + '.msg')
     msg_module = getattr(module, message_name)
     values_dictionary = yaml.load(values)
+    if not isinstance(values_dictionary, dict):
+        return 'The passed value needs to be a dictionary in YAML format'
 
     rclpy.init()
 


dhood:
what message are you testing with? I don't know that this output is immediately useful?

```
$ ros2 topic pub /imu sensor_msgs/Imu 5
The passed value needs to be a dictionary in YAML format, e.g.:
!!python/object/apply:collections.OrderedDict
- - - header
    - !!python/object/apply:collections.OrderedDict
      - - - stamp
          - !!python/object/apply:collections.OrderedDict
            - - [sec, 0]
              - [nanosec, 0]
        - [frame_id, '']
  - - orientation
    - !!python/object/apply:collections.OrderedDict
      - - [x, 0.0]
        - [y, 0.0]
        - [z, 0.0]
        - [w, 0.0]
  - - orientation_covariance
    - [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
  - - angular_velocity
    - !!python/object/apply:collections.OrderedDict
      - - [x, 0.0]
        - [y, 0.0]
        - [z, 0.0]
  - - angular_velocity_covariance
    - [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
  - - linear_acceleration
    - !!python/object/apply:collections.OrderedDict
      - - [x, 0.0]
        - [y, 0.0]
        - [z, 0.0]
  - - linear_acceleration_covariance
    - [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
```

dirk-thomas:
I only tried a `std_msgs/String` message. For that the output looked much saner :wink:

I will put it back in progress. It can be updated after Beta 3.

mikaelarguedas:
If we iterate on this after beta3 maybe working on providing autocomplete ala ros1 for the message in yaml would workaround this issue while providing better user experience

dirk-thomas:
I updated this for now to just give a reasonable error message. An example of a valid value and/or completion can be done in the future to further improve the user experience.

Back in review.

ros2topic/ros2topic/verb/pub.py
```diff
@@ -66,6 +66,8 @@ def publisher(message_type, topic_name, values):
     module = importlib.import_module(package_name + '.msg')
     msg_module = getattr(module, message_name)
     values_dictionary = yaml.load(values)
+    if not isinstance(values_dictionary, dict):
+        return 'The passed value needs to be a dictionary in YAML format'
```
dhood: is it appropriate to return the error message here instead of raise?

dirk-thomas: Because the return value of this function is also the return value of `main`.

A few lines below (line 82) it does the same.

dirk-thomas:
> Right now if we copy past the value from the help this message is returned

Can you please post the exact command you are invoking. For me the following command (as suggested by the help text) does work fine:

```
ros2 topic pub /chatter std_msgs/String "data: Hello World"
```

mikaelarguedas:
Nvm I was missing a space because of the line break in the help...
`ros2 topic pub /chatter std_msgs/String "data:Hello World"` fails but
`ros2 topic pub /chatter std_msgs/String "data: Hello World"` works

If taking a slightly more complex example:
`ros2 topic pub /chatter test_msgs/DynamicArrayPrimitives "bool_values:[True, False]"` fails but
`ros2 topic pub /chatter test_msgs/DynamicArrayPrimitives "{bool_values: [True, False]}"` passes. That's why I thought we needed a dictionary notation.

I guess we just need to make sure people don't forget to add a space between each key / value

PR #57, opened 2017-09-27
Support non-absolute topic names.

### LINKED ISSUE #54: ros2 topic echo cannot determine type for a non FQN topic
## Bug report

**Required Info:**

- Operating System:
  - Windows 10
- Installation type:
  - http://ci.ros2.org/view/packaging/job/packaging_windows/732/
- Version or commit hash:
  - def304b4270b6b935b7d6fbbf7671634c2532452
- DDS implementation:
  - Fast-RTPS
  - N/A

#### Steps to reproduce issue
In one terminal
```
ros2 topic pub /test_topic std_msgs/String "{data: Hello world}"
```

In another
```
ros2 topic echo test_topic
```

#### Expected behavior
I'd expect a subscription to `/test_topic`

#### Actual behavior
```
C:\Users\osrf>ros2 topic echo test_topic
Could not determine the type for the passed topic
```

#### Additional information

Giving a fully qualified name works
```
C:\Users\osrf>ros2 topic echo /test_topic
data: Hello world
```

I'd expect `test_topic` to be expanded to `/test_topic` and have the same behavior as if a FQN had been provided.
http://design.ros2.org/articles/topic_and_service_names.html

clalancette:
If the user passes "/topic_name" to the ros2 echo
command, it works properly.  If they pass "topic_name"
to the ros2 echo command, it fails to match.  This
change just allows us to deal with non-absolute topic
names.

This should fix #54 .  I ran the tests against the ros2cli package locally,
with no errors.  No CI since this is python code and it won't tell us very much.

Signed-off-by: Chris Lalancette <clalancette@gmail.com>

### DIFF
--- a/ros2topic/ros2topic/verb/echo.py
+++ b/ros2topic/ros2topic/verb/echo.py
@@ -18,8 +18,9 @@ import importlib
 import sys
 
 import rclpy
-from rclpy.qos import qos_profile_sensor_data
 from rclpy.expand_topic_name import expand_topic_name
+from rclpy.qos import qos_profile_sensor_data
+from rclpy.validate_full_topic_name import validate_full_topic_name
 from ros2cli.node.direct import DirectNode
 from ros2topic.api import get_topic_names_and_types
 from ros2topic.api import TopicNameCompleter
@@ -99,6 +100,7 @@ def subscriber(node, topic_name, message_type, callback):
     if message_type is None:
         topic_names_and_types = get_topic_names_and_types(node=node)
         expanded_name = expand_topic_name(topic_name, node.get_name(), node.get_namespace())
+        validate_full_topic_name(expanded_name, is_service=False)
         for n, t in topic_names_and_types:
             if n == expanded_name:
                 if len(t) > 1:
--- a/ros2topic/ros2topic/verb/echo.py
+++ b/ros2topic/ros2topic/verb/echo.py
@@ -100,7 +100,7 @@ def subscriber(node, topic_name, message_type, callback):
     if message_type is None:
         topic_names_and_types = get_topic_names_and_types(node=node)
         expanded_name = expand_topic_name(topic_name, node.get_name(), node.get_namespace())
-        validate_full_topic_name(expanded_name, is_service=False)
+        validate_full_topic_name(expanded_name)
         for n, t in topic_names_and_types:
             if n == expanded_name:
                 if len(t) > 1:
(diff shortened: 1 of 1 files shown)


wjwwood:
Why not use https://github.com/ros2/rclpy/blob/master/rclpy/rclpy/expand_topic_name.py#L18?

wjwwood:
I think it would be better, since it would respect namespaces as well.

clalancette:
Only because I didn't know about it :).  Thanks, I'll give that a try.

clalancette:
I added in the validation as @wjwwood suggested.  I'm also running a CI job, just on Linux, to make sure there aren't any lingering linting/pep8 errors: [Build Status](http://ci.ros2.org/job/ci_linux/3297/).  Assuming that is green, I'm going to go ahead and merge this.

ros2topic/ros2topic/verb/echo.py
```diff
@@ -97,8 +99,10 @@ def represent_ordereddict(dumper, data):
 def subscriber(node, topic_name, message_type, callback):
     if message_type is None:
         topic_names_and_types = get_topic_names_and_types(node=node)
+        expanded_name = expand_topic_name(topic_name, node.get_name(), node.get_namespace())
+        validate_full_topic_name(expanded_name, is_service=False)
```
mikaelarguedas: Nit: `False` being the default you shouldn't need to have to pass the `is_service` keyword

clalancette: Good call.  Will fix that before merging.

clalancette:
CI is green, I fixed the small problem, and it's approved.  Merging.

ros2topic/ros2topic/verb/echo.py
```diff
@@ -97,8 +99,10 @@ def represent_ordereddict(dumper, data):
 def subscriber(node, topic_name, message_type, callback):
     if message_type is None:
         topic_names_and_types = get_topic_names_and_types(node=node)
+        expanded_name = expand_topic_name(topic_name, node.get_name(), node.get_namespace())
+        validate_full_topic_name(expanded_name)
```
dirk-thomas: This function might raise an `InvalidTopicNameException` which I assume results in printing a stacktrace which is not a good user experience. The possible exceptions should be catched and a "nicer" error message (without a stacktrace) should be printed instead.

wjwwood: Yeah, when I made that suggestion I meant you could raise it here, catch it and do something nicer. If you just want the traceback, then that will happen when you try to create a pub or sub with the topic.

clalancette: Ah, OK.  I'll open a new PR with something nicer.

clalancette: #58

PR #58, opened 2017-10-05
Make sure to check errors when expanding the topic name.

clalancette:
We need to catch ValueErrors when actually doing the expansion,
then InvalidTopicNameException when doing the validation.

### DIFF
--- a/ros2topic/ros2topic/verb/echo.py
+++ b/ros2topic/ros2topic/verb/echo.py
@@ -102,7 +102,7 @@ def subscriber(node, topic_name, message_type, callback):
         try:
             expanded_name = expand_topic_name(topic_name, node.get_name(), node.get_namespace())
         except ValueError as e:
-            raise RuntimeError("Topic name '%s' is invalid" % (topic_name))
+            raise RuntimeError(e)
         try:
             validate_full_topic_name(expanded_name)
         except rclpy.exceptions.InvalidTopicNameException as e:
--- a/ros2topic/ros2topic/verb/echo.py
+++ b/ros2topic/ros2topic/verb/echo.py
@@ -99,8 +99,14 @@ def represent_ordereddict(dumper, data):
 def subscriber(node, topic_name, message_type, callback):
     if message_type is None:
         topic_names_and_types = get_topic_names_and_types(node=node)
-        expanded_name = expand_topic_name(topic_name, node.get_name(), node.get_namespace())
-        validate_full_topic_name(expanded_name)
+        try:
+            expanded_name = expand_topic_name(topic_name, node.get_name(), node.get_namespace())
+        except ValueError as e:
+            raise RuntimeError("Topic name '%s' is invalid" % (topic_name))
+        try:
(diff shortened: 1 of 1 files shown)


ros2topic/ros2topic/verb/echo.py
```diff
@@ -99,8 +99,14 @@ def represent_ordereddict(dumper, data):
 def subscriber(node, topic_name, message_type, callback):
     if message_type is None:
         topic_names_and_types = get_topic_names_and_types(node=node)
-        expanded_name = expand_topic_name(topic_name, node.get_name(), node.get_namespace())
-        validate_full_topic_name(expanded_name)
+        try:
+            expanded_name = expand_topic_name(topic_name, node.get_name(), node.get_namespace())
+        except ValueError as e:
+            raise RuntimeError("Topic name '%s' is invalid" % (topic_name))
```
dirk-thomas: The docblock of the function says the following which doesn't seem to match the error message:

> :raises: ValueError if the topic name, node name or namespace are invalid

sloretz: Does this `RuntimeError` get handled elsewhere, or does the user see the `RuntimeError` traceback? If the latter it looks like the `ValueError` raised by rclpy already includes a similar message. How about writing that message, `e.args[0]`, to stderr instead?

https://github.com/ros2/rclpy/blob/a6daedbde500a1dbd222c1f1851e1ae2aea1f2e1/rclpy/src/rclpy/_rclpy.c#L519-L520

dirk-thomas: The `RuntimeError` is being used to signal an error where the user only needs to see the message but not the stack trace: see https://github.com/ros2/ros2cli/blob/ce77579d2fd8cbdc27352b67ff3fcf67a2369209/ros2cli/ros2cli/cli.py#L67-L68

By handling this in a single place it allows to easily change the behavior in the future. That would not be the case if every code directly decides what to do.

clalancette: A `RuntimeError` gets handled by the higher layers of ros2cli and prints out a nice error message.  My original implementation here did what @sloretz proposed, but the problem is that the error message it actually generates is kind of ugly:

```
topic name '/test_topic/' is invalid: topic name is invalid, at /home/clalancette/ros2_ws/src/ros2/rcl/rcl/src/rcl/expand_topic_name.c:67
```

That being said, if we think that is a better way to go, I'll change it to that.

wjwwood: I agree with @dirk-thomas that the value error doesn't imply topic is bad only. Other than that I think the runtime error pattern is fine.

If you want to make the error message pretty then you can reformat it when putting it into the runtime error. Or if it's a general improvement to the error message, you could make it in rclpy's function.

clalancette: All right, I updated this to use the original exception instead.  I'll do a more general cleanup to that error message in another PR.

clalancette:
Thanks, merging.

Issue #40, opened 2017-07-11
Run arguments do not get passed to node executable


bponsler:
When attempting to pass command line arguments to a node using the **ros2 run [pkg] [exec]** syntax, the arguments get processed by the ros2 executable rather than the node executable being launched.

Example:

```
$ ros2 run ros1_bridge dynamic_bridge --help
usage: ros2 run [-h] package_name executable_name [argv [argv ...]]

Run a package specific executable

positional arguments:
  package_name     Name of the ROS package
  executable_name  Name of the executable
  argv             Pass arbitrary arguments to the executable (use '--' before
                   these arguments to ensure they are not handle by this
                   command)

optional arguments:
  -h, --help       show this help message and exit
```

Expected output:

```
$ ./install/lib/ros1_bridge/dynamic_bridge --help
Usage:
 -h, --help: This message.
 --show-introspection: Print output of introspection of both sides of the bridge.
 --print-pairs: Print a list of the supported ROS 2 <=> ROS 1 conversion pairs.
 --bridge-all-topics: Bridge all topics in both directions, whether or not there is a matching subscriber.
 --bridge-all-1to2-topics: Bridge all ROS 1 topics to ROS 2, whether or not there is a matching subscriber.
 --bridge-all-2to1-topics: Bridge all ROS 2 topics to ROS 1, whether or not there is a matching subscriber.
```

Although I'm sure you get the picture, here's another example:

```
$ ros2 run ros1_bridge dynamic_bridge --bridge-all-topics
usage: ros2 [-h] Call `ros2 <command> -h` for more detailed usage. ...
ros2: error: unrecognized arguments: --bridge-all-topics
```

Add one passing just a flag:

```
$ ros2 run ros1_bridge dynamic_bridge -f
usage: ros2 [-h] Call `ros2 <command> -h` for more detailed usage. ...
ros2: error: unrecognized arguments: -f
```

Arguments that are not prefixed with one or more dashes do allow the node to get executed (although I don't have evidence at the moment that they successfully get passed to the node).

E.g., (I don't have a master running, but it runs the bridge)

```
$ ros2 run ros1_bridge dynamic_bridge this_is_an_argument another arg3
[ERROR] [1499788917.336544285] [/ros_bridge]: [registerPublisher] Failed to contact master at [localhost:11311].  Retrying...
```

wjwwood:
From the `--help`:

```
argv             Pass arbitrary arguments to the executable (use '--' before
[...]

PR #61, opened 2017-11-14
Improve parameters passing to node in ros2run

aeudes:
This pull request try to improve on #40.
The use of narg=REMAINDER in [argparse](https://docs.python.org/3.5/library/argparse.html#nargs) exhibit the wanted behavior where every parameters after the node name are passed to the node without the need of the -- separator.
The prefix parameter (--prefix) still need to be placed before package and node to not be passed to the node. May be it is useful to print a warning if the prefix is had at the end of the command line ?

Let me know if i have missed something here.

### DIFF
--- a/ros2run/ros2run/command/run.py
+++ b/ros2run/ros2run/command/run.py
@@ -23,6 +23,7 @@ from ros2run.api import MultipleExecutables
 from ros2run.api import run_executable
 from argparse import REMAINDER
 
+
 class RunCommand(CommandExtension):
     """Run a package specific executable."""
 
--- a/ros2run/ros2run/command/run.py
+++ b/ros2run/ros2run/command/run.py
@@ -21,7 +21,7 @@ from ros2run.api import ExecutableNameCompleter
 from ros2run.api import get_executable_path
 from ros2run.api import MultipleExecutables
 from ros2run.api import run_executable
-
+from argparse import REMAINDER
 
 class RunCommand(CommandExtension):
     """Run a package specific executable."""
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Thank you for the patch. Can you please run the tests of that package locally and address the linter warnings. Thanks.

aeudes:
Done.

dirk-thomas:
I am still getting the following two warnings:

```
./ros2run/command/run.py:24:1: I100 Import statements are in the wrong order. from argparse should be before from ros2run.api
from argparse import REMAINDER
^
```

```
./ros2run/command/run.py:53:18: Q000 Remove bad quotes.
            help="Pass arbitrary arguments to the executable")
                 ^
```

mikaelarguedas:
@aeudes my guess is that you don't get these warnings locally because these are flake8 plugins not provided by the ubuntu package `python-flake8`
Line from the instructions to install the testing dependencies:
`sudo pip3 install flake8 flake8-blind-except flake8-builtins flake8-class-newline flake8-comprehensions flake8-deprecated flake8-docstrings flake8-import-order flake8-quotes pytest pytest-cov pytest-runner`

aeudes:
Sorry about that @dirk-thomas but I have only one warning on my setup. 
Thanks  @mikaelarguedas but when i install the dependency, it create a python error (the traceback point to pep8.py).
So for now I have just fix the two remaining warning (I hope). I will try to understand later what's wrong with my setup.

dirk-thomas:
@aeudes No problem. Thank you for the quick iteration. I moved the `argparse` import a bit higher to pass the linters (ba69f4e0f5d8eda49387be74cb179ab530326f10).

This is indeed a nice improvement. :+1:

mikaelarguedas:
We need to keep in mind that all tutorials should be updated accordingly when we tag the new release

dirk-thomas:
> We need to keep in mind that all tutorials should be updated accordingly when we tag the new release

I think the current instructions (recommending to separate the args with `--`) will continue to work.

mikaelarguedas:
> I think the current instructions (recommending to separate the args with --) will continue to work.

Yes they will (tested on Linux)

PR #64, opened 2017-12-03
add type completer for 'topic pub' and 'service call'

### LINKED ISSUE #63: [ros2 topic pub / ros2 service call] Autocomplete message/service type
#### Feature description

Right now these tools autopopulate the topic / service name based on the discovered topics / services, but dont populate the type.
The structure used (topic/service_names_and_types) also provides the types so we should autocomplete the type matching the selected name.

If no type matching the name is found in the structure we should apply the same logic as the service/message completer used for ros2msg/ros2srv to complete based on the interfaces installed on the system

dirk-thomas:
Connect to #63. Fixes #63.

### DIFF
--- a/ros2service/ros2service/api/__init__.py
+++ b/ros2service/ros2service/api/__init__.py
@@ -14,6 +14,7 @@
 
 from rclpy.topic_or_service_is_hidden import topic_or_service_is_hidden
 from ros2cli.node.strategy import NodeStrategy
+from ros2srv.api import service_type_completer
 
 
 def get_service_names_and_types(*, node, include_hidden_services=False):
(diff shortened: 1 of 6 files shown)


mikaelarguedas:
code change looks good overall. Though it would be great to hide the `--repeat` option if only a single service type matches.
e.g.
```
$ ros2 service call /add_two_ints <TAB><TAB> 
example_interfaces/AddTwoInts  -r  --repeat
```

I'd expect this to just complete the matching service name

dirk-thomas:
> Though it would be great to hide the `--repeat` option if only a single service type matches.

I don't think that is technically possible (the results of one completer deciding to ignore the results of other completers). Anyway I wouldn't consider this within the scope of this PR.

mikaelarguedas:
fair enough. Then maybe for another PR we can consider providing the same argument to `ros2 topic pub` ? Because it's a bit surprising from the user side to have a different behavior

dirk-thomas:
> Then maybe for another PR we can consider providing the same argument to `ros2 topic pub` ?

Sure, `topic pub` currently publishes with a fixed rate of once every second. A pull request to make this configurable would be great.

ros2service/ros2service/api/__init__.py
```diff
@@ -43,3 +44,21 @@ def __call__(self, prefix, parsed_args, **kwargs):
                 node=node,
                 include_hidden_services=getattr(
                     parsed_args, self.include_hidden_services_key))
+
+
+class ServiceTypeCompleter:
+    """Callable returning an existing service type or all service types."""
```
mikaelarguedas: Nit: "an existing service type" I would recommend the documentation that it's returning the service type matching the passed service_name or all types (same for topics below)

dirk-thomas: That sounds good but exceed the one line rule for the docblock :wink:

PR #66, opened 2017-12-03
[ros2topic] pub: add --repeat

mikaelarguedas:
I kept the default behavior (publishing every second) the same for now as we dont have latching.

Not sure what the right place is to put the `float_or_int` utility for reuse. I'm not convinced it belongs in the Base class but I can imagine it being reused in many places. Maybe a set of argparse utilities in ros2cli would make sense. 
@dirk-thomas thoughts?

### DIFF
--- a/ros2service/ros2service/verb/call.py
+++ b/ros2service/ros2service/verb/call.py
@@ -44,7 +44,7 @@ class CallVerb(VerbExtension):
                  '(e.g. "{a: 1, b: 2}"), ' +
                  'otherwise the service request will be published with default values')
         parser.add_argument(
-            '-r', '--rate', metavar='N', type=float, default=1,
+            '-r', '--rate', metavar='N', type=float, default=1.0,
             help='Calling rate in Hz (default: 1)')
         parser.add_argument(
             '-1', '--once', action='store_true',
@@ -55,7 +55,7 @@ class CallVerb(VerbExtension):
             raise RuntimeError('rate must be greater than zero')
 
         return requester(
-            args.service_type, args.service_name, args.values, 1 / args.rate, args.once)
+            args.service_type, args.service_name, args.values, 1. / args.rate, args.once)
 
 
 def requester(service_type, service_name, values, period, once):
--- a/ros2service/ros2service/verb/call.py
+++ b/ros2service/ros2service/verb/call.py
@@ -45,7 +45,7 @@ class CallVerb(VerbExtension):
                  'otherwise the service request will be published with default values')
         parser.add_argument(
             '-r', '--rate', metavar='N', type=float, default=1,
-            help='Calling rate (hz) default to 1')
+            help='Calling rate in Hz (default: 1)')
         parser.add_argument(
             '-1', '--once', action='store_true',
             help='Call service once and exit')
--- a/ros2service/ros2service/verb/call.py
+++ b/ros2service/ros2service/verb/call.py
@@ -44,23 +44,31 @@ class CallVerb(VerbExtension):
             help='Values to fill the service request with in YAML format ' +
                  '(e.g. "{a: 1, b: 2}"), ' +
                  'otherwise the service request will be published with default values')
-
-        def float_or_int(input_string):
-            try:
-                value = float(input_string)
-            except ValueError:
-                raise argparse.ArgumentTypeError("'%s' is not an integer or float" % input_string)
-            return value
-
         parser.add_argument(
-            '-r', '--repeat', metavar='N', type=float_or_int,
-            help='Repeat the call every N seconds')
+            '-r', '--rate', metavar='N', type=float,
+            help='Calling rate (hz) default to 1')
+        parser.add_argument(
+            '-1', '--once', action='store_true',
+            help='Cal service once and exit')
 
     def main(self, *, args):
-        return requester(args.service_type, args.service_name, args.values, args.repeat)
+        return requester(args.service_type, args.service_name, args.values, args.rate, args.once)
 
 
-def requester(service_type, service_name, values, repeat):
+def requester(service_type, service_name, values, rate, once):
+    if rate is not None:
+        if once:
+            raise RuntimeError('You cannot select both -r and -1 (--once)')
+        try:
+            rate = float(rate)
+        except ValueError:
+            raise argparse.ArgumentTypeError('rate must be a number')
+        if rate <= 0:
+            raise ValueError('rate must be greater than zero')
+    if rate is not None:
+        period = 1. / rate
+    else:
+        period = 1
     # TODO(wjwwood) this logic should come from a rosidl related package
     try:
         package_name, srv_name = service_type.split('/', 2)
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Instead of adding `--repeat` I would suggest to add options similar to the ones available in [ROS 1](https://github.com/ros/ros_comm/blob/f7c7b3c28a09feeee643c10c6a8c4aff7684dda7/tools/rostopic/src/rostopic/__init__.py#L1719-L1722).

mikaelarguedas:
thanks I'll do that. I guess I'll adapt `ros2 service call` accordingly.

> Not sure what the right place is to put the float_or_int utility for reuse. I'm not convinced it belongs in the Base class but I can imagine it being reused in many places. Maybe a set of argparse utilities in ros2cli would make sense.

Do you have any feedback on this?

dirk-thomas:
> Not sure what the right place is to put the float_or_int utility for reuse. I'm not convinced it belongs in the Base class but I can imagine it being reused in many places. Maybe a set of argparse utilities in ros2cli would make sense.

In general: the function seems to be so trivial that I wouldn't bother to declare it as a public API (which needs to be documented and tested at some point).

In this specific case it might be possible to simply use `type=float` since that is what this custom function is doing anyway. I haven't tried that though...

ros2service/ros2service/verb/call.py
```diff
@@ -44,23 +44,31 @@ def add_arguments(self, parser, cli_name):
...
+
+
+def requester(service_type, service_name, values, rate, once):
+    if rate is not None:
+        if once:
+            raise RuntimeError('You cannot select both -r and -1 (--once)')
+        try:
+            rate = float(rate)
+        except ValueError:
+            raise argparse.ArgumentTypeError('rate must be a number')
```
dirk-thomas: This is unnecessary. Since the argument is of `type=float` the variable `args.rate` is already of type `float` (or `None`).

ros2service/ros2service/verb/call.py
```diff
@@ -44,23 +44,31 @@ def add_arguments(self, parser, cli_name):
...
-
-
-def requester(service_type, service_name, values, repeat):
+        return requester(args.service_type, args.service_name, args.values, args.rate, args.once)
+
+
+def requester(service_type, service_name, values, rate, once):
+    if rate is not None:
+        if once:
+            raise RuntimeError('You cannot select both -r and -1 (--once)')
```
dirk-thomas: This argument checking should happen within argparse. You can use a mutually exclusive group for this.

ros2service/ros2service/verb/call.py
```diff
@@ -44,23 +44,31 @@ def add_arguments(self, parser, cli_name):
...
+        try:
+            rate = float(rate)
+        except ValueError:
+            raise argparse.ArgumentTypeError('rate must be a number')
+        if rate <= 0:
+            raise ValueError('rate must be greater than zero')
+    if rate is not None:
+        period = 1. / rate
+    else:
+        period = 1
```
dirk-thomas: I would suggest to use a default value for the argument instead. That way the code doesn't have to contain some not-documented constants.

That might affect the choice to group the arguments...

ros2service/ros2service/verb/call.py
mikaelarguedas: Will fix

ros2service/ros2service/verb/call.py
mikaelarguedas: Yeah I went back and forth on this.
I'm fine with adding the default vaIue for the rate and keeping the logic for the mutual exclusion. That logic can be in a `main` fuction before calling the actual requester like it's done for ros2topic. 

TBH I think that they dont have to be mutually exclusive. If you specify a rate you override a default of 1Hz, but if you specify `--once`, the rate (user specified or not) will not matter as we exit after the first call. I'm not convinced the executable should raise, it could just honor `once`

ros2topic/ros2topic/verb/pub.py
```diff
@@ -43,16 +43,25 @@ def add_arguments(self, parser, cli_name):
...
+            help='Publish one message and exit')
 
     def main(self, *, args):
         return main(args)
 
 
 def main(args):
-    return publisher(args.message_type, args.topic_name, args.values)
+    if args.rate <= 0:
+        raise ValueError('rate must be greater than zero')
```
dirk-thomas: I would suggest ti move this into `PubVerb.main`. That way the entity defining the command line arguments is also responsible of checking the sanity of the values. (The same way as it is for the `CallVerb` above.)

ros2service/ros2service/verb/call.py
```diff
@@ -44,23 +43,25 @@ def add_arguments(self, parser, cli_name):
...
+            raise ValueError('rate must be greater than zero')
+
+        return requester(args.service_type, args.service_name, args.values, args.rate, args.once)
 
 
-def requester(service_type, service_name, values, repeat):
+def requester(service_type, service_name, values, rate, once):
+    if rate is not None:
+        period = 1. / rate
+    else:
```
dirk-thomas: This is now never the case.

ros2topic/ros2topic/verb/pub.py
dirk-thomas: Does the ValueError result in a stack trace for the user? If so I would recommend to either use argparse to signal the error or raise a RuntimeError (both with the goal to not show a stack trace to the user).

ros2service/ros2service/verb/call.py
mikaelarguedas: forgot to commit local changes :/
https://github.com/ros2/ros2cli/pull/66/commits/29015d6be6193997f8d9179322e4f4957e49e05d

ros2topic/ros2topic/verb/pub.py
mikaelarguedas: Good point, I changed it to a RuntimeError

mikaelarguedas: > I would suggest ti move this into PubVerb.main. That way the entity defining the command line arguments is also responsible of checking the sanity of the values. (The same way as it is for the CallVerb above.)

Good point, moved in https://github.com/ros2/ros2cli/pull/66/commits/041b4111122b42a1c5dadebcff5dcca9d91e09b1

mikaelarguedas:
Thanks for the various fixups.
I tried the various scenarios it seems to work fine.
One thing that's a bit weird is that if the use passes an invalid rate and also pass `once` it raises. I could make it perform the check on the rate being only if `once` is not passed if we wanted to be more correct.

dirk-thomas:
> One thing that's a bit weird is that if the use passes an invalid rate and also pass `once` it raises. I could make it perform the check on the rate being only if `once` is not passed if we wanted to be more correct.

If we would check the argument directly within argparse that would also fail for invalid values even when they are later not used. So I am ok with it raising in that case.

dhood:
what's the motivation for having repeated service calls the default? just to match ros2 topic? Repeated service calls wasn't the case in ros1 and I don't think it's appropriate as the default for ros2

mikaelarguedas:
It was not the case for rostopic in ROS 1 either, we did it for consistency but I agree the default should likely be to send only once. I'm fine to change it if everybody agrees

mikaelarguedas:
#67

PR #67, opened 2017-12-06
[ros2service] call only once by default

mikaelarguedas:


dirk-thomas:
Calling a service repeatedly as fast as possible sounds like a reasonable use case to me. Therefore I wouldn't use that to represent "once".

Instead I would suggest to remove the default value of the `--rate` argument (which makes it `None`). And pass `1. / args.rate if args.rate else None` to the `requester` function (removing the `once` argument).

mikaelarguedas:
Fair enough, updated accordingly

PR #79, opened 2018-02-12
fix import order

mikaelarguedas:
new version of flake8 complains as xmlrpc is considered a system header and thus needs to be imported in the previous block.

cc @wjwwood as build cop


mikaelarguedas:
Before the patch:
* http://ci.ros2.org/view/nightly/job/nightly_linux_release/717/testReport/

With this patch:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3953/)

PR #81, opened 2018-02-22
print full help when no command is passed

### LINKED ISSUE #74: Calling ros2 cli tool without arguments throws an error instead of listing available commands
Calling `ros2` cli tool without arguments throws an error instead of listing available commands.

## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 16.04
- Installation type:
  - binary
- Version or commit hash:
  - Ardent

#### Steps to reproduce issue
```
$ ros2
```

#### Expected behavior

```
$ ros2
usage: ros2 [-h] Call `ros2 <command> -h` for more detailed usage. ...
<<<LIST OF COMMANDS>>>
```

#### Actual behavior

```
$ ros2
usage: ros2 [-h] Call `ros2 <command> -h` for more detailed usage. ...
ros2: error: the following arguments are required: Call `ros2 <command> -h` for more detailed usage.
```

#### Bug Source

It looks like this is being passed as the `metavar` instead of the help text to the argument parser.
https://github.com/ros2/ros2cli/blob/a4e00b2e3846db9e7b64431110f81d02f6efa0ee/ros2cli/ros2cli/command/__init__.py#L92-L95

dirk-thomas:
When no `command` / `verb` is passed the full help is printed as if the user would have passed `--help`.

Fixes #74.

The API is backward compatible so that the behavior of existing extensions isn't changes. Extensions implemented in other repositories could be updated in the same way (not to require the `verb`) if they want to show the full help text in that case.

### DIFF
--- a/ros2cli/ros2cli/command/daemon.py
+++ b/ros2cli/ros2cli/command/daemon.py
@@ -22,13 +22,18 @@ class DaemonCommand(CommandExtension):
     """Various daemon related sub-commands."""
 
     def add_arguments(self, parser, cli_name):
+        self._subparser = parser
         # get verb extensions and let them add their arguments
         verb_extensions = get_verb_extensions('ros2cli.daemon.verb')
-        add_subparsers(parser, cli_name, '_verb', verb_extensions)
+        add_subparsers(
+            parser, cli_name, '_verb', verb_extensions, required=False)
 
     def main(self, *, parser, args):
-        # the attribute should always exist
-        # otherwise argparse should have exited
+        if not hasattr(args, '_verb'):
+            # in case no verb was passed
+            self._subparser.print_help()
+            return 0
+
         extension = getattr(args, '_verb')
 
         # call the verb's main method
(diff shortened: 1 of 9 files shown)


ros2cli/ros2cli/command/daemon.py
```diff
@@ -22,11 +22,18 @@ class DaemonCommand(CommandExtension):
...
+            parser, cli_name, '_verb', verb_extensions, required=False)
 
     def main(self, *, parser, args):
+        if not hasattr(args, '_verb'):
+            # in case no verb was passed
+            self._subparser.print_help()
+            return 0
+
         # the attribute should always exist
         # otherwise argparse should have exited
```
mikaelarguedas: I don't think this comment is necessary anymore as the attribute existence is checked just above (same everywhere below)

dirk-thomas: Removed the obsolete comments.

Issue #78, opened 2018-02-08
Debian package requires manually sourcing ros2-argcomplete.bash/zsh


dirk-thomas:
It would be good when the user sources `/opt/ros/<distro>/setup.bash` that the environment hook for the completion is implicitly included.

PR #84, opened 2018-03-26
source bash completion script from setup file

dirk-thomas:
Fixes ros2/ros2cli#78. Connect to ros2/ros2cli#78.

I have applied the same changes to the installed location in `/opt/ros/ardent` which made it work for me from Debian packages when sourcing `/opt/ros/ardent/local_setup.bash`.

### DIFF
--- a/ros2cli/resource/local_setup.bash
+++ b/ros2cli/resource/local_setup.bash
@@ -1,0 +1,37 @@
+# reduced from ament_package/template/package_level/local_setup.bash.in
+
+# provide AMENT_CURRENT_PREFIX to shell script
+AMENT_CURRENT_PREFIX=$(builtin cd "`dirname "${BASH_SOURCE[0]}"`/../.." && pwd)
+# store AMENT_CURRENT_PREFIX to restore it before each environment hook
+_package_local_setup_AMENT_CURRENT_PREFIX=$AMENT_CURRENT_PREFIX
+
+# unset AMENT_ENVIRONMENT_HOOKS
+# if not appending to them for return
+if [ -z "$AMENT_RETURN_ENVIRONMENT_HOOKS" ]; then
+  unset AMENT_ENVIRONMENT_HOOKS
+fi
+
+# restore AMENT_CURRENT_PREFIX before evaluating the environment hooks
+AMENT_CURRENT_PREFIX=$_package_local_setup_AMENT_CURRENT_PREFIX
+# list all environment hooks of this package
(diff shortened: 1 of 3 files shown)


mikaelarguedas:
Will this work only for bash and not other shells like zsh or sh ?

dirk-thomas:
`sh` doesn't have completion functionality and there is no `zsh` specific prefix-level script at the moment.

mikaelarguedas:
> and there is no `zsh` specific prefix-level script at the moment

I'm not sure I understand that bit. Can you clarify?

Is the conclusion is that zsh users will need to source `ros2-argcomplete.zsh` manually?

dirk-thomas:
> Can you clarify?

I guess my install directory doesn't have a `.zsh` file since I am not building with `ament_tools`. I will add `local_setup.zsh` file to this PR so also automate this for `zsh`.

dirk-thomas:
See a151f89b47e4b758173fbccbcfce6dc3d4556a72 for a `zsh` specific local setup file.

dirk-thomas:
I propose to cherry-pick this change to the `ardent` branch and release this repo into `ardent`. That will allow us to confirm that it actually works in the Debian case (and the users will benefit from it early). Any thoughts on this?

wjwwood:
You have to not only cherry-pick to the `ardent` branch but also make a release to see if it works in the debian case right?

I'm not opposed to it either way, but I would prefer to not have any kind of partial re-release of ardent. One related concern is that someone getting the `ardent` branch would not be getting what is in the binaries, but then again they should probably use tags instead if they want exactly what's in binaries.

dirk-thomas:
> You have to not only cherry-pick to the ardent branch but also make a release to see if it works in the debian case right?

Yes, "cherry-pick this change to the ardent branch and release this repo into ardent".

> One related concern is that someone getting the ardent branch would not be getting what is in the binaries

> Does this imply rebuilding the artifacts on all platforms to have consistent release state everywhere ?

I don't think this is necessary. This problem is specific to our Debian packages. The fat archives as well as a from-source build are already working (since `ament_tools` is taking care of sourcing the completion script). Therefore I would not release new fat archives.

mikaelarguedas:
> I don't think this is necessary. This problem is specific to our Debian packages. The fat archives as well as a from-source build are already working (since ament_tools is taking care of sourcing the completion script). Therefore I would not release new fat archives.

I'm fine with it as no actual code is touched, ti will make the behavior consistent across the board and the release overhead will be minimal

dirk-thomas:
Released 0.4.1: ros2/rosdistro#77

Issue #86, opened 2018-04-02
ros2 pkg create: resource files not installed


dhood:
How have people been testing `ros2 pkg create`? I tried to run it but the templates aren't installed so I get:

```
$ ros2 pkg create dummy_package --license asdf
going to create a new package
package name: dummy_package
destination directory: /tmp
package format: 2
version: 0.0.0
description: TODO: Package description
maintainer: ['dhood <dhood@users.noreply.github.com>']
licenses: ['asdf']
build type: ament_cmake
dependencies: []
creating folder ./dummy_package
Traceback (most recent call last):
  File "/home/dhood/ros2_ws/install_isolated/ros2cli/bin/ros2", line 11, in <module>
    load_entry_point('ros2cli==0.4.0', 'console_scripts', 'ros2')()
  File "/home/dhood/ros2_ws/install_isolated/ros2cli/lib/python3.5/site-packages/ros2cli/cli.py", line 69, in main
    rc = extension.main(parser=parser, args=args)
  File "/home/dhood/ros2_ws/install_isolated/ros2pkg/lib/python3.5/site-packages/ros2pkg/command/pkg.py", line 39, in main
    return extension.main(args=args)
  File "/home/dhood/ros2_ws/install_isolated/ros2pkg/lib/python3.5/site-packages/ros2pkg/verb/create.py", line 154, in main
    create_package_environment(package, args.destination_directory)
  File "/home/dhood/ros2_ws/install_isolated/ros2pkg/lib/python3.5/site-packages/ros2pkg/api/create.py", line 97, in create_package_environment
    package_xml_config)
  File "/home/dhood/ros2_ws/install_isolated/ros2pkg/lib/python3.5/site-packages/ros2pkg/api/create.py", line 71, in _create_template_file
    raise FileNotFoundError('template not found:', template_path)
FileNotFoundError: [Errno template not found:] /home/dhood/ros2_ws/install_isolated/ros2pkg/lib/python3.5/site-packages/ros2pkg/resource/package_environment/package.xml.em
```

Do I need to run it from a particular directory or anything? Or do we just need to install the resource files (I don't see how they are being installed currently).

I am posing this as a question because it might be user error, given that I am the first one to run into it when it seems that others were testing it out in https://github.com/ros2/ros2cli/pull/42

mikaelarguedas:
I used it recently without problem. I can use it in different locations

```
$ cd work/ros2/current_ws/
$ source install_debug_isolated/local_setup.bash 
[...]

PR #87, opened 2018-04-03
Fix installation of templates for ros2pkg create

dhood:
connects to ros2/ros2cli#86
resolves ros2/ros2cli#86

Previously (when not using symlink install), templates were being looked for in `~/ros2_ws/install_isolated/ros2pkg/lib/python3.5/site-packages/ros2pkg/`.

I didn't see how to make `data_files` install to that location (and it might not make sense conceptually) so I'm installing to the `share` directory instead. Works with symlink or nonsymlink installs now.


ros2pkg/ros2pkg/api/create.py
```diff
@@ -18,7 +18,9 @@
 
 import em
 
-RESOURCE_PATH = os.path.join(os.path.dirname(__file__), '..', 'resource')
+from ros2pkg.api import get_prefix_path
+
+RESOURCE_PATH = os.path.join(get_prefix_path('ros2pkg'), 'share', 'ros2pkg', 'resource')
```
mikaelarguedas: Is this an absolute path ?
If so, `template_path` line 71 can just get rid of `__file__` as well and just use the resource path directly

dhood: yeah I think __file__ maybe didn't belong regardless. will fix

dhood:
CI for linters (thanks for testing manually): [Build Status](https://ci.ros2.org/job/ci_windows/4275/)

ros2pkg/ros2pkg/api/create.py
dhood: https://github.com/ros2/ros2cli/pull/87/commits/1f7b17eed2fbd5bcbe21589035647a1884baf561

ros2pkg/setup.py
```diff
@@ -5,6 +5,25 @@
...
+            'resource/cpp/library.cpp.em',
+            'resource/cpp/main.cpp.em',
+            'resource/cpp/visibility_control.h.em',
+        ]),
+        ('share/ros2pkg/resource/package_environment', [
+            'resource/package_environment/package.xml.em',
+        ]),
+    ],
     install_requires=['ros2cli'],
     zip_safe=False,
```
mikaelarguedas: With `__file__` completely removed, this package may very well be zip_safe now (just I hunch, I didn't actually check)

ros2pkg/ros2pkg/api/create.py
```diff
@@ -18,7 +18,10 @@
 
 import em
 
-RESOURCE_PATH = os.path.join(os.path.dirname(__file__), '..', 'resource')
+from ros2pkg.api import get_prefix_path
+
+RESOURCE_PATH = os.path.abspath(
+    os.path.join(get_prefix_path('ros2pkg'), 'share', 'ros2pkg', 'resource'))
```
dirk-thomas: Looking at this again I don't think we should couple the resource lookup with the ament index. Instead use the  `pkg_resources` API for this: like https://github.com/ament/ament_package/blob/28308125089b607f2797abee577371d7ee9a0934/ament_package/templates.py#L24

dhood: so install templates with `package_data` instead of `data_files`, is that right? I think that's the answer to the "how do I install to site-packges" question I was having, good to know that this exists (and that we don't have to specify each file manually!)

ros2pkg/package.xml
```diff
@@ -12,6 +12,8 @@
   <depend>ros2cli</depend>
 
   <exec_depend>ament_index_python</exec_depend>
+  <exec_depend>python3-empy</exec_depend>
+  <exec_depend>python3-setuptools</exec_depend>
```
mikaelarguedas: I don't think we declare (in the package.xml) dependency on setuptools for pure python packages (the same way we don't declare a dependency on cmake for ament_cmake packages).

Not sure about listing it in the install_requires section of the setup.py, it seems that some packages do and some don't. Maybe @dirk-thomas can give more details on that point

dhood:
I put this back in review. github collapsed it but there's an unresolved question in https://github.com/ros2/ros2cli/pull/87#discussion_r178999679 if anyone has input

ros2pkg/ros2pkg/api/create.py
```diff
@@ -17,8 +17,9 @@
 import sys
 
 import em
+import pkg_resources
 
-RESOURCE_PATH = os.path.join(os.path.dirname(__file__), '..', 'resource')
+RESOURCE_PATH = pkg_resources.resource_filename('ros2pkg', 'resource/')
```
dirk-thomas: Why the trailing slash here?

dhood: think this usage in general might be off now that I've read [this](http://setuptools.readthedocs.io/en/latest/pkg_resources.html#basic-resource-access): "Do not use os.path routines to manipulate resource paths, as they are not filesystem paths.". I was going for a minimal change but will replicate https://github.com/ament/ament_package/blob/28308125089b607f2797abee577371d7ee9a0934/ament_package/templates.py#L24 exactly

ros2pkg/package.xml
mikaelarguedas: I based my comment on the content of this repository. the `ros2cli` package (dependency of this one) uses `setuptools` and `pkg_resources` in multiple locations and doesnt declare a dependency on it in either the `package.xml` or the python files that import the module. 
Is the general rule that any package using pkg_resources should declare that dependency in their package.xml ?

dirk-thomas: Yes, that would make sense.

mikaelarguedas: I guess this should be python3-pkg-resources then

ros2pkg/ros2pkg/api/create.py
dhood: well aren't I learning a lot today..! `resource_filename` will _return_ a filesystem path, I think that comment I referenced above is referring to the arguments being passed _in_, where we aren't using `os.path` routines anyway. Still, I added 99f0127967c799edea03d018037790f038b40170. If I understood correctly, `resource_filename` will potentially extract files/directories if the package is installed zipped, and this way it will only extract the files when they're needed, as opposed to the whole resource directory being extracted (or we can swap back).

ros2pkg/ros2pkg/api/create.py
```diff
@@ -65,8 +68,7 @@ def _create_folder(folder_name, base_directory, exist_ok=True):
 
 
 def _create_template_file(template_file_name, output_directory, output_file_name, template_config):
-    template_path = os.path.abspath(
-        os.path.join(__file__, RESOURCE_PATH, template_file_name))
+    template_path = _get_template_path(template_file_name)
```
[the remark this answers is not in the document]
dhood: yeah, that would make sense f18f246fa95a6f98c692874ab8082bcbb6aa6c0f

ros2pkg/package.xml
mikaelarguedas: Keys used in `package.xml` are used to query the rosdep database. In general we try to make sure that the rosdep keys resolve to apt packages if such package exist and avoid pip rules as much as we can as we cannot embed these dependencies in our debs.

My guess is that is was an oversight when the [dependency was added to ament_package](https://github.com/ament/ament_package/pull/50) to fix failing downstream packages. As [setuptools depends on pkg_resources](https://packages.ubuntu.com/xenial/python3-setuptools), the more generic dependency does fix the problem.

dirk-thomas: I would guess the dependency was even added in to the xml before the rosdep rules existed. Any the assumption was probably at that time there is only `setuptools` - as you said `pkg-resources` is not a PyPI package.

dhood: ok: minimal, apt-favouring dependency of `python3-pkg-resources` used in https://github.com/ros2/ros2cli/pull/87/commits/7c2aef86b21be98700f93ee7ea7398de3604b1ea :+1:

dhood:
ci just for linters [Build Status](https://ci.ros2.org/job/ci_windows/4279/)

PR #90, opened 2018-05-22
add ament_package

Karsten1987:
follow up PR from https://github.com/ros2/ros2cli/pull/42#issuecomment-385086830


dhood:
@mikaelarguedas don't think those wiki edits should be removed, they're for fat archive instructions

PR #95, opened 2018-06-01
add ros2 param

dirk-thomas:
* `list` outputs the names of all parameters, either for all nodes or a node passed as an argument
* `get` outputs the value of a specific parameter, the type information can be suppressed
* `set` sets a parameter, the type is being determined by guessing the primitive types, no array types atm
* `delete` removed a specific parameter (at least it should but it seems that the service doesn't work for that use case).

### DIFF
--- a/ros2param/ros2param/verb/list.py
+++ b/ros2param/ros2param/verb/list.py
@@ -1,0 +1,98 @@
+# Copyright 2018 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import sys
+
+from rcl_interfaces.srv import ListParameters
+import rclpy
+from ros2cli.node.direct import DirectNode
+from ros2cli.node.strategy import add_arguments
+from ros2cli.node.strategy import NodeStrategy
+from ros2node.api import get_node_names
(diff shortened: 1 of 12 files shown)


ros2param/ros2param/verb/list.py
```diff
@@ -0,0 +1,98 @@
...
+                for node_name in [n for n in node_names if n not in futures]:
+                    # call as soon as ready
+                    client = clients[node_name]
+                    if client.service_is_ready():
+                        request = ListParameters.Request()
+                        print(node_name, request)
+                        future = client.call_async(request)
+                        futures[node_name] = future
+                    else:
+                        print(node_name, 'not ready')
```
sloretz: Looks like there are a few debug print statements. Use the node's logger for debug messages instead?

dirk-thomas: I removed the debug output in 55a5948af40f1cd8989599d1c0c4bd9205ceb05e.

mikaelarguedas:
Tested this along https://github.com/ros2/rclcpp/pull/489 and it works great :+1: 
(I haven't tested with many requests queued up, only one a a time)

I pushed a few nitpicks, feel free to revert them if they're not appropriate.

A few questions that come to mind:

- Could we add the option to provide the type explicitly ?
  - e.g. if I want to pass to a string parameter a value that can be converted to another type ("0", "True", "42.42"...)
- Is there something in the API preventing to sex the array types ? or was it just to keep the scope reasonable?

dirk-thomas:
> * Could we add the option to provide the type explicitly ?
> * Is there something in the API preventing to set the array types ? or was it just to keep the scope reasonable?

Additional feature can be added anytime. You could add it to this PR right now or keep it for a future PR. I just wanted to get a first version of this verb in to be able to use / test the parameters more easily.

mikaelarguedas:
> Additional feature can be added anytime. You could add it to this PR right now or keep it for a future PR. I just wanted to get a first version of this verb in to be able to use / test the parameters more easily.

Follow up is fine by me :+1: it's great to have a first version of this verb out!
I just wanted to make sure that there was not a technical blocker for implementing all parameter types in the other verbs, I didn't mean to appear requesting additional features in this PR :).

PR #96, opened 2018-06-03
Implementation for #89

### LINKED ISSUE #89: [ros2 pkg] add option to get package "share directory" directly
Currently you can only get a package's installation prefix with this command using `ros2 pkg prefix <package name>`, but it would be convenient to be able to get at least the "share" directory of the package directly, if not others as well, i.e. `<install prefix>/share/<package name>` rather than just `<install prefix>`.

The command could either be `ros2 pkg share <package name>` or something like `ros2 pkg prefix --share <package name>` instead.

Nickolaim:
The usage is below.  Closes #89.

``` bash
$ ros2 pkg prefix ament_flake8
/home/nick/ros2_ws/install

$ ros2 pkg prefix --share ament_flake8
/home/nick/ros2_ws/install/share/ament_flake8

$ ros2 pkg prefix -h
usage: ros2 pkg prefix [-h] [--share] package_name

Output the prefix path of a package

positional arguments:
  package_name  The package name

optional arguments:
  -h, --help    show this help message and exit
  --share       show share directory for the package

```

### DIFF
--- a/ros2pkg/ros2pkg/api/__init__.py
+++ b/ros2pkg/ros2pkg/api/__init__.py
@@ -15,7 +15,6 @@
 import os
 
 from ament_index_python import get_package_prefix
-from ament_index_python import get_package_share_directory
 from ament_index_python import get_packages_with_prefixes
 from ament_index_python import PackageNotFoundError
 
@@ -36,13 +35,6 @@ def get_prefix_path(package_name):
     except PackageNotFoundError:
         return None
     return prefix_path
-
-
-def get_share_directory(package_name):
-    try:
-        return get_package_share_directory(package_name)
-    except PackageNotFoundError:
-        return None
 
 
 def get_executable_paths(*, package_name):
--- a/ros2pkg/ros2pkg/api/__init__.py
+++ b/ros2pkg/ros2pkg/api/__init__.py
@@ -15,6 +15,7 @@
 import os
 
 from ament_index_python import get_package_prefix
+from ament_index_python import get_package_share_directory
 from ament_index_python import get_packages_with_prefixes
 from ament_index_python import PackageNotFoundError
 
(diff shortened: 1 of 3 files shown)


ros2pkg/ros2pkg/api/__init__.py
```diff
@@ -37,6 +38,13 @@ def get_prefix_path(package_name):
     return prefix_path
 
 
+def get_share_directory(package_name):
+    try:
+        return get_package_share_directory(package_name)
+    except PackageNotFoundError:
+        return None
```
dirk-thomas: Since this function did nothing more than calling another function I would rather not add this to the API module.

ros2pkg/ros2pkg/verb/prefix.py
```diff
@@ -24,10 +25,16 @@ def add_arguments(self, parser, cli_name):
...
+            help='show share directory for the package')
         arg.completer = package_name_completer
 
     def main(self, *, args):
         prefix_path = get_prefix_path(args.package_name)
         if prefix_path is None:
             return 'Package not found'
         print(prefix_path)
+        if args.share:
+            print('Share dir: %s' % get_share_directory(args.package_name))
```
dirk-thomas: In the current form this doesn't seem very useful beyond being read by a user.

Instead I would suggest to only show the share directory when using the new option. And also not show any text in front of the path. That will enable to use the result in cases where you want to construct a path, e.g. `cat $(ros2 pkg prefix --share PKGNAME)/package.xml`.

ros2pkg/ros2pkg/api/__init__.py
Nickolaim: Removed.

ros2pkg/ros2pkg/verb/prefix.py
Nickolaim: Please take a look at the latest version

Nickolaim:
Feedback addressed.

PR #97, opened 2018-06-11
add ros2 lifecycle

dirk-thomas:
Adding `ros2 lifecycle` with the following verbs: `get`, `list`, `set`.

Requires ros2/rosidl#281.

### DIFF
--- a/ros2lifecycle/package.xml
+++ b/ros2lifecycle/package.xml
@@ -1,0 +1,28 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="2">
+  <name>ros2lifecycle</name>
+  <version>0.4.0</version>
+  <description>
+    The lifecycle command for ROS 2 command line tools.
+  </description>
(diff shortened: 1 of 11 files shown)


mikaelarguedas:
Flake8 failures addressed in https://github.com/ros2/ros2cli/commit/606e49cff3c898b3ec2ea8bdbbf1da670255c9c3

wjwwood:
Retroactive summary for this feature:

add ros2 lifecycle
=============

### Purpose / Use Cases (Why implement this feature?)

This was implemented to provide a way to control and introspect "lifecycle" ROS nodes from the command line.

### Design (How does it work?)

It provides a new "command" to the `ros2` command line tool call `lifecycle` which in turn has several `verbs` for interacting with or introspecting ROS lifecycle nodes.

#### Inputs / Outputs / API (How to use it?)  

The usage of the tool is documented in the help text with `ros2 lifecycle -h`.

There is a "public" API which implements the actions the tool can take (used internally by the tool) but which is not currently documented, see:

https://github.com/ros2/ros2cli/blob/211fc1648c03e1f4dbf08910ec3f96be66ecca2e/ros2lifecycle/ros2lifecycle/api/__init__.py#L1-L169

PR #98, opened 2018-06-12
update ros2 param list output for a specific node

dirk-thomas:
Not showing the node name when a specific node was requested.

Follow up of #95.


dirk-thomas:
The information which node has been queried sounds useful (at least some time). An option to toggle between both behaviors would certainly be good.

PR #106, opened 2018-06-22
Update matching for lifecycle service types to support Opensplice

dhood:
@ironmig noticed that `ros2 service list` wasn't reporting Opensplice lifecycle nodes properly (e.g. `ros2 run lifecycle lifecycle_talker`).

The service name and type for opensplice that's reported by `get_service_names_and_types` is:

```
Topic: /lc_talker/get_state
Type: 'lifecycle_msgs/Sample_GetState'
```

I took a quick look at the opensplice typesupport, and it seems to be on purpose that we [use samples vs msgs names](https://github.com/ros2/rmw_opensplice/blob/e0696909726cce9240ce7e9a20de17421f60daf2/rosidl_typesupport_opensplice_cpp/resource/srv__type_support.cpp.em#L235) so I assume that that is a requirement of opensplice's typesupport. As such this PR loosens the check to match it, rather than attempting to remove Sample_ from the type. It may very well be feasible to make `get_service_names_and_types` report the type without having Sample_ in it, but if so we can do that at a later date IMO.

### DIFF
--- a/ros2lifecycle/ros2lifecycle/api/__init__.py
+++ b/ros2lifecycle/ros2lifecycle/api/__init__.py
@@ -35,7 +35,8 @@ def _has_lifecycle(node_name, service_names_and_types):
     for (service_name, service_types) in service_names_and_types:
         if (
             service_name == '/{node_name}/get_state'.format_map(locals()) and
-            'lifecycle_msgs/GetState' in service_types
+            ('lifecycle_msgs/GetState' in service_types or
+                'lifecycle_msgs/Sample_GetState' in service_types)  # Format used by OpenSplice
         ):
             return True
     return False


Karsten1987:
I can confirm that this resolves the problem.

mikaelarguedas:
Tested this and now instead of returning "node not found" it crashes:  `ros2 lifecycle get lc_talker`.
<details>

```
Failed to wait on wait set: failed to detach guard condition, at /home/mikael/work/ros2/bouncy_ws/src/ros2/rmw_opensplice/rmw_opensplice_cpp/src/rmw_wait.cpp:375, at /home/mikael/work/ros2/bouncy_ws/src/ros2/rcl/rcl/src/rcl/wait.c:647
```

</details>

@dhood Is this something you experienced when testing this change?

I think it's worth fixing the reported type to not include the `Sample` as this is messing with other tools as well. For example if I want to call the service directly with `ros2 service call /lc_talker/get_state lifecycle_msgs/Sample_GetState` it will crash because the our message structures have no idea of what `Sample_GetState` is.

So I recommend fixing it in `get_service_names_and_types` directly rather than working around it with this patch.

I'll try to submit a patch later today

dirk-thomas:
I am not sure if this is the right way to address the problem. Imo the underlying API needs to provide the same information across all rmw impl. otherwise all user level code has to handle the difference. Therefore I think this should be patched in `rmw_opensplice_cpp` instead.

dhood:
Yeah, realising that what this API returns is representing ROS msg types as opposed to actual DDS type, it's reasonable to change that conversion layer without modifying the type sent on the wire. Sounds like @mikaelarguedas is on it

mikaelarguedas:
Opened https://github.com/ros2/rmw_opensplice/pull/234 with the fix I had in mind

tfoote:
superseded by ros2/rmw_opensplice#234

PR #115, opened 2018-06-26
cancel timer before letting caller use the node to avoid spurious wakeups for consumers

mikaelarguedas:
@dirk-thomas Is this timer intended to stay alive and be used by consumers?

Otherwise I can update this PR to destroy the timer before exiting the init function

### DIFF
--- a/ros2cli/ros2cli/node/direct.py
+++ b/ros2cli/ros2cli/node/direct.py
@@ -37,11 +37,12 @@ class DirectNode:
             args, 'node_name_suffix', '_%d' % os.getpid())
         self.node = rclpy.create_node(HIDDEN_NODE_PREFIX + 'ros2cli_node' + node_name_suffix)
         timeout = getattr(args, 'spin_time', DEFAULT_TIMEOUT)
-        self.timer = self.node.create_timer(timeout, timer_callback)
+        timer = self.node.create_timer(timeout, timer_callback)
 
         while not timeout_reached:
             rclpy.spin_once(self.node)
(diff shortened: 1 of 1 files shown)


dirk-thomas:
I don't think the timer is used anywhere. It could be made "internal" to the constructor and be destroyed immediately.

mikaelarguedas:
Sounds good updated in https://github.com/ros2/ros2cli/pull/115/commits/cbe14dde029416989d717465f9b43496611b5968

mikaelarguedas:
It can be reproduced by adding a print in the timer callback.
```
         def timer_callback():
             nonlocal timeout_reached
+            print('timer_cb')
             timeout_reached = True

```
In one terminal run: `ros2 run demo_nodes_cpp talker`
In the second one run: `ros2 topic echo /chatter std_msgs/String`

### Behavior without the patch

```
$ ros2 topic echo /chatter std_msgs/String
timer_cb
timer_cb
timer_cb
timer_cb
timer_cb
timer_cb
timer_cb
timer_cb
data: 'Hello World: 1'

timer_cb
timer_cb
data: 'Hello World: 2'

timer_cb
timer_cb
data: 'Hello World: 3'

timer_cb
timer_cb
data: 'Hello World: 4'

timer_cb
timer_cb
```

### Behavior with the patch

```
$ ros2 topic echo /chatter std_msgs/String
timer_cb
data: 'Hello World: 1'

data: 'Hello World: 2'

data: 'Hello World: 3'

data: 'Hello World: 4'
```

PR #116, opened 2018-06-26
pass actual node object to subscriber function

### LINKED ISSUE #112: `ros2 topic echo` deserialization failed using connext
## Bug report

**Required Info:**

- Operating System:
  - Windows 10
- Installation type:
  - binaries
- Version or commit hash:
  - https://ci.ros2.org/view/packaging/job/packaging_windows/1046/
- DDS implementation:
  - RTI Connext
  - N/A

#### Steps to reproduce issue

Start a talker executable in one terminal
```bat
set RMW_IMPLEMENTATION=rmw_connext_cpp
ros2 run demo_nodes_cpp talker
```

Echo the topic in another terminal
```bat
set RMW_IMPLEMENTATION=rmw_connext_cpp
REM make sure the daemon uses connext
ros2 daemon stop
ros2 daemon start

ros2 topic echo /chatter
```

#### Expected behavior

Every message published would be echoed on  the console

#### Actual behavior

The first message is printed, but then an error message `deserialize from cdr buffer failed` appears. The command never prints another message despite the talker continuing to publish.

```bat
C:\dev\ros2-windows>ros2 topic echo /chatter
RTI Data Distribution Service Evaluation License issued to OSRF (OSRF01) dthomas@osrfoundation.org For non-production use only.
Expires on 5-Nov-2018 See www.rti.com for more information.
data: 'Hello World: 50'

deserialize from cdr buffer failed
```

#### Additional information
Additionally, the command `ros2 topic echo /chatter` cannot be interrupted using `CTRL` + `c`. I had to kill the process in the task manager.

mikaelarguedas:
Opening for visibility until a better workaround is found

Right now the spinning in DIrectNode causes #112 on Connext.

This PR hackily moves the logic to create and destroy the node in the echo verb itself as there is no need for a bootstrap phase for the node as it doesn't need to aggregate network information.

Fixes #112

### DIFF
--- a/ros2topic/ros2topic/verb/echo.py
+++ b/ros2topic/ros2topic/verb/echo.py
@@ -15,14 +15,13 @@
 from argparse import ArgumentTypeError
 from collections import OrderedDict
 import importlib
-import os
 import sys
 
 import rclpy
 from rclpy.expand_topic_name import expand_topic_name
 from rclpy.qos import qos_profile_sensor_data
 from rclpy.validate_full_topic_name import validate_full_topic_name
-from ros2cli.node.direct import DirectNode, HIDDEN_NODE_PREFIX
+from ros2cli.node.direct import DirectNode
 from ros2topic.api import get_topic_names_and_types
 from ros2topic.api import TopicNameCompleter
 from ros2topic.verb import VerbExtension
@@ -77,14 +76,8 @@ def main(args):
         callback = subscriber_cb(args)
     else:
         callback = subscriber_cb_csv(args)
-
-    rclpy.init()
-
-    node_name_suffix = '_%d' % os.getpid()
-    node = rclpy.create_node(HIDDEN_NODE_PREFIX + 'ros2topic_echo_node' + node_name_suffix)
-    subscriber(node, args.topic_name, args.message_type, callback)
-    node.destroy_node()
-    rclpy.shutdown()
+    with DirectNode(args) as node:
+        subscriber(node, args.topic_name, args.message_type, callback)
 
 
 def register_yaml_representer():
--- a/ros2topic/ros2topic/verb/echo.py
+++ b/ros2topic/ros2topic/verb/echo.py
@@ -77,7 +77,7 @@ def main(args):
     else:
         callback = subscriber_cb_csv(args)
     with DirectNode(args) as node:
-        subscriber(node, args.topic_name, args.message_type, callback)
+        subscriber(node.node, args.topic_name, args.message_type, callback)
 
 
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
I'm still trying to find out.
I created a pure listener that should be identical to DirectNode and that doesn't reproduce the issue: https://github.com/ros2/demos/commit/c686587df764cbab132924cfb0b14dc8dbb908a6

dirk-thomas:
Maybe it is related to the "relaying" of the attributes of the `DirectNode`? How about passing the actual node to the function?

```
    subscriber(node.node, args.topic_name, args.message_type, callback)
                   ^^^^^
```

mikaelarguedas:
I opted for this as it doesnt look like there's not much value in using DirectNode as it introduce an arbitrary 0.5sec delay, timer and spin that cannot be overriden by users and is not necessary for the good functioning of the tool.

I guess we could have a DirectNode and a DirectNodeWithTimer?
DirectNode would only take care of rclpy and node initialization/destruction:
https://github.com/ros2/ros2cli/blob/14ca069b9b3d4654379080db7dd2247f83c90a0d/ros2cli/ros2cli/node/direct.py#L34-L38

The DirectNodeWithTimer init function would add the timer and spinning logic:
```
super.__init__(args)
timeout = getattr(args, 'spin_time', DEFAULT_TIMEOUT)
self.timer = self.node.create_timer(timeout, timer_callback)

while not timeout_reached:
    rclpy.spin_once(self.node)
```

mikaelarguedas:
> Maybe it is related to the "relaying" of the attributes of the DirectNode? How about passing the actual node to the function?

Yeah I explored that for a bit at the beginning without success. But maybe I had other changes that made it not work. Let me check again

dirk-thomas:
> I guess we could have a DirectNode and a DirectNodeWithTimer?

Possible, but first it needs to be clear why the current one doesn't work as-is.

mikaelarguedas:
> Maybe it is related to the "relaying" of the attributes of the DirectNode? How about passing the actual node to the function?
```
    subscriber(node.node, args.topic_name, args.message_type, callback)
                   ^^^^^
```

Works! :tada: 

How weird I'm pretty sure I tried this and other things like inheriting from an actual rclpy node without success.

Can you also try it to confirm it doesn't work only on my machine ?

Shell1:
`RMW_IMPLEMENTATION=rmw_connext_cpp ros2 topic echo /chatter std_msgs/String`
Shell2:
`RMW_IMPLEMENTATION=rmw_connext_cpp ros2 run demo_nodes_cpp talker`

PR #119, opened 2018-06-27
fix echo for nested messages

dirk-thomas:
Without the patch only the root message got converted into an ordered dict. Submessages in collections like lists and dicts were not changed. Those were therefore not represented in yaml correctly and with the new PyYAML version those even failed to be processed since no custom representer for these types is defined.

The patch makes sure to recursively process all fields as well as items in collections.

I tried it with:
* `ros2 topic pub /test test_msgs/StaticArrayNested`
* `ros2 topic echo /test`

### DIFF
--- a/ros2topic/ros2topic/verb/echo.py
+++ b/ros2topic/ros2topic/verb/echo.py
@@ -145,7 +145,11 @@ def subscriber_cb(args):
 
 
 def msg_to_yaml(args, msg):
-    return yaml.dump(msg_to_ordereddict(args, msg), width=sys.maxsize)
+    return yaml.dump(
+        msg_to_ordereddict(
+            msg,
+            truncate_length=args.truncate_length if args.full_length else None
+        ), width=sys.maxsize)
 
 
 def subscriber_cb_csv(args):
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
I can confirm that this fixes the echo verb with the new pyyaml version.

There is one weirdness with this example:
byte_value sent is `b\x00` but echo shows
```
  byte_value: !!binary |
    AA==
```

Not a blocker to merge this but it may be worth looking into

dirk-thomas:
> byte_value sent is b\x00 but echo shows
> Not a blocker to merge this but it may be worth looking into

That is the same as before afaik. Since binary shouldn't be printed as-is to the console what would be your suggestion to improve it?

mikaelarguedas:
> That is the same behavior as before afaik. 

You're right that was already displayed like that before.

> Since binary shouldn't be printed as-is to the console what would be your suggestion to improve it?

We could display its' string representation as a builtin.str but that's a bit hacky so I'm not sure what would be better.

dirk-thomas:
> We could display its' string representation as a builtin.str ...

94c08923721f7fb3219191ae84b777c39b5bff53 might get a bit lengthy but it seems easier to read for users.

mikaelarguedas:
Looks nicer :+1:

PR #123, opened 2018-07-12
 [topic pub] add option to limit printing published msgs

dirk-thomas:
I found myself patching the code several times when I wanted to publish messages with a higher rate but without the overhead of flooding the console (which essentially slows down the loop). Hence the first commit adds an option to control how many published message should be printed (e.g. try `-r 1000 -p 250`).

The second commit only removes a default value which doesn't make sense. The default node name is not as indicated and empty string but is being set by the script later if no not-empty node name is provided.


mikaelarguedas:
as this has low risk of regression and taking it will allow us to fast-forward bouncy to the current state of master; I'm planning on integrating this to the bouncy patch release

PR #126, opened 2018-07-25
Fix echo for big array messages

yechun1:
Issue1: ros2 topic echo pointcould2(big arrays), has no response, updated the logical to make more sensible.

    a. (by default) full_length=false, truncate_length=128, then print max 128 (fix big arrays issue)

    b. pass truncate_length=X, then print max X.

    c. pass full_length=true (whatever truncate_length), then set truncate_length=None and print full_length.

Issue2: pass truncate_length to _convert_value(), this should be required.

connect to ros2/ros2cli#127

### DIFF
--- a/ros2topic/ros2topic/verb/echo.py
+++ b/ros2topic/ros2topic/verb/echo.py
@@ -148,7 +148,7 @@ def msg_to_yaml(args, msg):
     return yaml.dump(
         msg_to_ordereddict(
             msg,
-            truncate_length=args.truncate_length if args.full_length else None
+            truncate_length=args.truncate_length if not args.full_length else None
         ), width=sys.maxsize)
 
 
@@ -201,7 +201,7 @@ def msg_to_ordereddict(msg, truncate_length=None):
     # We rely on __slots__ retaining the order of the fields in the .msg file.
     for field_name in msg.__slots__:
         value = getattr(msg, field_name, None)
-        value = _convert_value(value)
+        value = _convert_value(value, truncate_length=truncate_length)
         # remove leading underscore from field name
         d[field_name[1:]] = value
     return d


ros2topic/ros2topic/verb/echo.py
```diff
@@ -201,7 +201,7 @@ def msg_to_ordereddict(msg, truncate_length=None):
     # We rely on __slots__ retaining the order of the fields in the .msg file.
     for field_name in msg.__slots__:
         value = getattr(msg, field_name, None)
-        value = _convert_value(value)
+        value = _convert_value(value, truncate_length)
```
dirk-thomas: Since `truncate_length` is a keyword argument please pass it explicitly as such: `truncate_length=truncate_length`.

yechun1: @dirk-thomas thanks for review, updated the code.

mikaelarguedas:
One surprising thing (but not directly related to this change) is that we sometimes end up with '...' and sometimes with '.....' for the truncated part (note it's 5 and not 6 dots, pretty surprising).

As this is a minor detail compared to the improvement provided by this PR, I'll merge this as is a look at or ticket the minor display hiccup

e.g.:
```
$ ros2 topic echo /image sensor_msgs/Image -l 3
header:
  stamp:
    sec: 0
    nanosec: 0
  frame_id: '29'
height: 240
width: 320
encoding: bgr...
is_bigendian: 0
step: 960
data: [0, 0, 0, '...']
```

```
$ ros2 topic echo /image sensor_msgs/Image -l 2
header:
  stamp:
    sec: 0
    nanosec: 0
  frame_id: '12'
height: 240
width: 320
encoding: bg...
is_bigendian: 0
step: 960
data: [0, 0, '.....']
```

dirk-thomas:
> As this is a minor detail compared to the improvement provided by this PR, I'll merge this as is a look at or ticket the minor display hiccup

Is the "minor display hiccup" introduced by this PR or can it be reproduced without? If it is the former this shouldn't be merged until the "hiccup" is addressed. If it is the latter this can be safely merged and the "hiccup" should be fixed (or at least ticketed if the time to fix it is significant).

mikaelarguedas:
As the truncation doesnt work without this patch, not sure how to reproduce it without applying this.

I'll have a quick look at the hiccup later today and will merge this + ticket or fix the hiccup before end of day

yechun1:
root caused the issue related with previous code which inside nested _convert_value function. 

# for test code, first time convert data to  (first if, for t in (bytes, list, str, tuple))
value [0, 0, '...']
k in value [0, 0, '...']
k: 0
k: 0
k: ...
# second time call nested convert_value by (second elfi, for t in (list, tuple))
when k=...
# for "ros2 topic echo /topic -l 2"
the len(k) = 3, truncate_length = 2.
convert the '...' as real data as '..' and append another '...'.
so the result is "....."

# correct 
when k='...'
for "ros2 topic echo /topic -l 3",  len("...") > truncate_length not ture. skip convert.

Is it possible to remove nested function for _convert_value, it's a little difficult to understand.

mikaelarguedas:
Thanks @yechun1 for looking into it. I ticketed it at https://github.com/ros2/ros2cli/issues/129 so that this can be merged and backported

mikaelarguedas:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4983/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1796/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4139/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4995/)

ros-discourse:

https://discourse.ros.org/t/image-and-pointcloud2-performance-issue-on-ros2-python/5391/1

PR #131, opened 2018-08-08
add support for parameter prefixes in ros2 param list

mikaelarguedas:


mikaelarguedas:
CI just for linters: [Build Status](https://ci.ros2.org/job/ci_linux/5015/)

ros2param/ros2param/verb/list.py
```diff
@@ -46,6 +49,9 @@ def main(self, *, args):  # noqa: D102
             if args.node_name not in node_names:
                 return 'Node not found'
             node_names = [args.node_name]
+        param_prefixes = []
+        if args.param_prefixes is not None:
+            param_prefixes = [args.param_prefixes]
```
wjwwood: Won't this always be a list of 1?

ros2param/ros2param/verb/list.py
```diff
@@ -36,6 +36,9 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
         parser.add_argument(
             '--include-hidden-nodes', action='store_true',
             help='Consider hidden nodes as well')
+        parser.add_argument(
+            '--param-prefixes', nargs='?',
```
sloretz: I'm not sure `nargs='?'` is the right choice because it allows an empty `--param-prefixes`

```
$ ros2 param list --param-prefixes
listener:
  bar.baz
  foo
  foo.bar
  foo.bar.baz
```

Is the intent to allow multiple prefixes to be specified? If so, `nargs='+'` or `action='append'` might be better choices.

mikaelarguedas:
Thanks for the feedback, I simplified the logic in https://github.com/ros2/ros2cli/pull/131/commits/c67f9d8a7846e5a78876972b737f16313a1323de and it should now support multiple prefixes

Issue #132, opened 2018-08-13
[ros2topic] enable ros2topic hz delay bw command


yechun1:
Reference to discussing ["rostopic command for ROS2"](https://discourse.ros.org/t/rostopic-command-for-ros2/2565/4)

ROS topic commands are very useful for simple performance tests such as hz/delay/bw, but those tools are not supported in ros2 topic right now, if there is no body work on there. I would like to contribute to port ROS topic(hz/delay/bw) to ROS2.

Todo list:
- [x] [ ros2 topic hz] (https://github.com/ros2/ros2cli/pull/133)
- [x] [ ros2 topic delay] (#139)
- [x] [ ros2 topic bw] (https://github.com/ros2/ros2cli/pull/190)

dirk-thomas:
Great! Let use know if you have any questions.

yechun1:
While porting bw command, did not find the api to get size of the subscribed msg. Could anyone help? 

ROS use `AnyMsg` and `len(data._buff)` to get the size (line to https://github.com/ros/ros_comm/blob/melodic-devel/tools/rostopic/src/rostopic/__init__.py#L440). But how to replace on ROS2?

josephduchesne:
I'm not sure if there is a canonical size of a ros message any more, especially exposed in python. 

Rclpy now hands serialization off to the cpp rcl then rmw layer from what I can tell.

https://github.com/ros2/demos/pull/185 and the other PRs linked therin add rmw+rcl+rclcpp support for returning the raw serialized message in a raw type callback, but I don't think this is exposed in rclpy (yet?).

I wonder which would be easier or better:
- Port bw to C++ and using the existing raw function (Can ros2cli verbs be written in C++?)
- Write a purpose built C++ glue function to get just this data and invoke it from the python side
- Add raw=False keyword arg to rclpy Node.create_subscription and all the way down to support returning a byte list subscription in python.

I may take a crack at the last option there and if it turns out to be reasonable, I'll submit a PR and link it here, otherwise I'll try to explain what trouble I ran into here.

josephduchesne:
I've created PR ros2/rclpy#242 that needs review for design, and some touching up to work towards providing a rospy implementation of raw message subscriptions like roscpp has thanks to ros2/demos#185

LucidOne:
[...]

PR #133, opened 2018-08-14
Enable ros2tpoic hz command

yechun1:
For performance test to print the average publishing rate to screen. Ported rostopic hz command from ros to ros2, based on the file as below

https://github.com/ros/ros_comm/blob/melodic-devel/tools/rostopic/src/rostopic/__init__.py

Connects to https://github.com/ros2/ros2cli/issues/132

### DIFF
--- a/ros2topic/ros2topic/verb/hz.py
+++ b/ros2topic/ros2topic/verb/hz.py
@@ -30,8 +30,9 @@
 # LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
 # ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
-#
-# Revision $Id$
+
+# This file is originally from:
+# https://github.com/ros/ros_comm/blob/6e5016f4b2266d8a60c9a1e163c4928b8fc7115e/tools/rostopic/src/rostopic/__init__.py
 
 # make sure we aren't using floor division
 from __future__ import division, print_function
--- a/ros2topic/ros2topic/verb/hz.py
+++ b/ros2topic/ros2topic/verb/hz.py
@@ -1,0 +1,2156 @@
+#!/usr/bin/env python
+# Software License Agreement (BSD License)
+#
+# Copyright (c) 2008, Willow Garage, Inc.
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#
+#  * Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+#  * Redistributions in binary form must reproduce the above
+#    copyright notice, this list of conditions and the following
+#    disclaimer in the documentation and/or other materials provided
+#    with the distribution.
+#  * Neither the name of Willow Garage, Inc. nor the names of its
+#    contributors may be used to endorse or promote products derived
+#    from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+# FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+# COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+# INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+# BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+#
+# Revision $Id$
+
+# make sure we aren't using floor division
+from __future__ import division, print_function
+
+NAME='rostopic'
+
+import argparse
+import os
+import sys
+import math
+import socket
+import time
+import traceback
+import yaml
+try:
+    from xmlrpc.client import Fault
+except ImportError:
+    from xmlrpclib import Fault
+
+from operator import itemgetter
+try:
+    from urllib.parse import urlparse
+except ImportError:
+    from urlparse import urlparse
+
+import genpy
+
+import roslib.message
+import rosgraph
+#TODO: lazy-import rospy or move rospy-dependent routines to separate location
+import rospy
+
+try:
+    long
+except NameError:
+    long = int
+
+
+class ROSTopicException(Exception):
+    """
+    Base exception class of rostopic-related errors
+    """
+    pass
+class ROSTopicIOException(ROSTopicException):
+    """
+    rostopic errors related to network I/O failures
+    """
+    pass
+
+def _check_master():
+    """
+    Make sure that master is available
+    :raises: :exc:`ROSTopicException` If unable to successfully communicate with master
+    """
+    try:
+        rosgraph.Master('/rostopic').getPid()
+    except socket.error:
+        raise ROSTopicIOException("Unable to communicate with master!")
+    
+def _master_get_topic_types(master):
+    try:
+        val = master.getTopicTypes()
+    except Fault:
+        #TODO: remove, this is for 1.1
+        sys.stderr.write("WARNING: rostopic is being used against an older version of ROS/roscore\n")
+        val = master.getPublishedTopics('/')
+    return val
+
+class ROSTopicHz(object):
+    """
+    ROSTopicHz receives messages for a topic and computes frequency stats
+    """
+    def __init__(self, window_size, filter_expr=None, use_wtime=False):
+        import threading
+        from collections import defaultdict
+        self.lock = threading.Lock()
+        self.last_printed_tn = 0
+        self.msg_t0 = -1
+        self.msg_tn = 0
+        self.times = []
+        self._last_printed_tn = defaultdict(int)
+        self._msg_t0 = defaultdict(lambda: -1)
+        self._msg_tn = defaultdict(int)
+        self._times = defaultdict(list)
+        self.filter_expr = filter_expr
+        self.use_wtime = use_wtime
+        
+        # can't have infinite window size due to memory restrictions
+        if window_size < 0:
+            window_size = 50000
+        self.window_size = window_size
+
+    def get_last_printed_tn(self, topic=None):
+        if topic is None:
+            return self.last_printed_tn
+        return self._last_printed_tn[topic]
+
+    def set_last_printed_tn(self, value, topic=None):
+        if topic is None:
+            self.last_printed_tn = value
+        self._last_printed_tn[topic] = value
+
+    def get_msg_t0(self, topic=None):
+        if topic is None:
+            return self.msg_t0
+        return self._msg_t0[topic]
+
+    def set_msg_t0(self, value, topic=None):
+        if topic is None:
+            self.msg_t0 = value
+        self._msg_t0[topic] = value
+
+    def get_msg_tn(self, topic=None):
+        if topic is None:
+            return self.msg_tn
+        return self._msg_tn[topic]
+
+    def set_msg_tn(self, value, topic=None):
+        if topic is None:
+            self.msg_tn = value
+        self._msg_tn[topic] = value
+
+    def get_times(self, topic=None):
+        if topic is None:
+            return self.times
+        return self._times[topic]
+
+    def set_times(self, value, topic=None):
+        if topic is None:
+            self.times = value
+        self._times[topic] = value
+
+    def callback_hz(self, m, topic=None):
+        """
+        ros sub callback
+        :param m: Message instance
+        :param topic: Topic name
+        """
+        # #694: ignore messages that don't match filter
+        if self.filter_expr is not None and not self.filter_expr(m):
+            return
+        with self.lock:
+            curr_rostime = rospy.get_rostime() if not self.use_wtime else \
+                    rospy.Time.from_sec(time.time())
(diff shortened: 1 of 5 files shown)


ros2topic/ros2topic/verb/hz.py
```diff
@@ -0,0 +1,239 @@
+# Copyright 2017 Open Source Robotics Foundation, Inc.
```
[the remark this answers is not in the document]
yechun1: Stayed with original license, but seems it does not match copyright test. 
    @pytest.mark.copyright
    def test_copyright():
        rc = main(argv=['.', 'test'])
>       assert rc == 0, 'Found errors'
E       AssertionError: Found errors
E       assert 1 == 0
test/test_copyright.py:23: AssertionError
Captured stderr call 
ros2topic/verb/hz.py: copyright=Willow Garage, Inc. (2008), license=<unknown>
1 errors

ros2topic/ros2topic/verb/hz.py
```diff
@@ -0,0 +1,239 @@
...
+
+    def callback_hz(self, m, topic=None):
+        """
+        Calculate interval time.
+
+        :param m: Message instance
+        :param topic: Topic name
+        """
+        with self.lock:
+            curr = time.time()
```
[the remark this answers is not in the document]
yechun1: There is no simple API similar with rospy.get_rostime (for example rclpy.get_rostime). Added TODO to replace time later. Would you share me the ros2 rostime reference code?

yechun1:
@mikaelarguedas thanks for your review and comments. Just fixed all those errors and added TODOs. One issue is running test_copyright.py will fail with original copyright. Will it make impact? or do you have suggest how to fix? thanks.

mikaelarguedas:
thanks @yechun1 for iterating.
Could you please submit new commits in the future rather than force-pushing ? this makes reviewing the new changes easier.

Regarding the copyright, it looks like the formatting was diverging from the BSD license template. You can see this commit https://github.com/ros2/ros2cli/commit/8db6c67afac741e266ef288f40513188cf18fbef for a fir that will make the copyright test pass.

Regarding ROS Time, this is pretty new in rclpy but hopefully you will find some useful information in https://github.com/ros2/rclpy/pull/210

yechun1:
@mikaelarguedas thanks for your suggestion. I have submitted new commits to fix TODOs. Would you please help to review? thanks.
One item to support multiple topics is not ported, I do not find the way to enable /topic_0 [/topic_1 [topic_2 [..]] to ROS2 topic arguments.

ros2topic/ros2topic/verb/hz.py
```diff
@@ -0,0 +1,315 @@
+# Software License Agreement (BSD License)
```
mikaelarguedas: Can you please move line 1 and 2 to line 3 and 4 (as recommended in previous review) this should be enough to fix the linter error (now that https://github.com/ament/ament_lint/pull/106 has been merged)

yechun1:
A small changes to fix code review comments. Code has been rebased and all tests passed in my local (by ament test).  @ros2/team will you please help review and rerun Ci.

> I faced a few inconsistencies when testing this patch for high frequency messages (the reported frequency was always lower but I didn't get a chance to investigate where the bottleneck was). Did you experience the same thing?
Example of commands I used for testing:
PYTHONOPTIMIZED=0 ros2 topic pub /image sensor_msgs/Image -r 1000 -p 1000
PYTHONOPTIMIZED=0 ros2 topic hz /image.

@mikaelarguedas 
I think the data reported by hz tool is real. This might be ROS2 performance issue related with pub-sub transport. I have reported similar issue (here)[https://discourse.ros.org/t/image-and-pointcloud2-performance-issue-on-ros2-python/5391].  I have write topic hz tool both by python and c++ in my local, found python test result is worse than c++ (even with PYTHONOPTIMIZE), and ros2 hz data is worse than ros, especially for big data transport. Suggest to file another issue about frequency decreases.

dirk-thomas:
> I reproduced the frequency continually reduce issue, I am investigating it.

@yechun1 What is the status on this?

yechun1:
@dirk-thomas @mikaelarguedas all found issues fixed in this PR. Function verified on my local. Would you please help to review and tell me if there is anything need to change? Thanks.

dirk-thomas:
Please use the same commit strategy as in this PR: https://github.com/ros2/ros2cli/pull/139#issuecomment-419556919

yechun1:
no problem, I'll handle this after #139(delay command) been merged. To save rework time.

yechun1:
squashed the comments and resubmitted again, please help review again.

ros2topic/ros2topic/verb/hz.py
```diff
@@ -0,0 +1,337 @@
...
+        arg = parser.add_argument(
+            'topic_name',
+            help="Name of the ROS topic to listen to (e.g. '/chatter')")
+        arg.completer = TopicNameCompleter(
+            include_hidden_topics_key='include_hidden_topics')
+        parser.add_argument(
+            '--window', '-w',
+            dest='window_size', type=unsigned_int, default=DEFAULT_WINDOW_SIZE,
+            help='window size, in # of messages, for calculating rate, '
+                 'string to (default: %d)' % DEFAULT_WINDOW_SIZE, metavar='WINDOW')
```
clalancette: Nit: I'm not sure what "string to" means here.  Can you reword or remove this piece of the help?

ros2topic/ros2topic/verb/hz.py
```diff
@@ -0,0 +1,337 @@
...
+    def main(self, *, args):
+        return main(args)
+
+
+def main(args):
+    topic = args.topic_name
+    if args.filter_expr:
+        def expr_eval(expr):
+            def eval_fn(m):
+                return eval(expr)
```
clalancette: Hm.  I know this is directly ported from the original ROS1 code, but having a naked `eval` lying around here seems pretty dangerous.  In particular, the string that is passed in here can do *any* arbitrary Python thing, including copying all of your hard drive out to the network, taking command-and-control inputs from the network, or deleting your filesystem.  I don't think that sort of power is actually required here; I think a simple regular expression could probably get most of the uses.  Could you maybe explain what you think the intended use case here is?

ros2topic/ros2topic/verb/hz.py
```diff
@@ -0,0 +1,337 @@
...
+        _rostopic_hz(node.node, topic, window_size=args.window_size, filter_expr=filter_expr,
+                     use_wtime=args.use_wtime)
+
+
+class ROSTopicHz(object):
+    """ROSTopicHz receives messages for a topic and computes frequency."""
+
+    def __init__(self, node, window_size, filter_expr=None, use_wtime=False):
+        import threading
+        from collections import defaultdict
```
clalancette: Nit; looking elsewhere in ros2cli, it looks like we generally hoist `defaultdict` to the top of the file: https://github.com/ros2/ros2cli/blob/master/ros2cli/ros2cli/entry_points.py#L16 .  Mostly a matter of style/taste, just trying to be consistent here.

ros2topic/ros2topic/verb/hz.py
```diff
@@ -0,0 +1,337 @@
...
+
+            # time reset
+            if curr_rostime.nanoseconds == 0:
+                if len(self.get_times(topic=topic)) > 0:
+                    print('time has reset, resetting counters')
+                    self.set_times([], topic=topic)
+                return
+
+            curr = curr_rostime.nanoseconds
+            if self.get_msg_t0(topic=topic) < 0 or self.get_msg_t0(topic=topic) > curr:
```
clalancette: This would be slightly more efficient as:

```
msg_t0 = self.get_msg_t0(topic=topic)
if msg_t0 < 0 or msg_t0 > curr:
   ...
```

ros2topic/ros2topic/verb/hz.py
```diff
@@ -0,0 +1,337 @@
...
+        if not self.get_times(topic=topic):
+            return
+        elif self.get_last_printed_tn(topic=topic) == 0:
+            self.set_last_printed_tn(self.get_msg_tn(topic=topic), topic=topic)
+            return
+        elif self.get_msg_tn(topic=topic) < self.get_last_printed_tn(topic=topic) + 1e9:
+            return
+        with self.lock:
+            # Get frequency every one minute
+            n = len(self.get_times(topic=topic))
```
clalancette: Again, probably more efficient to call `self.get_times(topic=topic)` right after taking the lock, then reuse the local variable.

ros2topic/ros2topic/verb/hz.py
yechun1: The filter_expr code is directly ported from the original ROS1. No changes from the first commit  ("rostopic into ros_comm" https://github.com/ros/ros_comm/commit/a82674ad87ab17e373e89d392bcf358c511026ae). One comment from code is "#694: ignore messages that don't match filter", line to https://github.com/ros/ros_comm/blob/a82674ad87ab17e373e89d392bcf358c511026ae/tools/rostopic/src/rostopic.py#L116 ", I don't know what #694 means, I guess it is internal bug track of rostopic before open source, and may noted the intended use case there. I have no idea about the requirement of the option `--filter`. Check with `echo` command of ros2topic also do not remain this option. Could we also removed here on `hz` command for code safety?

ros2topic/ros2topic/verb/hz.py
yechun1: removed. fixed in 
0ba39d3

ros2topic/ros2topic/verb/hz.py
yechun1: fixed in
0ba39d3

ros2topic/ros2topic/verb/hz.py
yechun1: thanks, fix in 0ba39d3

ros2topic/ros2topic/verb/hz.py
yechun1: fixed in 0ba39d3

ros2topic/ros2topic/verb/hz.py
clalancette: Yeah, I did some similar code archaeology, and like you said, this option existed before history starts.  I do not know the intended use case either.  @tfoote, @dirk-thomas , do you have any further history on what the `--filter` options was meant to achieve in ROS1, and whether we still need it?  The options on the table here seem to be:

1.  Remove `--filter` entirely until a use-case comes up for it.
2.  Restrict `--filter` to be a regex only.
3.  Leave `--filter` as the current `eval`, which seems pretty dangerous.

I'd vote for 1 (we can always add it back later), but I'm interested to hear additional thoughts/opinions.

mjcarroll:
I seem to recall using filter as a convenient way to create a frequency measurement around some value in the message, like if you wanted to measure the frequency of detections, or state changes, or the like.

It is also present on `rostopic echo` on ROS1.

clalancette:
> I seem to recall using filter as a convenient way to create a frequency measurement around some value in the message, like if you wanted to measure the frequency of detections, or state changes, or the like.
> 
> It is also present on `rostopic echo` on ROS1.

Yeah, from the implementation, that would have been my guess as well.  If we still think that is useful, that's fine, but I personally think that giving it `eval` is too powerful.  Since it sounds like it might be useful, then I'd think we could get most of its usefulness just by allowing a regex here.

mjcarroll:
Agree that `eval` is a bit generous in this case, since we haven't ported the `echo` variant of `--filter` either, I would be inclined to go with option 1, and then open a feature for reworking that command.

dirk-thomas:
`ast.literal_eval` could be considered for a replacement.

yechun1:
`ast.literal_eval ` seems not be satisfied  for matching a specified Python expression as ROS 1 provided. 
For examples:

```
$ rostopic echo --filter "m.data=='foo'"  /topic_name
$ rostopic echo --filter "m.transforms[0].child_frame_id == 'my_frame'" /tf
$ rostopic hz --filter "int(m.header.stamp.sec) % 3 == 0" /topic_name
```


About `ast.literal_eval ` APIs: https://docs.python.org/3/library/ast.html#ast.literal_eval
> ast.literal_eval(node_or_string)
>Safely evaluate an expression node or a string containing a Python literal or container display. The string or node provided may only consist of the following Python literal structures: strings, bytes, numbers, tuples, lists, dicts, sets, booleans, and None.
> 
> This can be used for safely evaluating strings containing Python values from untrusted sources without the need to parse the values oneself. It is not capable of evaluating arbitrarily complex expressions, for example involving operators or indexing.

dirk-thomas:
> `ast.literal_eval` seems not be satisfied for matching a specified Python expression as ROS 1 provided.

Good point. We should continue with option 3 then.

There is no "problem" from a security point of view. The user passes the filter string explicitly as a command line argument. Obviously a "malicious" parameter could be used. But the user could also just call `python -c "<malicious command>` directly...

yechun1:
I would like to keep with option 3, as rostopic echo/hz with --filter be used for a long time. Of course, if we have better solution, we could modify the code with new PR.
@clalancette other small issues from your comments fixed in 0ba39d3, please help review again if there is anything need to update. Thanks.

clalancette:
I'm going to run CI on this (mostly for the linters); assuming that goes well, I'll merge this later today.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5340/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2075/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4438/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5308/)

ros2topic/ros2topic/verb/hz.py
```diff
@@ -0,0 +1,339 @@
...
+
+    # TODO(dirk-thomas) this logic should come from a rosidl related package
+    try:
+        package_name, message_name = message_type.split('/', 2)
+        if not package_name or not message_name:
+            raise ValueError()
+    except ValueError:
+        raise RuntimeError('The passed message type is invalid')
+    module = importlib.import_module(package_name + '.msg')
+    return getattr(module, message_name)
```
dirk-thomas: The two methods `get_msg_class` and `_get_msg_class` are a 1-to-1 copy from the `delay` module. Please don't cop-n-paste the code but move it into the `api` module and reuse it in both locations.

yechun1:
Moved get_msg_class and get_msg_class_ from delay and hz modues to api module. Fixed in bbc2293.

clalancette:
All right, one more CI to check the new stuff, then I'll merge.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5365/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2092/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4455/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5325/)

clalancette:
All the warnings are from other, known problems (they should be fixed now).  Given that, I'm going to merge up.  Thanks for the contribution.

yechun1:
@mikaelarguedas @dirk-thomas @clalancette @mjcarroll thank you so much for taking effort on patch review and giving good comments. Glad to see this PR has been merged.

dirk-thomas:
Thank you for making this feature available!

Issue #129, opened 2018-08-01
[ros2topic] nested calls to convert_value results in inconsistent printing format


mikaelarguedas:
## Bug report

originally reported at https://github.com/ros2/ros2cli/pull/126#issuecomment-409378728, and investigated by @yechun1 at https://github.com/ros2/ros2cli/pull/126#issuecomment-409434386.

#### Steps to reproduce issue

Shell 1:
```
ros2 run image_tools cam2image -b
```
Shell 2:
```
$ ros2 topic echo /image sensor_msgs/Image -l 2
```

#### Expected behavior

Print the following to console
```
header:
  stamp:
    sec: 0
    nanosec: 0
  frame_id: '12'
height: 240
width: 320
encoding: bg...
is_bigendian: 0
step: 960
data: [0, 0, '...']
```
#### Actual behavior

The last element is printed with 5 dots
```
header:
  stamp:
    sec: 0
    nanosec: 0
  frame_id: '12'
height: 240
width: 320
encoding: bg...
is_bigendian: 0
step: 960
data: [0, 0, '.....']
```

#### Additional information
From https://github.com/ros2/ros2cli/pull/126#issuecomment-409434386:

root caused the issue related with previous code which inside nested _convert_value function. 

 Execute:  "ros2 topic echo /topic -l 2"

with test code, first time convert data to  (first if, for t in (bytes, list, str, tuple))
```
value [0, 0, '...']
k in value [0, 0, '...']
k: 0
k: 0
k: ...
```
second time call nested convert_value by (second elfi, for t in (list, tuple))
when k=...
len(k) = 3, truncate_length = 2.
convert the '...' as real data as '..' and append another '...'
so the result is "....."

while "ros2 topic echo /topic -l 3" is correct 
when k='...'
 if (len("...") > truncate_length) is not ture. skip convert.

Is it possible to remove nested function for _convert_value, it's a little difficult to understand.

PR #135, opened 2018-08-16
Fix echo sometimes printing .....

sloretz:
Fixes ros2/ros2cli#129

If `truncate_length` was less than 3 then the `...` added to a list or sequence would be truncated. This fixes the issue by refactoring `_convert_value` to only have one block of conditionals.

I also changed `bytes` from being displayed as `"b'somebytestring'"` to a string where every byte has been converted with `chr`.

To test this PR manually

```
ros2 run image_tools cam2image -b
```

and
```
ros2 topic echo /image sensor_msgs/Image -l 2
```

CI
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5054/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1850/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4196/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5048/)

connects to ros2/ros2cli#129

### DIFF
--- a/ros2topic/ros2topic/verb/echo.py
+++ b/ros2topic/ros2topic/verb/echo.py
@@ -208,40 +208,31 @@ def msg_to_ordereddict(msg, truncate_length=None):
 
 
 def _convert_value(value, truncate_length=None):
-    if any(isinstance(value, t) for t in (bytes, list, str, tuple)):
-        # truncate value if requested
+    if isinstance(value, bytes):
         if truncate_length is not None and len(value) > truncate_length:
-            value = value[:truncate_length]
-            if any(isinstance(value, t) for t in [list, tuple]):
-                value.append('...')
-            elif isinstance(value, bytes):
-                value += b'...'
-            elif isinstance(value, str):
-                value += '...'
-            else:
-                assert False
-
-    if any(isinstance(value, t) for t in (dict, OrderedDict)):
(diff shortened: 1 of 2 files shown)


ros2topic/ros2topic/verb/echo.py
```diff
@@ -208,40 +208,31 @@ def msg_to_ordereddict(msg, truncate_length=None):
...
+                value = type(value)([_convert_value(v, truncate_length) for v in value] + ['...'])
+        else:
+            # Truncate every item in the list
+            value = type(value)([_convert_value(v, truncate_length) for v in value])
+    elif isinstance(value, dict) or isinstance(value, OrderedDict):
         # convert each key and value in the mapping
         new_value = {} if isinstance(value, dict) else OrderedDict()
         for k, v in value.items():
-            new_value[_convert_value(k)] = _convert_value(
+            new_value[_convert_value(k, truncate_length=truncate_length)] = _convert_value(
```
dirk-thomas: I do see a major problem with this change: the keys in dictionary are now also being truncated which was not the case before. This will potentially lead to different keys being reduce to the same shortened key and therefore "loose" data.

mikaelarguedas: good point !

PR #139, opened 2018-08-24
Enable ros2topic delay command

yechun1:
Porting rostopic delay command from ROS to ROS2. Display delay of topic from timestamp in header.

Usage: ros2 topic delay /topic_name

Connects to: #132

### DIFF
--- a/ros2topic/ros2topic/verb/delay.py
+++ b/ros2topic/ros2topic/verb/delay.py
@@ -1,0 +1,2145 @@
+#!/usr/bin/env python
+# Software License Agreement (BSD License)
+#
+# Copyright (c) 2008, Willow Garage, Inc.
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#
+#  * Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+#  * Redistributions in binary form must reproduce the above
+#    copyright notice, this list of conditions and the following
+#    disclaimer in the documentation and/or other materials provided
+#    with the distribution.
+#  * Neither the name of Willow Garage, Inc. nor the names of its
+#    contributors may be used to endorse or promote products derived
+#    from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+# FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+# COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+# INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+# BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+#
+# Revision $Id$
+
+# make sure we aren't using floor division
+from __future__ import division, print_function
+
+NAME='rostopic'
+
+import argparse
+import os
+import sys
+import math
+import socket
+import time
+import traceback
+import yaml
+try:
+    from xmlrpc.client import Fault
+except ImportError:
+    from xmlrpclib import Fault
+
+from operator import itemgetter
+try:
+    from urllib.parse import urlparse
+except ImportError:
+    from urlparse import urlparse
+
+import genpy
+
+import roslib.message
+import rosgraph
+#TODO: lazy-import rospy or move rospy-dependent routines to separate location
+import rospy
+
+try:
+    long
+except NameError:
+    long = int
+
+
+class ROSTopicException(Exception):
+    """
+    Base exception class of rostopic-related errors
+    """
+    pass
+class ROSTopicIOException(ROSTopicException):
+    """
+    rostopic errors related to network I/O failures
+    """
+    pass
+
+def _check_master():
+    """
+    Make sure that master is available
+    :raises: :exc:`ROSTopicException` If unable to successfully communicate with master
+    """
+    try:
+        rosgraph.Master('/rostopic').getPid()
+    except socket.error:
+        raise ROSTopicIOException("Unable to communicate with master!")
+    
+def _master_get_topic_types(master):
+    try:
+        val = master.getTopicTypes()
+    except Fault:
+        #TODO: remove, this is for 1.1
+        sys.stderr.write("WARNING: rostopic is being used against an older version of ROS/roscore\n")
+        val = master.getPublishedTopics('/')
+    return val
+
+class ROSTopicHz(object):
+    """
+    ROSTopicHz receives messages for a topic and computes frequency stats
+    """
+    def __init__(self, window_size, filter_expr=None, use_wtime=False):
+        import threading
+        from collections import defaultdict
+        self.lock = threading.Lock()
+        self.last_printed_tn = 0
+        self.msg_t0 = -1
+        self.msg_tn = 0
+        self.times = []
+        self._last_printed_tn = defaultdict(int)
+        self._msg_t0 = defaultdict(lambda: -1)
+        self._msg_tn = defaultdict(int)
+        self._times = defaultdict(list)
+        self.filter_expr = filter_expr
+        self.use_wtime = use_wtime
+        
+        # can't have infinite window size due to memory restrictions
+        if window_size < 0:
+            window_size = 50000
+        self.window_size = window_size
+
+    def get_last_printed_tn(self, topic=None):
+        if topic is None:
+            return self.last_printed_tn
+        return self._last_printed_tn[topic]
+
+    def set_last_printed_tn(self, value, topic=None):
+        if topic is None:
+            self.last_printed_tn = value
+        self._last_printed_tn[topic] = value
+
+    def get_msg_t0(self, topic=None):
+        if topic is None:
+            return self.msg_t0
+        return self._msg_t0[topic]
+
+    def set_msg_t0(self, value, topic=None):
+        if topic is None:
+            self.msg_t0 = value
+        self._msg_t0[topic] = value
+
+    def get_msg_tn(self, topic=None):
+        if topic is None:
+            return self.msg_tn
+        return self._msg_tn[topic]
+
+    def set_msg_tn(self, value, topic=None):
+        if topic is None:
+            self.msg_tn = value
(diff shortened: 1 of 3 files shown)


dirk-thomas:
The code in this repository require a C++14 compiler. Some compilers depending on their version default to newer C++ standards. In your case that doesn't seem to be the case and you should manually force your compiler to use C++14 (e.g. by setting the CMake variable `-DCMAKE_CXX_STANDARD=14` or passing the appropriate compiler flag).

dirk-thomas:
Sorry, wrong issue - ignore my previous comment.

dirk-thomas:
Thank you for working on this feature.

In the current form the patch is pretty difficult to review. Please split the patch into separate commit with roughly the following structure:

* first commit containing the exact copy of the upstream file
* additional commits for any changes you applied, e.g.
  * style changes, comments, removing obsolete change in one commit
  * ROS 2 API changes in a second commit
  * functional changes in a third commit

yechun1:
Thanks for guide, I'll resubmit commits early this week.

ros2topic/ros2topic/verb/delay.py
```diff
@@ -186,218 +152,90 @@ def get_delay(self):
...
+        rt.print_delay()
 
-    if matches:
-        t, t_type = matches[0]
-        if t_type == rosgraph.names.ANYTYPE:
-            return None, None, None
-        return t_type, t, msgevalgen(topic[len(t):])
-    else:
-        return None, None, None
+    timer = node.create_timer(1, timer_callback)
```
dirk-thomas: You could register `rt.print_delay` here directly getting rid of the local function.

ros2topic/ros2topic/verb/delay.py
```diff
@@ -124,7 +123,7 @@ def callback_delay(self, msg):
                                    (msg.header.stamp.sec + msg.header.stamp.nanosec * 1e-9))
                 self.msg_tn = curr
 
-            if len(self.delays) > self.window_size - 1:
+            if len(self.delays) > self.window_size:
```
dirk-thomas: Please describe why this was changed.

ros2topic/ros2topic/verb/delay.py
```diff
@@ -149,7 +148,7 @@ def get_delay(self):
             min_delta = min(self.delays)
 
             self.last_msg_tn = self.msg_tn
-        return mean, min_delta, max_delta, std_dev, n + 1
+        return mean, min_delta, max_delta, std_dev, n
```
dirk-thomas: Please describe why this was changed.

ros2topic/ros2topic/verb/delay.py
```diff
@@ -58,6 +58,35 @@
...
+
+class DelayVerb(VerbExtension):
+    """Display delay of topic from timestamp in header."""
+
+    def add_arguments(self, parser, cli_name):
+        arg = parser.add_argument(
+            'topic',
+            help='Topic name to be calcurated the delay')
+        arg.completer = TopicNameCompleter(
+            include_hidden_topics_key='include_hidden_topics')
```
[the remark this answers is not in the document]
yechun1: Could work, same usage ref to pub.py and list.py
from ros2topic.api import TopicNameCompleter
https://github.com/ros2/ros2cli/blob/ce9077fd22cc7ccfa082d199fbdeb8553ffa9d8b/ros2topic/ros2topic/api/__init__.py#L45-L46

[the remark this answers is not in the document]
dirk-thomas: Never mind, the argument is added one level above in https://github.com/ros2/ros2cli/blob/ce9077fd22cc7ccfa082d199fbdeb8553ffa9d8b/ros2topic/ros2topic/command/topic.py#L25-L27 and will work for all subverbs under `topic`.

ros2topic/ros2topic/verb/delay.py
yechun1: To fix the issue that no data print when set window as 1, 
```ros2 topic delay /image -w 1```

Comment (yechun1) [reply ros2topic/ros2topic/verb/delay.py]: Don't know why here need to window_size - 1 ?

File: ros2topic/ros2topic/verb/delay.py [hunk 5]
```diff
@@ -58,6 +58,35 @@
...
+        parser.add_argument(
+            '--window', '-w', default=-1,
```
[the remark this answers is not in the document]
yechun1: fixed (5e7cc6a)

ros2topic/ros2topic/verb/delay.py
```diff
@@ -135,38 +90,51 @@ def __init__(self, window_size):
         self.delays = []
 
         # can't have infinite window size due to memory restrictions
-        if window_size < 0:
+        if window_size < 0 or window_size > 50000:
```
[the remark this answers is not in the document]
yechun1: fixed a5aec05

ros2topic/ros2topic/verb/delay.py
```diff
@@ -135,38 +90,51 @@ def __init__(self, window_size):
...
-            if curr_rostime.is_zero():
+            if curr_rostime.nanoseconds == 0:
                 if len(self.delays) > 0:
-                    print("time has reset, resetting counters")
+                    print('time has reset, resetting counters')
                     self.delays = []
                 return
 
-            curr = curr_rostime.to_sec()
+            curr = curr_rostime.nanoseconds * 1e-9
```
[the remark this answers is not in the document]
yechun1: fixed 2c01515 
As ROS there is to_sec and to_time method, but I found similar in ROS2, especially about msg_header.stamp. this is builtin_interface/Time type, does this function like stamp.to_time()?

ros2topic/ros2topic/verb/delay.py
yechun1: fixed 9c6ad41

yechun1:
thanks for review, I have added changes for fixing.

ros2topic/ros2topic/verb/delay.py
```diff
@@ -0,0 +1,236 @@
...
+        import threading
+        self.lock = threading.Lock()
+        self.last_msg_tn = 0
+        self.msg_t0 = -1.
+        self.msg_tn = 0
+        self.delays = []
+
+        # can't have infinite window size due to memory restrictions
+        if window_size < 0:
+            window_size = 50000
```
dirk-thomas: This "magic" upper bound is nowhere visible to the user. Please use it as a default value of the argument instead (and ensure in code that the user didn't pass a negative value).

ros2topic/ros2topic/verb/delay.py
```diff
@@ -0,0 +1,236 @@
...
+        ret = self.get_delay()
+        if ret is None:
+            print('no new messages')
+            return
+        delay, min_delta, max_delta, std_dev, window = ret
+        print('average delay: %.3f\n\tmin: %.3fs max: %.3fs std dev: %.5fs window: %s'
+              % (delay, min_delta, max_delta, std_dev, window))
+
+
+def _rostopic_delay(node, topic, window_size=-1, blocking=False):
```
dirk-thomas: I would suggest making `window_size` a required argument - otherwise this function needs to pick an opaque default value.

How does `blocking=False` work if the `msg_class` can't be determined yet?

ros2topic/ros2topic/verb/delay.py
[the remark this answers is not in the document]
dirk-thomas: Not at the moment: see ros2/rclpy#236

ros2topic/ros2topic/verb/delay.py
```diff
@@ -0,0 +1,236 @@
...
+            if self.msg_t0 < 0 or self.msg_t0 > curr:
+                self.msg_t0 = curr
+                self.msg_tn = curr
+                self.delays = []
+            else:
+                curr_sec = curr_rostime.to_msg().sec
+                curr_nanosec = curr_rostime.to_msg().nanosec
+
+                delay = ((curr_sec + curr_nanosec * 1e-9) -
+                         (msg.header.stamp.sec + msg.header.stamp.nanosec * 1e-9))
```
dirk-thomas: You should be able to compute the delay (as a Duration) by doing `curr_rostime.to_msg() - rclpy.Time.from_msg(msg.header.stamp)` (see https://github.com/ros2/rclpy/blob/62012d3650de790d89c725232451c4e998b396fb/rclpy/rclpy/time.py#L131) and store the duration instance in `self.delays`.

ros2topic/ros2topic/verb/delay.py
yechun1: this is original design, I have not changed while porting. If user did not set the value. As previous design, the default value of window is -1, and here will re-set to max 50000 as memory restrictions.

ros2topic/ros2topic/verb/delay.py
```diff
@@ -0,0 +1,241 @@
...
+                    self.delays = []
+                return
+
+            curr = curr_rostime.nanoseconds
+            if self.msg_t0 < 0 or self.msg_t0 > curr:
+                self.msg_t0 = curr
+                self.msg_tn = curr
+                self.delays = []
+            else:
+                # store the duration instrance in self.delays
```
dirk-thomas: Spelling: "instrance" -> "instance".

ros2topic/ros2topic/verb/delay.py
```diff
@@ -0,0 +1,241 @@
...
+                duration = (curr_rostime - Time.from_msg(msg.header.stamp))
+                self.delays.append(duration.nanoseconds * 1e-9)
```
dirk-thomas: This isn't doing what the comment two lines above states - it still stores the duration as a floating point number in seconds.

ros2topic/ros2topic/verb/delay.py
dirk-thomas: Sure, it would still benefit from being improved.

ros2topic/ros2topic/verb/delay.py
yechun1: This is original design. I think general user will not concern about window_size and just easy to use the tool. It will give user a little more trouble every time to type --window as required argument. It may also confuse user about the window meaning (first time I use rostopic, I really don't know what this window mean :))

Sorry but not considered when blocking=False, thanks for review, fixed by 0e79d06

ros2topic/ros2topic/verb/delay.py
yechun1: resubmit the commit 34e400a

ros2topic/ros2topic/verb/delay.py
yechun1: set blocking=True by default. commit again: c3e506f

yechun1:
@dirk-thomas commits with minor change be submitted. Do you have any other suggests? so that I could sync with you and fix immediately when you are online:)

dirk-thomas:
Thank you for the updates!

I just ran CI builds for this branch and the results look good:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5226/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1987/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4345/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5203/)

Can you please add the BSD license to the package manifest (as in https://github.com/ros2/ros2cli/pull/133/files#diff-27285248de861ae535a0217d2fa40c9cR11).

To keep the git history a bit shorter can you then reduce the commits as follows:
* keep the first two commits (1-to-1 copy of file, add the link to "upstream")
* squash the remaining commits into a third commit

After that this should be ready to be merged.

yechun1:
> To keep the git history a bit shorter can you then reduce the commits as follows:
    keep the first two commits (1-to-1 copy of file, add the link to "upstream")
    squash the remaining commits into a third commit

Add BSD license and merged the comments as 
b805ac6, 3fd69ec, a317f3c

clalancette:
CI looks good, and the patch set has been redone according to what @dirk-thomas suggested.  I also took a brief look at it, and it looks OK to me.  I'm going to merge this; thanks for the contribution.

yechun1:
@dirk-thomas @clalancette thanks for review and merge.

PR #142, opened 2018-08-29
Make `ros2 node list` include the full namespace

### LINKED ISSUE #105: ros2 node list doesn't include node namespace
Requires ros2/rmw#142

I first noticed this running a node with a remapped namespace, but it also happens if I hard-code a namespace. It happens:
- hard-coded or using static remapping
- if the node uses rclcpp or rclpy
- for a daemon using connext, fastrtps, or opensplice.

E.g.:
```
$ ros2 run demo_nodes_cpp talker  # (modified to use hard-coded namespace)
[INFO] [my_ns.talker]: Publishing: 'Hello World: 1'

$ ros2 node list
talker
```

I would consider this a bug because I expected the output to include the node's namespace, unless we are aware that this is not supported?

mjcarroll:
Fixes #105

### DIFF
--- a/ros2cli/ros2cli/daemon/__init__.py
+++ b/ros2cli/ros2cli/daemon/__init__.py
@@ -57,6 +57,8 @@ def main(*, script_name='_ros2_daemon', argv=None):
             server.register_function(
                 _print_invoked_function_name(node.get_node_names))
             server.register_function(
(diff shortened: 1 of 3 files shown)


dirk-thomas:
@mjcarroll Do you still need this branch?

PR #143, opened 2018-08-29
[ros2node] Make `ros2 node list` list namespaces as well as names

### LINKED ISSUE #105: ros2 node list doesn't include node namespace

mjcarroll:
Fixes #105 

To test:
```
ros2 run demo_nodes_cpp talker __ns:=/demo & 
ros2 run demo_nodes_cpp talker
```

Expected output:
```
➜ ros2 node list
/talker
/demo/talker
```

### DIFF
--- a/ros2cli/ros2cli/daemon/__init__.py
+++ b/ros2cli/ros2cli/daemon/__init__.py
@@ -57,6 +57,8 @@ def main(*, script_name='_ros2_daemon', argv=None):
             server.register_function(
                 _print_invoked_function_name(node.get_node_names))
             server.register_function(
+                _print_invoked_function_name(node.get_node_names_and_namespaces))
+            server.register_function(
                 _print_invoked_function_name(node.get_topic_names_and_types))
             server.register_function(
                 _print_invoked_function_name(node.get_service_names_and_types))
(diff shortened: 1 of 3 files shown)


ros2cli/ros2cli/daemon/__init__.py
```diff
@@ -56,6 +56,8 @@ def main(*, script_name='_ros2_daemon', argv=None):
             # expose getter functions of node
             server.register_function(
                 _print_invoked_function_name(node.get_node_names))
```
dirk-thomas: Should this function be removed since it only provides incomplete information?

mjcarroll: It is currently still being used by ros2lifecycle.  We haven't had a bug issued against that, but I'll take this as an opportunity to update that as well.

mjcarroll: That is, I'll remove it in the PR that updates ros2lifecycle.

dirk-thomas:
Superseded by #146.

PR #145, opened 2018-08-29
add ros2 multicast commands

dirk-thomas:
The `multicast` command contains two verbs:
* `receive` waiting for a single UDP multicast packet
* `send` sending a single UDP multicast packet

The tool can be used to check if multicast UDP packets are passed between two endpoints.

This package uses a `setup.cfg` file and therefore requires colcon/colcon-ros#38 (only when using `--symlink-install`).


dirk-thomas:
@mjcarroll Thanks for the review. In order to be mergable the referenced PR needs to be reviewed and merged though.

PR #147, opened 2018-09-07
add timestamp in the header of publish msg

yechun1:
publish msg with timestamp could be used to verify ros2opic delay command porting


ros2topic/ros2topic/verb/pub.py
```diff
@@ -90,6 +93,12 @@ def publisher(
 
     node = rclpy.create_node(node_name)
 
+    # Clock that has support for ROS time.
+    # TODO(dhood): use sim time if parameter has been set on the node.
+    _clock = ROSClock()
+    _time_source = TimeSource(node=node)
+    _time_source.attach_clock(_clock)
```
dirk-thomas: Instead of creating a new clock please use the clock already available in the node: `clock = node.get_clock()`

ros2topic/ros2topic/verb/pub.py
```diff
@@ -107,6 +116,9 @@ def timer_callback():
         count += 1
         if print_nth and count % print_nth == 0:
             print('publishing #%d: %r\n' % (count, msg))
+        curr_rostime = _clock.now().nanoseconds
+        msg.header.stamp.sec = int(curr_rostime * 1e-9)
+        msg.header.stamp.nanosec = int(curr_rostime % 1e9)
```
dirk-thomas: To avoid doing this "calculation" I added a new method `seconds_nanoseconds` on the `Time` class in #235. Please use that one:

```
msg.header.stamp.sec, msg.header.stamp.nanosec = clock.now().seconds_nanoseconds
```

yechun1:
@dirk-thomas thanks for review, updated the code. the patch depends on ros2/rclpy#235.

Pass verified on msg with header(sensor_msgs/Image) and without header(std_msgs/String)
$ ros2 topic pub /msg sensor_msgs/Image

> publisher: beginning loop
> publishing #1: sensor_msgs.msg.Image(header=std_msgs.msg.Header(stamp=builtin_interfaces.msg.Time(sec=0, nanosec=0), frame_id=''), height=0, width=0, encoding='', is_bigendian=0, step=0, data=[])
> publishing #2: sensor_msgs.msg.Image(header=std_msgs.msg.Header(stamp=builtin_interfaces.msg.Time(sec=1536561756, nanosec=761655296), frame_id=''), height=0, width=0, encoding='', is_bigendian=0, step=0, data=[])
> publishing #3: sensor_msgs.msg.Image(header=std_msgs.msg.Header(stamp=builtin_interfaces.msg.Time(sec=1536561757, nanosec=761586944), frame_id=''), height=0, width=0, encoding='', is_bigendian=0, step=0, data=[])
> 

$ ros2 topic pub /msg std_msgs/String

> publisher: beginning loop
> publishing #1: std_msgs.msg.String(data='')
> publishing #2: std_msgs.msg.String(data='')
> publishing #3: std_msgs.msg.String(data='')

dirk-thomas:
Having the option to set the header timestamp to the current is great. If the user wan't to publish a specific message with a specific timestamp though this isn't possible anymore. I would recommend to add an option to conditionally control this "magic" step. Opt out is likely a good choice.

yechun1:
when user set specific message, the time basically will not set to zero. Below is changes to fill timestamp if msg has header but the timestamp is not zero. is this ok? 
```

--- a/ros2topic/ros2topic/verb/pub.py
+++ b/ros2topic/ros2topic/verb/pub.py
@@ -103,14 +103,24 @@ def publisher(
     print('publisher: beginning loop')
     count = 0
 
+    # check if the msg has header, and the header has specific timestamp
+    timestamp_auto_fill = False
+    if hasattr(msg, 'header'):
+       if (msg.header.stamp.sec == 0) and (msg.header.stamp.nanosec == 0):
+           timestamp_auto_fill = True
+       else:
+           timestamp_auto_fill = False
+
     def timer_callback():
         nonlocal clock
         nonlocal count
+        nonlocal timestamp_auto_fill
         count += 1
         if print_nth and count % print_nth == 0:
             print('publishing #%d: %r\n' % (count, msg))
 
-        if hasattr(msg, 'header'):
+        # add timestamp as ROS time if msg with header has no specific timestamp
+        if timestamp_auto_fill:
             msg.header.stamp.sec, msg.header.stamp.nanosec = clock.now().seconds_nanoseconds()
         pub.publish(msg)

```

yechun1:
Oh. Do you mean to add optional argument like "--timestamp" "-t" ?

dirk-thomas:
> Do you mean to add optional argument like "--timestamp" "-t" ?

Yes, the ROS 1 tool `rostopic pub` offer the option `-s` / `--substitute-keywords`:

> When publishing with a rate, performs keyword ('now' or 'auto') substitution for each message

Since that seems to work fine it might be best to follow this approach.

dirk-thomas:
I would suggest to rather implement the same option as in ROS 1 as it is strictly more flexible / capable

ros2topic/ros2topic/verb/pub.py
dhood: I would think you'd use [`now().to_msg()`](https://github.com/ros2/rclpy/pull/235/files#diff-cf0da220e7cfb92120fc2d55444635c4R129) here instead

yechun1:
the `-s` implementation depended on genpy library which used [ `genpy.message.fill_message_args()`](https://github.com/ros/ros_comm/blob/melodic-devel/tools/rostopic/src/rostopic/__init__.py#L1721). Is there any replacement api on ROS2 now? Do we have plan in the future to port ROS messages and services generators (genpy/gencpp/genlisp ...)?

dirk-thomas:
Message generation is very different between the ROS versions. The packages in ROS 2 are called e.g. `rosidl_generator_cpp|py`. If some of the logic from these ROS 1 packages is needed in ROS 2 then it can be ported to the existing ROS 2 packages. Please consider to create a pull request for those of they could be used for this ticket

yechun1:
This patch is not necessary. closed it.

PR #149, opened 2018-09-12
comply with new node representation

Karsten1987:
I think the `ros2lifecycle` is currently not working currently since https://github.com/ros2/ros2cli/pull/146 got merged.

### DIFF
--- a/ros2lifecycle/ros2lifecycle/api/__init__.py
+++ b/ros2lifecycle/ros2lifecycle/api/__init__.py
@@ -27,7 +27,7 @@ def get_node_names(*, node, include_hidden_nodes=False):
         node=node, include_hidden_nodes=include_hidden_nodes)
     service_names_and_types = get_service_names_and_types(node=node)
     return [
-        n.full_name for n in node_names
+        n for n in node_names
         if _has_lifecycle(n.full_name, service_names_and_types)]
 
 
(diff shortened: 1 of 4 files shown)


mjcarroll:
Yes, I was supposed to follow up on that, I didn't realize that #146 got merged.  Sorry for the inconvenience.

ros2lifecycle/ros2lifecycle/verb/set.py
```diff
@@ -44,7 +44,12 @@ def main(self, *, args):  # noqa: D102
             node_names = get_node_names(
                 node=node, include_hidden_nodes=args.include_hidden_nodes)
 
-        if args.node_name not in node_names:
+        node_found = False
+        for node_name in node_names:
+            if args.node_name in node_name.full_name:
+                node_found = True
+                break
+        if not node_found:
```
dirk-thomas: This could be much shorter: `if args.node_name not in {n.full_name for n in node_names}:`

Karsten1987: of course there is a one-liner for python :) thanks!

ros2lifecycle/ros2lifecycle/api/__init__.py
```diff
@@ -27,14 +27,14 @@ def get_node_names(*, node, include_hidden_nodes=False):
         node=node, include_hidden_nodes=include_hidden_nodes)
     service_names_and_types = get_service_names_and_types(node=node)
     return [
-        n for n in node_names
-        if _has_lifecycle(n, service_names_and_types)]
+        n.full_name for n in node_names
```
dirk-thomas: Would it be more consistent with the same function make in `ros2node` to return the "full" `n` here and let the callers pick the`full_name` from it?

Karsten1987: I can change it if you want. But given that this function anyway only really serves for the lifecycle implementation i didn't do it at first.

Issue #154, opened 2018-10-31
String formatting for ros2 topic echo


stonier:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - Binaries & Source
- Version or commit hash:
  - 72961082678fa3491b2cf778dc004bf820cd4c7c
- DDS implementation:
  - Default
- Client library (if applicable):
  - N/A

#### Steps to reproduce issue

Talker:

```
#!/usr/bin/env python3

import rclpy
import rclpy.node
import std_msgs.msg as std_msgs

class Talker(rclpy.node.Node):

    def __init__(self):
        super().__init__('talker')
        self.pub = self.create_publisher(std_msgs.String, 'chatter')
        timer_period = 1.0
        self.tmr = self.create_timer(timer_period, self.timer_callback)
        self.toggle = False
        self.msg_with_newlines = std_msgs.String(
            data="Dude\nand\nDudettes"
        )
        self.msg_with_escape_sequences = std_msgs.String(
            data="\x1b[34mDude\x1b[0m\nand\nDudettes"
        )

    def timer_callback(self):
        msg = std_msgs.String()
        if self.toggle:
            msg = self.msg_with_escape_sequences
        else:
            msg = self.msg_with_newlines
        self.toggle = not self.toggle
        print("-----------\n{}".format(msg.data))
        self.pub.publish(msg)

if __name__ == '__main__':
    rclpy.init()
    node = Talker()
    try:
        rclpy.spin(node)
    except KeyboardInterrupt:
        pass
    node.destroy_node()
    rclpy.shutdown()
```

Execution:

```
# In a first shell
$ ./talker
# in a second shell
$ ros2 topic echo /chatter
data: 'Dude

  and

  Dudettes'
data: "\e[34mDude\e[0m\nand\nDudettes"
```

#### Expected behavior

String formatting that pays attention to newlines, tabs, console sequences etc (as per ROS1).

#### Actual behavior

No string formatting at all.

#### Additional information

I expect there is something going topsy turvy in the yaml.dump mechanism given the changes in string handling by yaml from Python2 - Python 3

stonier:
Some digging - might be that the `yaml.dump` of an `OrderedDict` with string values, doesn't treat them as double quoted strings. In this situation it ignores both special characters and escape sequences.

[...]

PR #155, opened 2018-11-01
Block literal printing for std_msgs.msg.String

stonier:
**Tentative!**

I really don't like this hack and would like to trigger a discussion (hopefully the ugliness will encourage that!).

**Goal**

Enable `ros2cli topic echo /chatter` to format the string messages that are sent across the nether so that we can have easy-to-read informative latched publishers that help describe / introspect the system at runtime. Currently the strings are completely unformatted (in particular, newlines and console sequences for colouring would be handy). 

**Background**

ROS1 could do this and it was very useful. It didn't take advantage of `yaml.dump` though, so no assistance there.

**Problems**

As mentioned in #154, we could use a customs str representer that sets an appropriate block literal style, but yaml will override our override in certain situations. So, as a last resort, the patch here would be the minimal that will at least enable formatting for strings in `std_msgs.msg.String` (as opposed to any embedded string) by avoiding any entanglement with yaml altogether.

**Considerations**

* Effort upstream to get things fixed in yaml, that's not going to help in the short term though
* Could alternatively use the representers as demonstrated in #154 and monkeypatch the yaml `Emitter` class here as a temporary fix (dangerous, since the Emitter class could change without our knowing)
* If anything temporary is done, make sure to have comments directing people to the golden path when it is available.
* Other?

### DIFF
--- a/ros2topic/ros2topic/verb/echo.py
+++ b/ros2topic/ros2topic/verb/echo.py
@@ -25,6 +25,7 @@ from ros2cli.node.direct import DirectNode
 from ros2topic.api import get_topic_names_and_types
 from ros2topic.api import TopicNameCompleter
 from ros2topic.verb import VerbExtension
+from std_msgs.msg import String
 import yaml
 
 DEFAULT_TRUNCATE_LENGTH = 128
@@ -140,7 +141,17 @@ def subscriber(node, topic_name, message_type, callback):
 def subscriber_cb(args):
     def cb(msg):
         nonlocal args
-        print(msg_to_yaml(args, msg))
+        if (type(msg) == String):
+            # emulate block literal yaml format, but accept unicode
+            # chars that yaml does not yet
+            if len(msg.data.splitlines()) > 1:
+                print("data: |-" )
+                for line in msg.data.splitlines():
+                    print("  {}".format(line))
+            else:
+                print("data: {}".format(msg.data))
+        else:
+            print(msg_to_yaml(args, msg))
     return cb
 
 


clalancette:
After looking at this, it seems to me that the utility of having `std_msg.msg.String` does not outweigh the complexity of explaining why the strings look different in different situations.  Additionally, it seems to me that if we were to add in pretty-printing of strings, we would also want some way to print out the "raw" versions of the strings as well, to help with debugging in some situations.

@stonier Is https://github.com/yaml/pyyaml/pull/228 plus the representer changes in https://github.com/ros2/ros2cli/issues/154#issuecomment-435024555 enough to make this work for all strings?

stonier:
@clalancette sorry, missed the notification for your question.

The changes here won't work for all strings, just messages of the type `std_msgs.msg.String`. It won't do anything for an `std_msgs.msg.String` field embedded in another message type. In light of that, it's just throwing a cat in amongst the pigeons. I don't like this PR.

It's also worth noting that the approach used here in roscli seems to have been adopted even before ROS2. A similar change was made to genpy [here](https://github.com/ros/genpy/pull/79) in an effort to make sure that the output of rostopic echo could be piped into a publisher. That's a good use case to support, but it also means there that any string formatting in ROS1 (even basic newlines and tabs) is also unreadable on the other end of rostopic echo there.

----

Fundamentally though, not being able to easily introspect your system via formatted text on publishers makes me very sad. Whether it's lining up columns of serial packet data so you can easily catch the bad byte, or putting statistical data from your algorithm in rows and columns, being able to interpret that with minimal effort was a great time saver.

stonier:
Dropping this as discussion towards a solution has not really moved forward. Will need a rethink.

PR #159, opened 2018-11-16
Add ros2 node info verb

ross-desmond:
*Summary:* Provide the info verb under the node command

`ros2 node info /minimal_publisher`

Example:
```
/minimal_publisher
  Subscribers:
    /parameter_events: rcl_interfaces/ParameterEvent
  Publishers:
    /parameter_events: rcl_interfaces/ParameterEvent
    /topic: std_msgs/String
  Services:
    /minimal_publisher/describe_parameters: rcl_interfaces/DescribeParameters
    /minimal_publisher/get_parameter_types: rcl_interfaces/GetParameterTypes
    /minimal_publisher/get_parameters: rcl_interfaces/GetParameters
    /minimal_publisher/list_parameters: rcl_interfaces/ListParameters
    /minimal_publisher/set_parameters: rcl_interfaces/SetParameters
    /minimal_publisher/set_parameters_atomically: rcl_interfaces/SetParametersAtomically
```

Connects to ros2/rcl#333

### DIFF
--- a/ros2node/ros2node/api/__init__.py
+++ b/ros2node/ros2node/api/__init__.py
@@ -20,7 +20,19 @@ from ros2cli.node.strategy import NodeStrategy
 HIDDEN_NODE_PREFIX = '_'
 
 NodeName = namedtuple('NodeName', ('name', 'namespace', 'full_name'))
+TopicInfo = namedtuple('Topic', ('name', 'types'))
 
+
+def parse_node_name(full_node_name):
+    tokens = full_node_name.split("/");
+    if 1 > len(tokens):
+        raise Exception("Invalid node name")
+    node_name = full_node_name
+    namespace = "/"
+    if len(tokens) > 1:
+        node_name = tokens[-1]
+        namespace = "/".join(tokens[:-1])
+    return NodeName(node_name, namespace, full_node_name)
 
 def get_node_names(*, node, include_hidden_nodes=False):
     node_names_and_namespaces = node.get_node_names_and_namespaces()
@@ -36,6 +48,23 @@ def get_node_names(*, node, include_hidden_nodes=False):
         )
     ]
 
+def get_topics(remote_node_name, func):
+    node = parse_node_name(remote_node_name)
+    names_and_types = func(node.name, node.namespace)
+    return [
+        TopicInfo(
+            name=t[0],
+            types=t[1])
+        for t in names_and_types]
+
+def get_subscriber_info(*, node, remote_node_name):
+    return get_topics(remote_node_name, node.get_subscriber_names_and_types_by_node)
+
+def get_publisher_info(*, node, remote_node_name):
+    return get_topics(remote_node_name, node.get_publisher_names_and_types_by_node)
+
+def get_service_info(*, node, remote_node_name):
+    return get_topics(remote_node_name, node.get_service_names_and_types_by_node)
 
 class NodeNameCompleter:
     """Callable returning a list of node names."""
--- a/ros2node/ros2node/api/__init__.py
+++ b/ros2node/ros2node/api/__init__.py
@@ -24,15 +24,16 @@ TopicInfo = namedtuple('Topic', ('name', 'types'))
 
 
 def parse_node_name(full_node_name):
-    tokens = full_node_name.split("/");
+    tokens = full_node_name.split('/')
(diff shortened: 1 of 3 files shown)


ross-desmond:
Connected to [rcl](https://github.com/ros2/rcl/pull/333)

thomas-moulard:
@ross-desmond PTAL at https://github.com/thomas-moulard/ros2cli/tree/node_graph_impl - this fixes issues with flake8
Confirmed tests are passing locally.

ros2node/ros2node/api/__init__.py
```diff
@@ -20,6 +20,19 @@
 HIDDEN_NODE_PREFIX = '_'
 
 NodeName = namedtuple('NodeName', ('name', 'namespace', 'full_name'))
+TopicInfo = namedtuple('Topic', ('name', 'types'))
+
+
+def parse_node_name(full_node_name):
+    tokens = full_node_name.split('/')
+    if 1 > len(tokens):
+        raise Exception('Invalid node name')
```
clalancette: I don't see any other uses of a bare `Exception` being thrown here.  This should probably be `RuntimeError` to conform to the rest of the code.  Also, it would be helpful to append the name of the invalid node to the error message.

ros2node/ros2node/verb/info.py
```diff
@@ -0,0 +1,57 @@
...
+                print('  Publishers:')
+                print_names_and_types(publishers)
+                services = get_service_info(node=node, remote_node_name=args.node_name[1:])
+                print('  Services:')
+                print_names_and_types(services)
+        else:
+            print('All nodes found:')
+            print(*[n.full_name for n in node_names], sep=', ')
+            print('Unable to find node ' + args.node_name)
+            return
```
clalancette: Fairly minor, but it looks like the return value from `main` is used to determine whether the command "succeeded" or "failed".  In this `else` clause, we probably want to `return 'Unable to find node'` so that it is a failure case.

ros2node/ros2node/api/__init__.py
ross-desmond: On it

ros2node/ros2node/verb/info.py
ross-desmond: Awesome

clalancette: I ended up changing this again in eb3dc36 (sorry I wasn't more clear the first time).  Two things: the command-line tools already print out whatever you return here, and the full list of node names is already available through `ros2 node list -a`.  Thus, I simplified this to just a return error message.

clalancette:
OK, I figured out the segfault and fixed it in https://github.com/ros2/rclpy/pull/247/commits/dc8edb3bece8c0f831ff393d71e5ee08e3e61ce8 .  So I'll approve this now.

clalancette:
OK, I figured out the segfault and fixed it in https://github.com/ros2/rclpy/pull/247/commits/dc8edb3bece8c0f831ff393d71e5ee08e3e61ce8 .  So I'll approve this now.

clalancette:
OK, I figured out the segfault and fixed it in https://github.com/ros2/rclpy/pull/247/commits/dc8edb3bece8c0f831ff393d71e5ee08e3e61ce8 .  So I'll approve this now.

ros2node/ros2node/api/__init__.py
```diff
@@ -20,6 +20,19 @@
 HIDDEN_NODE_PREFIX = '_'
 
 NodeName = namedtuple('NodeName', ('name', 'namespace', 'full_name'))
+TopicInfo = namedtuple('Topic', ('name', 'types'))
+
+
+def parse_node_name(full_node_name):
+    tokens = full_node_name.split('/')
+    if 1 > len(tokens):
```
jacobperron: Maybe we should have a guard against cases for empty string or "'/'" instead? I think `split` will always return at least one element, even if it is the empty string.

ros2node/setup.py
```diff
@@ -34,6 +34,7 @@
         ],
         'ros2node.verb': [
             'list = ros2node.verb.list:ListVerb',
+            'info = ros2node.verb.info:InfoVerb',
```
jacobperron: nit: perhaps this should be alphabetized

PR #162, opened 2018-12-03
Make listing hidden services symmetric with topics

sloretz:
The `ros2 service` command and `ros2 service list` command both had `--include-hidden-services` arguments. Because of the latter the first argument did nothing. This makes listing hidden services the same as listing hidden topics

Before

``` bash
# this would show no error, and not show hidden services
ros2 service --include-hidden-services list
# this would show hidden services
ros2 service list --include-hidden-services
```

With this PR
```bash
# this shows hidden services.
ros2 service --include-hidden-services list
# this shows hidden topics
ros2 topic --include-hidden-topics list
```


sloretz:
CI only linux for linters [Build Status](https://ci.ros2.org/job/ci_linux/5740/)

PR #177, opened 2018-12-13
use collections.abc.Iterable

dirk-thomas:
Same as ros2/launch#158.

### DIFF
--- a/ros2msg/test/test_api.py
+++ b/ros2msg/test/test_api.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from collections import Iterable
+from collections.abc import Iterable
 import os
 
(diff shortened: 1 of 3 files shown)


dirk-thomas:
> Needs to be tested against Xenial, imo.

Can you elaborate why? `collections.abc` exists since Python 3.3 and Xenial uses Python 3.4.

wjwwood:
Was `Iterable` in `collections.abc`? I dunno I guess you can also just check the documentation to see if it should work or not. I just think running a Xenial job up to launch would accomplish the same thing and be certain.

dirk-thomas:
Confirmed in a Xenial Docker:

```
# python3 --version
Python 3.5.2
# python3 -c "import collections.abc; print(collections.abc.Iterable)" 
<class 'collections.abc.Iterable'>
```

PR #174, opened 2018-12-11
Remove unused cli option

Karsten1987:
fixes ros2/ros2cli#170

has to be rebased when ros2/ros2cli#167 is merged.

This PR basically removed the obsolete argument `--transitions` for the `ros2 lifecycle get` command. The transitions can be displayed by invoking `ros2 lifecycle list [-a] <node_name>`

Issue #178, opened 2018-12-13
[ros2param] ros2 param list requires fully qualified node name


sloretz:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu bionic
- Installation type:
  -  debs
- Version or commit hash:
  - crystal prerelease
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - N/A

#### Steps to reproduce issue

1. Save the following as `demo_params.yaml`
   ```
   talker:
       ros__parameters:
           some_int: 42
           a_string: "Hello world"
           some_lists:
               some_integers: [1, 2, 3, 4]
               some_doubles : [3.14, 2.718]
   ```
1. Launch a node with those parameters
   ```
   ros2 run demo_nodes_cpp talker __params:=demo_params.yaml
   ```
1. List the parameters using a fully qualified node name
   ```
   ros2 param list /talker
   ```
1. List the parameters using a non fully qualified node name
   ```
   ros2 param list talker
   ```

#### Expected behavior

Both commands would list the parameters on the talker node. In the first case the name is fully qualified already so the node to list is unambiguous. In the second case the namespace of the CLI tool is `/` by default, so it should list the params for `/` + `talker`. 

#### Actual behavior

The non-fully qualified version can't find the node.

```bash 
$ ros2 param list /talker
  a_string
  some_int
  some_lists.some_doubles
  some_lists.some_integers
$ ros2 param list talker
Node not found
```

Karsten1987:
we could address this in https://github.com/ros2/ros2cli/pull/179
opinions?

Karsten1987:
we could address this in #179
opinions?

PR #179, opened 2018-12-13
add slash for node name

Karsten1987:
same fix as for get: https://github.com/ros2/ros2cli/pull/174

### DIFF
--- a/ros2lifecycle/ros2lifecycle/verb/get.py
+++ b/ros2lifecycle/ros2lifecycle/verb/get.py
@@ -20,6 +20,7 @@ from ros2cli.node.strategy import NodeStrategy
 from ros2lifecycle.api import call_get_states
 from ros2lifecycle.api import get_node_names
 from ros2lifecycle.verb import VerbExtension
+from ros2node.api import get_absolute_node_name
 from ros2node.api import NodeNameCompleter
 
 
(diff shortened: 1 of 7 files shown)


Karsten1987:
the latest commit (https://github.com/ros2/ros2cli/pull/179/commits/be9e65a45d490b387c90f362305ca6d370cc7ac9) should also address https://github.com/ros2/ros2cli/issues/178

nuclearsandwich:
It seems a bit odd to make this same transformation in every different verb. Is there a reason not to create a `normalize_node_name` or somesuch function?

Karsten1987:
that is correct. I think we should actually have a function for this in rcl expanding the node name in a similar fashion as we do with topics (see https://github.com/ros2/rcl/blob/master/rcl/include/rcl/expand_topic_name.h) But given the time, I opted for this solution.

wjwwood:
There could be a function to do this in rcl and it could be exposed via rclpy, like the expand topic name function:

https://github.com/ros2/rclpy/blob/91f7782619b1ccd9c812012ff85f6336186baab9/rclpy/rclpy/expand_topic_name.py#L18-L33

Karsten1987:
So my referenced function does actually not really help. It does pretty much the opposite. I guess I stick with my reasoning then that the rcl solution would be preferred but given the time I think it's not feasible.

Karsten1987:
I just introduced a new small function in `ros2node` called `get_absolute_node_name`. It's really just checking for the leading slash because I think we can all agree that a real node name expansion function should be there at one point (I'll create a ticket for it).

@wjwwood @jacobperron please iterate one more time over this. Also feel free to directly push on this ticket if you see room for improvement :)
thanks!

nuclearsandwich:
Whs the `ros2 node info` verb intentionally omitted from this feature?

https://github.com/ros2/ros2cli/blob/aa832722af1c6295eb4b55951b5cd6a50c25108e/ros2node/ros2node/verb/info.py

PR #187, opened 2019-01-06
Move --include-hidden-topic to list verb

### LINKED ISSUE #182: [ros2{topic,service}] the --include-hidden-* option should be an option of list verb not command
## Bug report

**Required Info:**

- Operating System:
  - any
- Installation type:
  - any
- Version or commit hash:
  - Crystal release candidate
- DDS implementation:
  - any
  - N/A

#### Steps to reproduce issue

```
$ ros2 topic -h
$ ros2 topic list -h
```

#### Expected behavior

Options that only affect `list` verb would show up in its own help and be passed after the verb. At the very least it should be seen as an option when you do `ros2 topic list -h` (for example).

#### Actual behavior

Option only appears in command help, e.g. `ros2 topic -h`.

#### Additional information

I'm not sure why it's this way in the first place, possibly because the option used to affect the echo/call verbs. But right now it's definitely confusing and difficult to discover this option.

rarvolt:
This fixes #182

### DIFF
--- a/ros2topic/ros2topic/command/topic.py
+++ b/ros2topic/ros2topic/command/topic.py
@@ -22,9 +22,6 @@ class TopicCommand(CommandExtension):
 
     def add_arguments(self, parser, cli_name):
         self._subparser = parser
-        parser.add_argument(
-            '--include-hidden-topics', action='store_true',
-            help='Consider hidden topics as well')
 
         # get verb extensions and let them add their arguments
         verb_extensions = get_verb_extensions('ros2topic.verb')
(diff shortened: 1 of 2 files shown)


dirk-thomas:
See https://github.com/ros2/ros2cli/issues/182#issuecomment-462004052 why this patch would break the functionality for all the other verbs.

dirk-thomas:
@rarvolt Thank you for your effort on this!

After iterating on the issue I will close this in favor of #196.

PR #188, opened 2019-01-07
register 'get_node_names' function, used by ros2cli node info tool.

sagniknitr:
The PR will solve python error which occures on running the `ros2cli node info` tool.


dirk-thomas:
Please edit the title of the PR to reflect the content since when doing a squash merge it will become the commit message.

dirk-thomas:
I tried to use `ros2 node info` with Crystal and it seems to work just fine. Can you please provide more context what the problem is, where this function is used and the exact commands you used to reproduce.

sagniknitr:
@dirk-thomas closing this request as unable to reproduce the error. Yes, I double checked `ros2 node info` , and it is working fine for Crystal. Thanks for your help

PR #194, opened 2019-02-06
use all linters by default

dirk-thomas:
Imo all linters should be enabled by default with the comment being available how to skip them. Otherwise the high probability is that they won't we used by users which just expand the template but don't edit the file.


dirk-thomas:

PR #195, opened 2019-02-07
[ros2service] auto-detect the service type

stonier:
Some uncertainties...

* Is there a situation in which the service type needs to be provided?
* Is there a situation in which the retrieved names need to be expanded as [here](https://github.com/ros2/ros2cli/blob/master/ros2topic/ros2topic/verb/echo.py#L103)?

### DIFF
--- a/ros2service/ros2service/verb/call.py
+++ b/ros2service/ros2service/verb/call.py
@@ -15,8 +15,10 @@
 import importlib
 import time
 
+import os
 import rclpy
 from ros2cli.node import NODE_NAME_PREFIX
+from ros2service.api import get_service_names_and_types
 from ros2service.api import ServiceNameCompleter
 from ros2service.api import ServiceTypeCompleter
 from ros2service.verb import VerbExtension
@@ -34,9 +36,6 @@ class CallVerb(VerbExtension):
             help="Name of the ROS service to call to (e.g. '/add_two_ints')")
         arg.completer = ServiceNameCompleter(
             include_hidden_services_key='include_hidden_services')
-        arg = parser.add_argument(
-            'service_type',
-            help="Type of the ROS service (e.g. 'std_srvs/Empty')")
         arg.completer = ServiceTypeCompleter(
             service_name_key='service_name')
         parser.add_argument(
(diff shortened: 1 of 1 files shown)


stonier:
That's curious. Those are errors from `ros2topic` / `ros2cli.node` / `rclpy`, not `ros2service`.

clalancette:
@stonier I'm pretty sure that the problem is that this branch is targeting master, but the PR job is targeting Crystal.  There's been some things unrelated to this PR that got merged that essentially break that situation, so no worries on the PR job failing.  Once we do a bit of reviewing, someone will manually kick off a CI job for it.

ros2service/ros2service/verb/call.py
```diff
@@ -53,11 +52,29 @@ def main(self, *, args):
...
+def requester(service_name, values, period):
+    rclpy.init()
 
+    node = rclpy.create_node(NODE_NAME_PREFIX + '_requester_%s' % (str(os.getpid())))
+    names_and_types = get_service_names_and_types(
+        node=node
+    )
+    for n, t in names_and_types:
+        if n == service_name:
+            if len(t) > 1:
```
clalancette: Does it make any sense to allow the user to pass the service type as an option?  Would that allow us to resolve this ambiguity in the case that the service has multiple types?

stonier:
Updated the link above to properly point at the way `ros2 topic echo` handles this similarly for topics ([link](https://github.com/ros2/ros2cli/blob/master/ros2topic/ros2topic/verb/echo.py#L103)).

ros2service/ros2service/verb/call.py
stonier: Aye, I was very surprised to find a list being returned by `get_service_names_and_types()`. `ros2 topic echo` doesn't handle it so I'm curious if it is a use case I'm not aware of or just some relic of a previous architectural design.

stonier: If we can't answer the question, we could a) re-insert the argument optionally to disambiguate, just in case, but I don't like adding complexity without understanding why or b) add a comment to advise if someone actually runs into the use case we are not aware of

clalancette: I discussed this with @stonier , and we came to the agreement that the current code is fine.  If we do end up in the situation where someone runs into the `RuntimeError` below, we can always lift the restriction then and add an option.

ros2service/ros2service/verb/call.py
```diff
@@ -34,9 +36,6 @@ def add_arguments(self, parser, cli_name):
             help="Name of the ROS service to call to (e.g. '/add_two_ints')")
         arg.completer = ServiceNameCompleter(
             include_hidden_services_key='include_hidden_services')
-        arg = parser.add_argument(
-            'service_type',
-            help="Type of the ROS service (e.g. 'std_srvs/Empty')")
         arg.completer = ServiceTypeCompleter(
```
dirk-thomas: This line adds a completer for the `service_type` argument. With the removal of the argument it now overwrites the completer of the previous field.

dirk-thomas:
I still think the `service_type` argument is important. E.g. if we want to provide completion of the actual `values` knowing the service type without having to query the graph will be important for the completion speed. Therefore I would suggest to modify the patch to continue to pass the `service_type` and make it optional instead.

Side question: why is removing the argument actually "needed"? The argument does have auto completion so the user doesn't have to type it...

ros2service/ros2service/verb/call.py
stonier: Aye, of course. Too hasty first time through.

stonier:
Updated as an **optional** argument (`'-t', '--service-type'`) for a) speed in scripted calls and b) when multiple types exist.

**Needed**: No, merely nice-to-have. If 99% of use cases don't exercise it and you use it often enough,  even the convenience of tab-completion seems like an exercise in futility.

Having said that, this now represents an inconsistency in the way types are handled across the `ros2cli` tools. It should either get adopted everywhere OR ... we just endeavour to find satisfaction in the rigour of [completeness](https://en.wikipedia.org/wiki/Completeness) via completion!

I think I'll close this one out since it's not worth the disruption.

PR #200, opened 2019-02-18
fixup ros2multicast version

mikaelarguedas:
to match https://github.com/ros2/ros2cli/blob/66efd10c61bd54ebdff0035102b9878a2acf5b4b/ros2multicast/package.xml#L5


dirk-thomas:
Thanks for the patch.

PR #204, opened 2019-03-09
Added type information to "ros2 topic info"

timobanziger:
When inspecting topics it is useful to get also information about the message type of a topic.

Basically adapted a few lines of code from "ros2 topic echo"

### DIFF
--- a/ros2topic/ros2topic/verb/info.py
+++ b/ros2topic/ros2topic/verb/info.py
@@ -14,6 +14,7 @@
 
 from ros2cli.node.direct import DirectNode
 from ros2topic.api import TopicNameCompleter
+from ros2topic.api import get_topic_names_and_types
 from ros2topic.verb import VerbExtension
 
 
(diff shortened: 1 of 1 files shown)


Karsten1987:
Thanks for the patch. Just for my understanding, is it important to have this information during `echo`? Couldn't you get that information with `ros2 topic list -t` ?

ros2topic/ros2topic/verb/info.py
```diff
@@ -30,6 +31,21 @@ def add_arguments(self, parser, cli_name):
     def main(self, *, args):
         with DirectNode(args) as node:
             topic_name = args.topic_name
+            type_name = "NOT FOUND!"
```
dirk-thomas: In the case of the topic not being found I would rather expect a non-zero return code (instead of the usual output with a different string as `Type: `).

ros2topic/ros2topic/verb/info.py
```diff
@@ -30,6 +31,21 @@ def add_arguments(self, parser, cli_name):
...
+
+            all_topic_names_and_types = get_topic_names_and_types(node=node, include_hidden_topics=True)
+            for topic, type in all_topic_names_and_types:
```
dirk-thomas: Please don't use the name of a builtin as a variable name.

dirk-thomas:
@timobanziger What is the status of this PR? Will you be able to address the feedback in the near future?

artivis:
Isn't this feature the equivalent to `rostopic type /my_topic` ??  
Could one imagine a `ros2 topic type \my_topic` keyword to better mirror ROS 1  ?  
Up for helping out.

dirk-thomas:
> Isn't this feature the equivalent to `rostopic type /my_topic` ??

While in ROS 1 `rostopic info` as well as `rostopic type` output the type of a topic both verb makes sense and are useful in different cases.

> Could one imagine a `ros2 topic type \my_topic` keyword to better mirror ROS 1 ?

Absolutely, that verb could be added beside having `info` also output the topic type (beside other information).

> Up for helping out.

Please consider to contribute the `type` verb in a separate pull request. We would be more than happy to review / merge that improvement.

dirk-thomas:
@timobanziger Friendly ping.

ros2topic/ros2topic/verb/info.py
sloretz: Recommend `type` -> `type_` per [PEP 8](https://www.python.org/dev/peps/pep-0008/#id36)

> single_trailing_underscore_: used by convention to avoid conflicts with Python keyword [...]

clalancette:
This functionality was put in via a different way, so I'm going to close this one out.

PR #212, opened 2019-03-28
use yaml.safe_load instead of deprecated yaml.load

mikaelarguedas:
Similar to colcon/colcon-metadata#14


dirk-thomas:
@mjcarroll Are you planning on triggering CI for this?

:+1: from me (to merge without CI).

Issue #202, opened 2019-03-04
Add CLI tool for Actions


jacobperron:
#### Feature description

Analogous to *topics* and *services*, there should be a command line tool for **actions**:

        ros2 action

Here is a list of tool capabilities and proposed signatures:

- list action names associated with any running action servers or action clients

        # For example
        $ ros2 action list
        /fibonacci
        /some/name/space/my_action

- list action servers and action clients associated with an action name

        ros2 action info ACTION_NAME
        # For example:
        $ ros2 action info /fibonacci
        Action: /fibonacci
        Action Clients: 2
            /node/foo
            /another/node/bar 
        Action Servers: 1
            /baz
        
- display active goals on an action server

        ros2 action goals ACTION_NAME
        # For example
        $ ros2 action goals /fibonacci
        152790028afb49a3b3c61383e6b97320
        8c3449e71c9941ed8ad3d68fdd7f553c
        b07cb18ca2dc4cca95acfc131fc49223

- display the arguments for an action goal

        ros2 action goal info ACTION_NAME GOAL_UUID
        # For example:
        $ ros2 action goal info /fibonacci 152790028afb49a3b3c61383e6b97320
        Action: /fibonacci
        Goal ID: 152790028afb49a3b3c61383e6b97320
        order: 10
        
- display the type of an action's goal, feedback, and result

        ros2 action show ACTION_TYPE
        # For example:
        $ ros2 action show example_interfaces/Fibonacci
        # Goal
        int32 order
        ---
        # Result
        int32[] sequence
        ---
        # Feedback
        int32[] sequence

- find actions by action type

        # Accomplished with a ros2 action list option:
        $ ros2 action list --show-types
        /fibonacci [example_interfaces/Fibonacci]
        /some/name/space/my_action [test_msgs/NestedMessage]

- echo feedback, status, and result for an action goal

        ros2 action echo [--feedback] [--status] [--result] ACTION_NAME GOAL_UUID
        # For example, getting feedback, status, and result messages:
        $ ros2 action echo /fibonacci 152790028afb49a3b3c61383e6b97320
        Feedback:
            sequence: [0, 1, 1]
        Feedback:
            sequence: [0, 1, 1, 2]
        Feedback:
            sequence: [0, 1, 1, 2, 3]
        Result:
            sequence: [0, 1, 1, 2, 3]
        Status: SUCCESS
        # For example, only echo status and result messages:
[...]

PR #214, opened 2019-04-03
Add Action CLI

jacobperron:
Created a new package, `ros2action`, defining the command `action`. This PR adds implementation for the following verbs:

```sh
  info       Print information about an action
  list       Output a list of action names
  send_goal  Send an action goal
  show       Output the action definition
```

Connects to ros2/ros2cli#202
Requires ros2/rcl#411
Requires ros2/rclpy#306


jacobperron:
I still need to add auto-completion and do more testing.

Running CI to find any issues early:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6593/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2931/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5412/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6417/)

Karsten1987:
Not sure how trivial it is, but somehow echoing the feedback after submitting a goal would be extremely handy I believe.
Maybe the sending the goal can return an ID which can then be used to reference the feedback or so.

jacobperron:
Currently, the `send_goal` verb is blocking until the goal completes and echos the sent goal, result, and optional feedback. For example:

```sh
$ ros2 action send_goal --feedback /fibonacci example_interfaces/Fibonacci "order: 5"
Waiting for an action server to become available...
Sending goal:
     order: 5

Feedback:
    sequence: [0, 1, 1]

Feedback:
    sequence: [0, 1, 1, 2]

Feedback:
    sequence: [0, 1, 1, 2, 3]

Feedback:
    sequence: [0, 1, 1, 2, 3, 5]

Result:
    sequence: [0, 1, 1, 2, 3, 5]

Goal finished with status: SUCCEEDED
```

I agree that also echoing the goal ID would be useful.

jacobperron:
This should be ready for review.
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6600/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2934/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5416/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6424/)

Karsten1987:
The command line tool looks great to me. I can send a command, get the feedback and see the action definitions. 👍 

Only nitpick I had was for the `ros2 action info` command. I can't really figure out why we would need to have `-c` as an option. The other thing is that `ros2 action info <topic>` is different from `ros2 action /<topic>`. I believe as with other `ros2cli` tools, the forward slash should be added automatically if suited.

ros2action/package.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="2">
+  <name>ros2action</name>
+  <version>0.6.3</version>
```
Karsten1987: should the version number be reset here?

ros2action/ros2action/verb/__init__.py
```diff
@@ -0,0 +1,44 @@
+# Copyright 2017 Open Source Robotics Foundation, Inc.
```
Karsten1987: year

ros2action/setup.py
```diff
@@ -0,0 +1,43 @@
+from setuptools import find_packages
+from setuptools import setup
+
+setup(
+    name='ros2action',
+    version='0.6.3',
```
Karsten1987: same comment for version number

jacobperron:
> The other thing is that ros2 action info <topic> is different from ros2 action /<topic>. I believe as with other ros2cli tools, the forward slash should be added automatically if suited.

I was trying to mimic `ros2 topic info`, which also has a `-c` option. I'll look into adding the forward slash.

ros2action/package.xml
jacobperron: I'm not sure, but I think for the purpose of tagging releases for all packages in the repo at once it is more convenient to keep the same version across packages. Otherwise, there's the issue of associating Git release tags and the different versions across packages.

ros2action/ros2action/verb/__init__.py
jacobperron: Fixed 4f39c2b

jacobperron:
@Karsten1987 

> ros2 action info <topic> is different from ros2 action /<topic>. I believe as with other ros2cli tools, the forward slash should be added automatically if suited.

I think I've addressed this in 0d13d00.

Also, now printing the goal UUID in hex format (8ef62e0)

Karsten1987:
The forward slash situation was addressed. 👍 

I have two more remarks though:

* ros2 action show: Would be great to have a `-t, --types` option just like `ros2 topic list` to have an idea of the actions and their types. That also makes calling `ros2 action show` easier.
* ros2 action send_cancel: Now that `send_goal` returns a ID for the action, do you think it's feasible to add a `cancel` command which takes this ID and thus sends out a cancel request. I think that could be helpful.

jacobperron:
Thanks @Karsten1987!



> ros2 action show: Would be great to have a -t, --types option just like ros2 topic list to have an idea of the actions and their types. That also makes calling ros2 action show easier.

I plan to make this feature as part of a new `interface` command as described in https://github.com/ros2/ros2cli/issues/202#issuecomment-472540339. Expect a follow up PR with this.

> ros2 action send_cancel: Now that send_goal returns a ID for the action, do you think it's feasible to add a cancel command which takes this ID and thus sends out a cancel request. I think that could be helpful.

This reminded me that the intention was to cancel the goal if the user terminates the `send_goal` command before an action completes (I'll look into adding that to this PR).

I can also see the value in a separate `cancel` verb for interacting with a goal not started by the CLI tool, but I think we can save it for a follow-up ticket. Maybe good to discuss as part of https://github.com/ros2/ros2cli/issues/202.

ros2action/test/test_pep257.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2017 Open Source Robotics Foundation, Inc.
```
sloretz: nit 2019

ros2action/test/test_flake8.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2017 Open Source Robotics Foundation, Inc.
```
sloretz: 2019

ros2action/test/test_copyright.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2017 Open Source Robotics Foundation, Inc.
```
sloretz: 2019

ros2action/ros2action/verb/show.py
```diff
@@ -0,0 +1,38 @@
...
+from ros2action.verb import VerbExtension
+
+
+class ShowVerb(VerbExtension):
+    """Output the action definition."""
+
+    def add_arguments(self, parser, cli_name):
+        arg = parser.add_argument(
+            'action_type',
+            help="Type of the ROS action (e.g. 'examples_interfaces/Fibonacci')")
```
sloretz: `examples_...` -> `example_...`

jacobperron: 5bc307f

ros2action/test/test_copyright.py
jacobperron: 5bc307f

ros2action/test/test_flake8.py
jacobperron: 5bc307f

ros2action/test/test_pep257.py
jacobperron: 5bc307f

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6670/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2989/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5478/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6504/)

ros2action/package.xml
```diff
@@ -0,0 +1,30 @@
...
+  <description>
+    The action command for ROS 2 command line tools.
+  </description>
+  <maintainer email="dthomas@osrfoundation.org">Dirk Thomas</maintainer>
```
dirk-thomas: Please add yourself as the maintainer.

ros2action/ros2action/api/__init__.py
```diff
@@ -0,0 +1,147 @@
...
+        content, _ = get_resource('rosidl_interfaces', package_name)
+    except LookupError:
+        return []
+    interface_names = content.splitlines()
+    # TODO(jacobperron) this logic should come from a rosidl related package
+    # Only return actions in action folder
+    return list(sorted({
+        n[7:-7]
+        for n in interface_names
+        if n.startswith('action/') and n[-7:] in ('.idl', '.action')}))
```
dirk-thomas: How about moving this into `rosidl_runtime_py`?

(Could be done in a follow up PR.)

Same for the similar commented blocks below.

ros2action/ros2action/api/__init__.py
```diff
@@ -0,0 +1,147 @@
...
+    action_servers = []
+
+    try:
+        expanded_name = expand_topic_name(action_name, node.get_name(), node.get_namespace())
+    except ValueError as e:
+        raise RuntimeError(e)
+    try:
+        validate_full_topic_name(expanded_name)
+    except rclpy.exceptions.InvalidTopicNameException as e:
+        raise RuntimeError(e)
```
dirk-thomas: Why raise a generic `RuntimeError` in this API? For the actual CLI I see why it should be a `RuntimeError` (to suppress the stacktrace) but for the API itself it would be more helpful to raise the specific exceptions.

So I would suggest to move the `try` / `except` logic to the verb using the API.

ros2action/package.xml
jacobperron: Done 71de0e5

ros2action/ros2action/api/__init__.py
jacobperron: I think that makes sense. Created a ticket to track this https://github.com/ros2/ros2cli/issues/218

ros2action/ros2action/api/__init__.py
jacobperron: Makes sense. Moved bbd8893

jacobperron:
> I would rather only catch expected exceptions here - basically the two handle before within the function. That way in case of an unexpected exception you still get a valuable stackable.

Done 6a62cb8

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6681/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3000/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5489/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6511/)

PR #224, opened 2019-04-17
set ttl for ros2cli multicast to 10 instead of 1

codebot:
When debugging multicast on non-trivial networks, TTL > 1 is required. I'm not sure of the exact value, but we just had a situation where it needed to be 2, which made debugging challenging :man_facepalming: . Network ops folks suggested we set to 10 to avoid confusion in the future.

Ideally TTL could be a command line option, but for now at least, anything greater than 1 would allow this tool to work for our use case. Just submitting the PR to start the conversation. Cheers

### DIFF
--- a/ros2multicast/ros2multicast/api/__init__.py
+++ b/ros2multicast/ros2multicast/api/__init__.py
@@ -17,10 +17,13 @@ import struct
 
 DEFAULT_GROUP = '225.0.0.1'
 DEFAULT_PORT = 49150
+DEFAULT_TTL = 10
 
 
 def send(data, *, group=DEFAULT_GROUP, port=DEFAULT_PORT):
     s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM, socket.IPPROTO_UDP)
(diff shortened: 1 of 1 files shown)


mjcarroll:
For reference from: https://www.tldp.org/HOWTO/Multicast-HOWTO-2.html
```
TTL     Scope
----------------------------------------------------------------------
   0    Restricted to the same host. Won't be output by any interface.
   1    Restricted to the same subnet. Won't be forwarded by a router.
 <32         Restricted to the same site, organization or department.
 <64 Restricted to the same region.
<128 Restricted to the same continent.
<255 Unrestricted in scope. Global.
```
> Nobody knows what "site" or "region" mean exactly. It is up to the administrators to decide what this limits apply to.

mjcarroll:
We discussed this during the Waffle triage, and the consensus was that changing the default could cause some unexpected side effects.

The most reasonable approach would be to add this as a command line argument as you have already mentioned.

dirk-thomas:
Closing in favor of #282 which implements the suggestion from the comment.

PR #226, opened 2019-04-24
Improve ros2 component CLI

### LINKED ISSUE #225: ros2component verb to start standalone component
#### Feature description

From some discussion around a recent show-and-tell:

Add a `ros2 component` verb to be an equivalent of `nodelet standalone`, where a component container is started with a single component loaded.  It should also support the arguments that `ros2 component load` has (renaming, remapping, etc).

hidmic:
Fixes #225. This pull request adds a `run` verb to start a component on a standalone container, for instance:

```sh
ros2 component run composition composition::Talker -n my_talker
```

But it also:
- Adds `--quiet` flag to existing verbs to ease shell scripting.
- Adds missing dependencies to the `package.xml`
- Mildly refactors the internal package API

### DIFF
--- a/ros2component/package.xml
+++ b/ros2component/package.xml
@@ -9,11 +9,15 @@
   <maintainer email="michel@ekumenlabs.com">Michel Hidalgo</maintainer>
   <license>Apache License 2.0</license>
 
-  <depend>composition_interfaces</depend>
-  <depend>ros2cli</depend>
-  <depend>ros2node</depend>
-  <depend>ros2param</depend>
-  <depend>ros2pkg</depend>
+  <exec_depend>ament_index_python</exec_depend>
+  <exec_depend>composition_interfaces</exec_depend>
+  <exec_depend>rcl_interfaces</exec_depend>
+  <exec_depend>rclcpp_components</exec_depend>
+  <exec_depend>rclpy</exec_depend>
+  <exec_depend>ros2cli</exec_depend>
+  <exec_depend>ros2node</exec_depend>
+  <exec_depend>ros2param</exec_depend>
+  <exec_depend>ros2pkg</exec_depend>
 
   <test_depend>ament_copyright</test_depend>
   <test_depend>ament_flake8</test_depend>
--- a/ros2component/ros2component/api/__init__.py
+++ b/ros2component/ros2component/api/__init__.py
@@ -256,19 +256,19 @@ def add_component_arguments(parser):
         'plugin_name', help='Type name of the component to be loaded'
     )
     argument.completer = ComponentTypeNameCompleter(package_name_key='package_name')
-    parser.add_argument('-n', '--node-name', default=None, help='Component node name')
-    parser.add_argument('--node-namespace', default=None, help='Component node namespace')
-    parser.add_argument('--log-level', default=None, help='Component node log level')
+    parser.add_argument('-n', '--node-name', help='Component node name')
+    parser.add_argument('--node-namespace', help='Component node namespace')
+    parser.add_argument('--log-level', help='Component node log level')
     parser.add_argument(
-        '-r', '--remap-rule', action='append', default=None, dest='remap_rules',
+        '-r', '--remap-rule', action='append', dest='remap_rules',
         help="Component node remapping rules, in the 'from:=to' form"
     )
     parser.add_argument(
-        '-p', '--parameter', action='append', default=None, dest='parameters',
+        '-p', '--parameter', action='append', dest='parameters',
         help="Component node parameters, in the 'name:=value' form"
     )
     parser.add_argument(
-        '-e', '--extra-argument', action='append', default=None, dest='extra_arguments',
+        '-e', '--extra-argument', action='append', dest='extra_arguments',
         help="Extra arguments for the container, in the 'name:=value' form"
     )
 
--- a/ros2component/ros2component/api/__init__.py
+++ b/ros2component/ros2component/api/__init__.py
@@ -271,3 +271,17 @@ def add_component_arguments(parser):
         '-e', '--extra-argument', action='append', default=None, dest='extra_arguments',
         help="Extra arguments for the container, in the 'name:=value' form"
     )
+
+
+def run_standalone_container(*, container_node_name):
+    """Run a standalone component container."""
+    try:
+        paths = get_executable_paths(package_name='rclcpp_components')
+    except PackageNotFound:
+        raise RuntimeError("Package 'rclcpp_components' not found")
+
+    executable_path = next((p for p in paths if 'component_container' in p), None)
+    if executable_path is None:
+        raise RuntimeError('No component container node found!')
+
+    return subprocess.Popen([executable_path, '__node:=' + container_node_name])
--- a/ros2component/ros2component/api/__init__.py
+++ b/ros2component/ros2component/api/__init__.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 from collections import namedtuple
+import subprocess
 
 from ament_index_python import get_resource
 from ament_index_python import get_resources
@@ -28,7 +29,8 @@ from ros2cli.node.strategy import NodeStrategy
 from ros2node.api import get_node_names
 from ros2node.api import get_service_info
 from ros2param.api import get_parameter_value
-
+from ros2pkg.api import get_executable_paths
+from ros2pkg.api import PackageNotFound
 
 COMPONENTS_RESOURCE_TYPE = 'rclcpp_components'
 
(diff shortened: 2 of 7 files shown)


hidmic:
@mjcarroll incidentally, I think I found a bug in the composable nodes implementation. If you execute:

```sh
ros2 component run composition composition::Talker
```

The composed node is given the same name as that of the container, which is being passed upon container executation via `__node:=standalone_container_<unique 12 chars long hex>`. It seems that upon loading, container remapping rules like `__node` affect the components as well.

hidmic:
CI (build up to and test only `ros2component`):

* Linux [Build Status](https://ci.ros2.org/job/ci_linux/6777/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3078/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5573/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6605/)

ros2component/package.xml
```diff
@@ -9,16 +9,22 @@
   <maintainer email="michel@ekumenlabs.com">Michel Hidalgo</maintainer>
   <license>Apache License 2.0</license>
 
+  <depend>ament_index_python</depend>
```
dirk-thomas: This shouldn't be needed a build time?

Same below - also for various dependencies not touched in this PR.

ros2component/ros2component/verb/list.py
```diff
@@ -59,7 +52,7 @@ def main(self, *, args):
                         node=node, remote_container_node_name=args.container_node_name
                     )
                     if any(components):
-                        print(*['{}  {}'.format(c.uid, c.name) for c in components], sep='\n')
+                        print(*['{}    {}'.format(c.uid, c.name) for c in components], sep='\n')
```
dirk-thomas: What motivates the change in indentation?

ros2component/ros2component/verb/run.py
```diff
@@ -0,0 +1,57 @@
...
+                parameters=args.parameters, extra_arguments=args.extra_arguments
+            )
+
+        while container.returncode is None:
+            try:
+                container.communicate()
+            except KeyboardInterrupt:
+                # the subprocess will also receive the signal and should shut down
+                # therefore we continue here until the process has finished
+                pass
```
dirk-thomas: Doesn't this still need to wait for the subprocess to actually be finished before trying to retrieve the return code?

ros2component/package.xml
hidmic: Correct, I replicated below's bad practices. Fixed

ros2component/ros2component/verb/list.py
hidmic: To get closer to a conventionally 4 spaces wide tabulation, but otherwise as arbitrary as the former 2 spaces. 

My intent is to be consistent about indentation while at the same time making it easy to use for someone that's shell scripting. Any thoughts on usability in that regard are welcome

ros2component/ros2component/verb/run.py
hidmic: That's what we're doing with loop and [`Popen.communicate()`](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate). I actually took this from the [`ros2 run` CLI](https://github.com/ros2/ros2cli/blob/master/ros2run/ros2run/api/__init__.py#L62). Are you worried about a potential exception that's not `KeyboardInterrupt`?

hidmic:
> It would be much easier to read if these unrelated changes would be separate PRs (or at least separate commits). I don't see a reason why they should be bundled in a single commit?

No technical reason, just to avoid having to git plumb it. 

> To distinguish run more clearly from load how about naming it standalone instead?

So I subjectively prefer `ros2 component run` over `ros2 component standalone` (thus why I didn't do it that way initially) but it's not a strong preference. Changed.

ros2component/ros2component/verb/run.py
dirk-thomas: I think I just missed that it is within a loop.

ros2component/ros2component/verb/list.py
dirk-thomas: Since other verbs use 2 spaces for indentation I would rather keep it with 2 for now.

hidmic:
@mjcarroll @dirk-thomas mind to review?

ros2component/ros2component/api/__init__.py
```diff
@@ -182,13 +184,7 @@ def unload_component_from_container(*, node, remote_container_node_name, compone
...
             rclpy.spin_until_future_complete(node, future)
             response = future.result()
-            if not response.success:
-                return 'Failed to unload component {} from {} container\n  {}'.format(
-                    uid, remote_container_node_name, response.error_message
-                )
-            print('Unloaded component {} from {} container'.format(
-                uid, remote_container_node_name
-            ))
+            yield uid, not response.success, response.error_message
```
dirk-thomas: Why `yield` here? I don't see any `async` anywhere around for this function.

ros2component/ros2component/api/__init__.py
```diff
@@ -248,3 +244,44 @@ def __init__(self, *, package_name_key=None):
...
+    """Add component specific arguments to the CLI parser."""
+    argument = parser.add_argument(
+        'package_name', help='Name of the package where the component is to be found'
+    )
+    argument.completer = package_with_components_name_completer
+    argument = parser.add_argument(
+        'plugin_name', help='Type name of the component to be loaded'
+    )
+    argument.completer = ComponentTypeNameCompleter(package_name_key='package_name')
+    parser.add_argument('-n', '--node-name', default=None, help='Component node name')
```
dirk-thomas: Nitpick: `default=None` doesn't have to be specified, it is the default already.

Same below.

ros2component/ros2component/verb/list.py
dirk-thomas: Still pending.

ros2component/ros2component/verb/list.py
```diff
@@ -69,8 +62,5 @@ def main(self, *, args):
                         )
                         if any(components):
                             print(*[
-                                2 * '  ' + '{}  {}'.format(c.uid, c.name) for c in components
+                                4 * ' ' + '{}    {}'.format(c.uid, c.name) for c in components
```
dirk-thomas: Same here.

ros2component/ros2component/verb/load.py
```diff
@@ -35,51 +31,32 @@ class LoadVerb(VerbExtension):
...
-        )
-        parser.add_argument(
-            '-p', '--parameter', action='append', default=None, dest='parameters',
-            help="Component node parameters, in the 'name:=value' form"
-        )
-        parser.add_argument(
-            '-e', '--extra-argument', action='append', default=None, dest='extra_arguments',
-            help="Extra arguments for the container, in the 'name:=value' form"
+            '-q', '--quiet', action='store_true', default=False,
+            help='Print bare minimum output: component unique IDs and names only'
```
dirk-thomas: Nitpick: I would drop the phrase "bare minimum output" and just state what it does:

> Only print component unique IDs and names

Same below.

ros2component/ros2component/verb/load.py
```diff
@@ -35,51 +31,32 @@ class LoadVerb(VerbExtension):
...
+                node_name=args.node_name, node_namespace=args.node_namespace,
+                log_level=args.log_level, remap_rules=args.remap_rules,
+                parameters=args.parameters, extra_arguments=args.extra_arguments
+            )
+            if not args.quiet:
+                print("Loaded component {} into '{}' container node as '{}'".format(
+                    component_uid, args.container_node_name, component_name
+                ))
+            else:
+                print('{}    {}'.format(component_uid, component_name))
```
dirk-thomas: Same here.

hidmic: Done.

ros2component/ros2component/verb/load.py
hidmic: Sounds good.

ros2component/ros2component/verb/list.py
hidmic: Changed.

ros2component/ros2component/api/__init__.py
hidmic: Alright, too explicit maybe.

ros2component/ros2component/api/__init__.py
hidmic: Because it's a generator.

ros2component/ros2component/verb/list.py
hidmic: Alright, fair enough.

ros2component/ros2component/api/__init__.py
dirk-thomas: Oh, I see it in the `for` loop now. :+1:

PR #232, opened 2019-05-02
add xmllint linter test

mikaelarguedas:
Similar to https://github.com/ros2/rosidl_python/pull/53.

Maybe there is a better way of doing this? is there something like `ament_lint_common` for python packages?


dirk-thomas:
Thanks for the improvement.

> is there something like ament_lint_common for python packages?

Not at the moment but there could be.

We will likely wait with running CI until after todays API freeze.

dirk-thomas:
Thanks for the patch.

[Build Status](https://ci.ros2.org/job/ci_linux/6931/)

Issue #156, opened 2018-11-07
show QoS setting of pubs / subs, option to choose transient local


dirk-thomas:
Since transient local doesn't get matched with non-transient local the command line tools might only work with some topics. It would be good if these QoS settings would be visible and an options added to use transient local in the tool in order to match with other transient local entities.

stonier:
If possible, even better if, e.g. `rostopic echo` is able to detect the type of publisher and the subscriber configures itself to be compatible.

dirk-thomas:
> If possible, even better if, e.g. `rostopic echo` is able to detect the type of publisher and the subscriber configures itself to be compatible.

This might work if all publishers use the same QoS but if they are heterogeneous an option to choose is required.

stonier:
Assuming the `ros2 topic echo` node can introspect on the QoS settings of the publisher, wouldn't it be just a matter of matching durability and reliability settings (as per tables [here](https://index.ros.org/doc/ros2/Concepts/About-Quality-of-Service-Settings/#qos-compatibilities))?

dirk-thomas:
> wouldn't it be just a matter of matching durability and reliability settings

Are you implying to create multiple subscribers? If yes, how do you make sure to not receive duplicates?

stonier:
> Are you implying to create multiple subscribers?

1. Introspect the QoS settings of the publisher (if it's possible)
2. Configure QoS settings on the subscriber that are compatible

dirk-thomas:
How about the scenario I mentioned above: to publishers with different QoS? There might be no QoS setting for a single subscriber which allows to receive messages from both.

stonier:
Aye, sorry - makes sense. I was getting distracted fumbling in the dark with problems on pubs/subs that have different QoS. 

So, learning that concept of a topic really is a more complex beast. Given that it can mask multiple, independent flows of data, both `ros2 topic list` and `rqt_graph` no longer tell the complete story about how data is moving around a runtime system. That complexity seems like a shame, but given the non-centralised and asynchronous connection handling, I can't see how you can eat that cake and get all the QoS goodies at the same time. Would be curious to learn what thoughts you had on this.

Nonetheless, it would still be easy to manually orchestrate a robot launch to separate topics for incompatible connections, e.g. `/lidar/cloud/reliable` and `/lidar/cloud_unreliable` and you'll get that simple architecture for free, but development/debugging tools on the command line would still be eminently useful.

Example use case that would have been useful for me recently (on topics):

[...]

PR #238, opened 2019-05-16
[ros2topic] Update pub to use qos command line settings.

mjcarroll:
Add functionality to `ros2 topic pub` that will allow for QoS parameters to be passed in on the command line.

Partially addresses #156 

Signed-off-by: Michael Carroll <michael@openrobotics.org>

### DIFF
--- a/ros2topic/ros2topic/verb/pub.py
+++ b/ros2topic/ros2topic/verb/pub.py
@@ -56,19 +56,19 @@ class PubVerb(VerbExtension):
             '-1', '--once', action='store_true',
             help='Publish one message and exit')
         parser.add_argument(
-            '-n', '--node-name', type=str,
+            '-n', '--node-name',
             help='Name of the created publishing node')
         parser.add_argument(
-            '--qos-profile', type=str,
+            '--qos-profile',
             choices=QOS_PROFILE_OPT.keys(),
             help='Quality of service profile to publish with')
         parser.add_argument(
-            '--qos-reliability', type=str,
+            '--qos-reliability',
             choices=QOS_RELIABILITY_OPT.keys(),
             help='Quality of service reliability setting to publish with. ' +
                  '(Will override reliability value of --qos-profile option)')
         parser.add_argument(
-            '--qos-durability', type=str,
+            '--qos-durability',
             choices=QOS_DURABILITY_OPT.keys(),
             help='Quality of service durability setting to publish with. ' +
                  '(Will override durability value of --qos-profile option)')
@@ -101,8 +101,6 @@ def publisher(
 
     # Build a QoS profile based on user-supplied arguments
     profile = rclpy.qos.qos_profile_system_default
-    if qos_profile:
-        profile = QOS_PROFILE_OPT[qos_profile]
     if qos_durability:
         profile.durability = QOS_DURABILITY_OPT[qos_durability]
     if qos_reliability:
--- a/ros2topic/ros2topic/verb/pub.py
+++ b/ros2topic/ros2topic/verb/pub.py
@@ -43,8 +43,8 @@ class PubVerb(VerbExtension):
             topic_name_key='topic_name')
         parser.add_argument(
             'values', nargs='?', default='{}',
-            help='Values to fill the message with in YAML format ' +
-                 '(e.g. "data: Hello World"), ' +
+            help='Values to fill the message with in YAML format '
+                 '(e.g. "data: Hello World"), '
                  'otherwise the message will be published with default values')
         parser.add_argument(
             '-r', '--rate', metavar='N', type=float, default=1.0,
@@ -65,12 +65,12 @@ class PubVerb(VerbExtension):
         parser.add_argument(
             '--qos-reliability',
             choices=QOS_RELIABILITY_OPT.keys(),
-            help='Quality of service reliability setting to publish with. ' +
+            help='Quality of service reliability setting to publish with. '
                  '(Will override reliability value of --qos-profile option)')
         parser.add_argument(
             '--qos-durability',
             choices=QOS_DURABILITY_OPT.keys(),
-            help='Quality of service durability setting to publish with. ' +
+            help='Quality of service durability setting to publish with. '
                  '(Will override durability value of --qos-profile option)')
 
     def main(self, *, args):
@@ -108,7 +108,7 @@ def publisher(
 
     node = rclpy.create_node(node_name)
 
-    pub = node.create_publisher(msg_module, topic_name, qos_profile=profile)
+    pub = node.create_publisher(msg_module, topic_name, profile)
 
     msg = msg_module()
     try:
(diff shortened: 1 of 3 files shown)


mjcarroll:
Currently, I'm verifying with the `qos_subscription.py` file included in this PR.  When that is executed, various configurations can be checked against the [QoS Compatibility Matrix](https://index.ros.org/doc/ros2/Concepts/About-Quality-of-Service-Settings/#qos-compatibilities)

For example, reliable publication is received on all subscriptions:
```
ros2 topic pub /topic std_msgs/String 'data: Hello World' --qos-reliability=reliable -1
{'default': 0, 'transient': 0, 'volatile': 0, 'reliable': 0, 'best_effort': 0}
{'default': 1, 'transient': 1, 'volatile': 1, 'reliable': 1, 'best_effort': 1}
```

Best effort publication won't be received on reliable subscriptions (default is reliable, in this case).
```
ros2 topic pub /topic std_msgs/String 'data: Hello World' --qos-reliability=best_effort -1
# Results in:
{'default': 0, 'transient': 0, 'volatile': 0, 'reliable': 0, 'best_effort': 0}
{'default': 0, 'transient': 1, 'volatile': 1, 'reliable': 0, 'best_effort': 1}
```

ros2topic/ros2topic/verb/pub.py
```diff
@@ -56,6 +58,20 @@ def add_arguments(self, parser, cli_name):
         parser.add_argument(
             '-n', '--node-name', type=str,
             help='Name of the created publishing node')
+        parser.add_argument(
+            '--qos-profile', type=str,
```
dirk-thomas: `type=str` is the default afaik should could be removed.

Same below.

ros2topic/ros2topic/verb/pub.py
```diff
@@ -56,6 +58,20 @@ def add_arguments(self, parser, cli_name):
...
+            choices=QOS_PROFILE_OPT.keys(),
+            help='Quality of service profile to publish with')
```
dirk-thomas: Just an idea (doesn't have be done): it would be neat if the help message would print the details about every profile so the users sees what they stand for :bulb:

ros2topic/ros2topic/verb/pub.py
```diff
@@ -81,9 +99,18 @@ def publisher(
         node_name = NODE_NAME_PREFIX + '_publisher_%s' % (message_type.replace('/', '_'), )
     rclpy.init()
 
+    # Build a QoS profile based on user-supplied arguments
+    profile = rclpy.qos.qos_profile_system_default
+    if qos_profile:
+        profile = QOS_PROFILE_OPT[qos_profile]
```
dirk-thomas: If you map `QOS_PROFILE_OPT[None]` to `rclpy.qos.qos_profile_system_default` then you could use just `profile = QOS_PROFILE_OPT` here without the conditional.

ros2topic/test/qos_subscription.py
```diff
@@ -0,0 +1,95 @@
...
+            'topic',
+            lambda msg: counter.onMsg('best_effort', msg),
+            qos_profile=QoSProfile(
+                reliability=QoSReliabilityPolicy.RMW_QOS_POLICY_RELIABILITY_BEST_EFFORT))
+
+    assert sub_default
+    assert sub_transient
+    assert sub_volatile
+    assert sub_reliable
+    assert sub_best_effort
```
dirk-thomas: Neither of the above API can return `None`, correct? Then these checks seem unnecessary.

mjcarroll: I don't believe so, but this is consistent with other places where we have done this to suppress the unused variable warnings.

dirk-thomas: Oh, I see. That is the main purpose of these lines. A comment would have clarified :wink:

mjcarroll: Comment added.

ros2topic/ros2topic/verb/pub.py
```diff
@@ -54,8 +56,22 @@ def add_arguments(self, parser, cli_name):
...
+            '-n', '--node-name',
             help='Name of the created publishing node')
+        parser.add_argument(
+            '--qos-profile',
+            choices=QOS_PROFILE_OPT.keys(),
+            help='Quality of service profile to publish with')
+        parser.add_argument(
+            '--qos-reliability',
+            choices=QOS_RELIABILITY_OPT.keys(),
+            help='Quality of service reliability setting to publish with. ' +
```
dirk-thomas: Nitpick: the trailing ` +` is not necessary.

Same below.

ros2topic/test/qos_subscription.py
```diff
@@ -0,0 +1,96 @@
...
+            'topic',
+            lambda msg: counter.onMsg('reliable', msg),
+            qos_profile=QoSProfile(
+                reliability=QoSReliabilityPolicy.RMW_QOS_POLICY_RELIABILITY_RELIABLE))
+
+    sub_best_effort = node.create_subscription(
+            String,
+            'topic',
+            lambda msg: counter.onMsg('best_effort', msg),
+            qos_profile=QoSProfile(
```
jacobperron: This is deprecated API. Instead, drop the keyword:
```suggestion
            QoSProfile(
```

jacobperron: Also, constructing `QoSProfile()` without a `depth` keyword argument is deprecated.

ros2topic/ros2topic/verb/pub.py
```diff
@@ -81,9 +99,16 @@ def publisher(
...
+    profile = rclpy.qos.qos_profile_system_default
+    if qos_durability:
+        profile.durability = QOS_DURABILITY_OPT[qos_durability]
+    if qos_reliability:
+        profile.reliability = QOS_RELIABILITY_OPT[qos_reliability]
+
     node = rclpy.create_node(node_name)
 
-    pub = node.create_publisher(msg_module, topic_name, 10)
+    pub = node.create_publisher(msg_module, topic_name, qos_profile=profile)
```
jacobperron: This is deprecated. Prefer:
```suggestion
    pub = node.create_publisher(msg_module, topic_name, profile)
```

ros2topic/test/qos_subscription.py
```diff
@@ -0,0 +1,96 @@
...
+    rclpy.init(args=args)
+    node = rclpy.create_node('qos_subscription')
+
+    counter = Counter()
+
+    sub_default = node.create_subscription(
+            String,
+            'topic',
+            lambda msg: counter.onMsg('default', msg),
+            qos_profile=rclpy.qos.qos_profile_default)
```
jacobperron: `qos_profile_default` is deprecated. You can replace this with `QoSProfile(depth=10)`, which I think is equivalent.

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7126/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3342/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5836/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6947/)

PR #242, opened 2019-05-17
use new type identification for service calls

Karsten1987:
Similar problem as described here: https://github.com/ros2/ros2cli/pull/241
`ros2 service call ...` fails because the new service types are parsed incorrectly.

Signed-off-by: Karsten Knese <karsten@openrobotics.org>

### DIFF
--- a/ros2service/ros2service/verb/call.py
+++ b/ros2service/ros2service/verb/call.py
@@ -59,7 +59,7 @@ class CallVerb(VerbExtension):
 def requester(service_type, service_name, values, period):
     # TODO(wjwwood) this logic should come from a rosidl related package
     try:
-        package_name, srv_name = service_type.split('/', 2)
+        package_name, _, srv_name = service_type.split('/', 3)
         if not package_name or not srv_name:
             raise ValueError()
     except ValueError:
--- a/ros2service/ros2service/verb/call.py
+++ b/ros2service/ros2service/verb/call.py
@@ -58,17 +58,15 @@ class CallVerb(VerbExtension):
 
 def requester(service_type, service_name, values, period):
     # TODO(wjwwood) this logic should come from a rosidl related package
+    # TODO(karsten1987) as of dashing, types are split in three parts indicating
+    # package_Name, middle_module (e.g. srv, msg, action), srv_name
+    # This might change in the future and has to be re-addressed if so.
     try:
-        package_name, _, srv_name = service_type.split('/', 3)
+        package_name, middle_module, srv_name = service_type.split('/', 3)
         if not package_name or not srv_name:
             raise ValueError()
     except ValueError:
         raise RuntimeError('The passed service type is invalid')
-
-    # TODO(sloretz) node API to get topic types should indicate if action or srv
-    middle_module = 'srv'
-    if service_name.endswith('/_action/get_result') or service_name.endswith('/_action/send_goal'):
-        middle_module = 'action'
 
     module = importlib.import_module(package_name + '.' + middle_module)
     srv_module = getattr(module, srv_name)
--- a/ros2service/ros2service/verb/call.py
+++ b/ros2service/ros2service/verb/call.py
@@ -59,7 +59,7 @@ class CallVerb(VerbExtension):
 def requester(service_type, service_name, values, period):
     # TODO(wjwwood) this logic should come from a rosidl related package
     # TODO(karsten1987) as of dashing, types are split in three parts indicating
-    # package_Name, middle_module (e.g. srv, msg, action), srv_name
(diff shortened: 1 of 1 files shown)


ros2service/ros2service/verb/call.py
```diff
@@ -59,7 +59,7 @@ def main(self, *, args):
 def requester(service_type, service_name, values, period):
     # TODO(wjwwood) this logic should come from a rosidl related package
     try:
-        package_name, srv_name = service_type.split('/', 2)
+        package_name, _, srv_name = service_type.split('/', 3)
```
dirk-thomas: Instead of ignoring the second value this could consolidate the `middle_module` below?

Karsten1987: with that new logic, I believe we don't need that `middle_module` logic any longer. @sloretz do you think this could address the TODO here?

As far as I understand, this new type identification serves exactly the case of separating actions from services, right? So storing the second value of the split call as the middle module should suffice.

sloretz: > with that new logic, I believe we don't need that middle_module logic below any longer. @sloretz do you think this could address the TODO here?

I think that would resolve the TODO on line 68. IIUC in the future it might be possible for there to be more than 3 parts since it's coming from a scoped name in IDL, so you might add a TODO for that

Karsten1987: done in https://github.com/ros2/ros2cli/pull/242/commits/65a3a218e3a148a4a4714fc4ade7c5acf3ea3f9c

dirk-thomas:
Why not just (without error checking)?

```
parts = service_type.split('/')
module = importlib.import_module('.'.join(parts[:-1]))
srv_module = getattr(module, parts[-1])
```

Karsten1987:
I don't have enough insights about this to answer whether or not the error checking makes sense or not. I'd naively say yes (error checking makes sense) so that we catch easier when this behavior changes in the future, but I am open for whatever is best practice here.

jacobperron:
I think more generally having what @dirk-thomas suggests makes sense, since we haven't settle on how many parts to the message namespace there can be. Probably a better way to error check is to wrap the whole thing in a try-except since the import_module / getattr should tell us if the name is valid, right?

Karsten1987:
@dirk-thomas is the proposed way what you envisioned? I am happy to change it to whatever is recommended.

dirk-thomas:
My snippet was meant as this-is-how-the-logic-should-work and would get rid of all the extra steps currently being used. The snippet doesn't do any error checking - but the actual patch should certainly do that - I just left that out for brevity.

Karsten1987:
@jacobperron @dirk-thomas does https://github.com/ros2/ros2cli/pull/242/commits/468289d0745833b860c766ccf3f03da028d1f7d7 look like what you've suggested?

Karsten1987:
Did it work before this patch with actions? I haven't tested it with actions. Do we have a working example for this?

dirk-thomas:
I assume it worked before. Since you are changing the code dealing with action types in this patch please also check it with an action type. I don't have an example at hand.

Karsten1987:
How do we get the type information of the actual action service?
Let's say I run:
```
ros2 run examples_rclcpp_minimal_action_server action_server_member_functions
```
what's the respective type to the service:
```
ros2 service call /fibonacci/_action/send_goal <type> "{order: 10}"
```
I've tried:
```
ros2 service call /fibonacci/_action/send_goal example_interfaces/action/Fibonacci_SendGoal_Request "{order: 10}"
```
but this fails with `AttributeError: module 'example_interfaces.action' has no attribute 'Fibonacci_SendGoal_Request'`. This makes me believe that the split logic works correctly, but I've specified the wrong type.

jacobperron:
The "send goal" service is defined in the module `example_interfaces.action._fibonacci`. Even with the code before, I don't think it is possible to make the service call work for actions. But maybe now with multiple namespaces allowed, it could be possible with something like:

```
ros2 service call /fibonacci/_action/send_goal example_interfaces/action/_fibonacci/Fibonacci_SendGoal "{goal: {order: 10}}"
```

Karsten1987:
Great! @jacobperron example does work.

Karsten1987:
CI (build: `--packages-up-to ros2service` | test: `--packages-select ros2service`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7175/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3386/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5876/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7006/)

Issue #244, opened 2019-05-21
ros2 topic pub fails


Karsten1987:
## Bug report

**Required Info:**

- Operating System:
  - OSX
- Installation type:
  - source
- Version or commit hash:
  - HEAD

#### Steps to reproduce issue

```
ros2 run demo_nodes_cpp listener
```
```
ros2 topic pub /chatter std_msgs/msg/String "{data: Hello}"
```

#### Expected behavior
publishing `Hello`

#### Actual behavior
```
 ➭ ros2 topic pub /chatter std_msgs/msg/String "{data: Hello}"
Traceback (most recent call last):
  File "/Users/karsten/workspace/osrf/ros2_full/install/bin/ros2", line 11, in <module>
    load_entry_point('ros2cli', 'console_scripts', 'ros2')()
  File "/Users/karsten/workspace/osrf/ros2_full/build/ros2cli/ros2cli/cli.py", line 69, in main
    rc = extension.main(parser=parser, args=args)
  File "/Users/karsten/workspace/osrf/ros2_full/build/ros2topic/ros2topic/command/topic.py", line 43, in main
    return extension.main(args=args)
  File "/Users/karsten/workspace/osrf/ros2_full/build/ros2topic/ros2topic/verb/pub.py", line 77, in main
    return main(args)
  File "/Users/karsten/workspace/osrf/ros2_full/build/ros2topic/ros2topic/verb/pub.py", line 84, in main
    args.qos_profile, args.qos_reliability, args.qos_durability)
  File "/Users/karsten/workspace/osrf/ros2_full/build/ros2topic/ros2topic/verb/pub.py", line 100, in publisher
    profile = rclpy.qos.QoSPresetProfiles.get_from_short_key(qos_profile)
  File "/Users/karsten/workspace/osrf/ros2_full/install/lib/python3.7/site-packages/rclpy/qos.py", line 365, in get_from_short_key
    return cls[name.upper()].value
AttributeError: 'NoneType' object has no attribute 'upper'
```

Not sure if this is already addressed somewhere

PR #240, opened 2019-05-17
Use rclpy qos name translations instead of defining here

emersonknapp:
Depends on ros2/rclpy#352

Signed-off-by: Emerson Knapp <eknapp@amazon.com>

PR #245, opened 2019-05-21
use system_default as qos for ros2 topic pub

Karsten1987:
fixes https://github.com/ros2/ros2cli/issues/244

I am not sure whether this is the right default value, but given the name, it sounds sensible to me.
If the idea is to necessarily expose a QoS profile, I think we should change the command line to force a qos profile and not being optional.

Signed-off-by: Karsten Knese <karsten@openrobotics.org>


jacobperron:
Looks like the bug was introduced in https://github.com/ros2/ros2cli/pull/240. Previously, if no QoS profile was provided "system default" was used.

We ultimately want application developers to provide a history / depth setting explicitly, so we might consider requiring these settings in particular for the CLI tools. But I don't think it is necessary to require the entire QoS profile.

PR #255, opened 2019-05-23
Better error message for ros2 action show

jacobperron:
Catch a potential error when splitting the input string.
This is how it is done for 'ros2 msg show' and 'ros2 srv show'.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7208/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3409/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5905/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7035/)

### DIFF
--- a/ros2action/ros2action/verb/show.py
+++ b/ros2action/ros2action/verb/show.py
@@ -27,8 +27,12 @@ class ShowVerb(VerbExtension):
         arg.completer = action_type_completer
 
     def main(self, *, args):
-        package_name, action_name = args.action_type.split('/', 2)
-        if not package_name or not action_name:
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Shouldn't this operate on 3-part names?

jacobperron:
Right now, none of the `show` verbs work on 3-part names. Though, I think it's probably good to update all to handle the old style and the new 3-part style.

jacobperron:
I *think* because things related to the ament resource index haven't been updated, we have to ignore the middle part for showing purposes.

PR #258, opened 2019-05-23
abort pkg create if destination directory exists

### LINKED ISSUE #256: ros2 pkg create can destroy work
## Bug report

I've been testing out `ros2 pkg create`, and it seems like it can destroy work if the same command is run twice.  For instance, if I run `ros2 pkg create foo`, then edit `foo/package.xml`, and then run `ros2 pkg create foo`, it will destroy any edits I already made to `foo/package.xml`.  This is obviously user error, but it seems too easy to destroy existing code.

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - Source
- Version or commit hash:
  - d98854ebad7aeb32a3124b9016e50ab904555c8e
- DDS implementation:
  - Fast-RTPS
  - rclcpp

#### Steps to reproduce issue

```
ros2 pkg create foo
cat foo/package.xml # notice that version is 0.0.0
sed -i -e 's/0.0.0/1.0.0/' foo/package.xml
ros2 pkg create foo
cat foo/package.xml # notice that version is back to 0.0.0
```

#### Expected behavior

`ros2 pkg create foo` throws an error if the package already exists in the current directory.

#### Actual behavior

`ros2 pkg create foo` overwrites files in the current directory.

#### Additional information

This is clearly user error, but it is also surprising.  Also, this has been the case for a while, so this isn't a new issue.

dirk-thomas:
Fixes #256.

### DIFF
--- a/ros2pkg/ros2pkg/verb/create.py
+++ b/ros2pkg/ros2pkg/verb/create.py
@@ -143,6 +143,11 @@ class CreateVerb(VerbExtension):
         if args.cpp_library_name:
             print('cpp_library_name:', args.cpp_library_name)
 
+        package_path = os.path.join(args.destination_directory, package.name)
+        if os.path.exists(package_path):
+            return '\nAborted!\nThe directory already exists: ' + package_path + '\nEither ' + \
+                'remove the directory or choose a different destination directory or package name'
+
         package_directory, source_directory, include_directory = \
             create_package_environment(package, args.destination_directory)
         if not package_directory:
--- a/ros2pkg/ros2pkg/verb/create.py
+++ b/ros2pkg/ros2pkg/verb/create.py
@@ -128,6 +128,11 @@ class CreateVerb(VerbExtension):
             exports=[Export('build_type', content=args.build_type)]
         )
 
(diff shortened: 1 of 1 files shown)


ros2pkg/ros2pkg/verb/create.py
```diff
@@ -143,6 +143,11 @@ def main(self, *, args):
         if args.cpp_library_name:
             print('cpp_library_name:', args.cpp_library_name)
 
+        package_path = os.path.join(args.destination_directory, package.name)
```
clalancette: This is definitely better.  However, I think it would be better to have this error occur before the `print`s above, just so the user doesn't get confused by some output followed by an error.

dirk-thomas: Done: 48a1941cea62d5b55c255a3de809b96a2d03e53b

dirk-thomas:
@clalancette I will go ahead and merge this without your review in order to include this fix in the nightly builds. Please review the PR anyway even after it was merged.

clalancette:
I can't hit "Approve" on already merged PRs, so I'll just say "Looks good to me".

PR #259, opened 2019-05-23
use three-part interface names in msg/srv/action show and msg/srv/ list

### LINKED ISSUE #257: ros2 srv show doesn't support <pkg>/srv/<servicename>
## Bug report

The output from `ros2 service list -t` shows something like:

```
/add_two_ints [example_interfaces/srv/AddTwoInts]
```

However, if you try to show that service type with `ros2 srv show example_interfaces/srv/AddTwoInts`, you get an error

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - Source
- Version or commit hash:
  - eb5d2b79fabbd3f4cf888e0302f7841589acb79b
- DDS implementation:
  - Fast-RTPS
  - rclpy

#### Steps to reproduce issue

In terminal 1:
```
ros2 run demo_nodes_cpp add_two_ints_server
```

In terminal 2:
```
ros2 service list -t  # Among other things, shows /add_two_ints [example_interfaces/srv/AddTwoInts]
ros2 srv show example_interfaces/srv/AddTwoInts
```

#### Expected behavior
Outputs:
```
int64 a
int64 b
---
int64 sum
```

#### Actual behavior
Outputs:
```
The passed service type is invalid
```

#### Additional information

Running `ros2 srv show example_interfaces/AddTwoInts` does the correct thing.

dirk-thomas:
Fixes #255 and fixes #257.

### DIFF
--- a/ros2msg/ros2msg/api/__init__.py
+++ b/ros2msg/ros2msg/api/__init__.py
@@ -65,7 +65,7 @@ def message_type_completer(**kwargs):
     for package_name, message_names in get_all_message_types().items():
         for message_name in message_names:
             message_types.append(
-                '{package_name}/{message_name}'.format_map(locals()))
+                '{package_name}/msg/{message_name}'.format_map(locals()))
     return message_types
 
 
--- a/ros2action/ros2action/api/__init__.py
+++ b/ros2action/ros2action/api/__init__.py
@@ -118,7 +118,7 @@ def action_type_completer(**kwargs):
     for package_name, action_names in get_all_action_types().items():
         for action_name in action_names:
             action_types.append(
-                '{package_name}/{action_name}'.format_map(locals()))
+                '{package_name}/action/{action_name}'.format_map(locals()))
     return action_types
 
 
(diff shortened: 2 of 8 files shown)


ros2msg/ros2msg/api/__init__.py
```diff
@@ -65,7 +65,7 @@ def message_type_completer(**kwargs):
     for package_name, message_names in get_all_message_types().items():
         for message_name in message_names:
             message_types.append(
-                '{package_name}/{message_name}'.format_map(locals()))
+                '{package_name}/msg/{message_name}'.format_map(locals()))
```
jacobperron: Maybe it's worth adding a comment that we're assuming all messages are in the subnamespace `msg`, which won't necessarily be true in the future. E.g. if we want to also show messages defined in `srv`, `action`, or an arbitrary namespace defined in an idl file.

Same for services below.

dirk-thomas: I guess that remark applies to the majority of the message generation, resource index and usage pipeline. So I would rather not bother to add a comment in this specific case - especially since the explicit string literal states just that (imo).

jacobperron: Fair enough.

PR #261, opened 2019-05-29
[ros2action] Support multiple part action type names for 'send_goal' verb

jacobperron:
Similar to the changes made in #247 and #259.

### DIFF
--- a/ros2action/ros2action/verb/send_goal.py
+++ b/ros2action/ros2action/verb/send_goal.py
@@ -82,7 +82,7 @@ def send_goal(action_name, action_type, goal_values, feedback_callback):
         try:
             # TODO(jacobperron): This logic should come from a rosidl related package
             parts = action_type.split('/')
-            if len(parts) < 2:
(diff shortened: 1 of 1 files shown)


ros2action/ros2action/verb/send_goal.py
```diff
@@ -79,12 +79,21 @@ def send_goal(action_name, action_type, goal_values, feedback_callback):
     node = None
     action_client = None
     try:
-        # TODO(jacobperron): This logic should come from a rosidl related package
-        package_name, action_type = action_type.split('/', 2)
-        if not package_name or not action_type:
+        try:
+            # TODO(jacobperron): This logic should come from a rosidl related package
+            parts = action_type.split('/')
+            if len(parts) < 2:
```
dirk-thomas: I would suggest to use `== 1` as in all other occurrences in #259.

jacobperron: Done 17cdf4e

PR #266, opened 2019-05-29
[ros2lifecycle] Only return the state for the node requested

### LINKED ISSUE #233: ros2 lifecycle get always shows state of all nodes even when selecting one by name
## Bug report

When calling ros2 lifecycle get /nodename, the state of _all_ nodes is shown. Expected behavior is that only the state of the selected node is shown.

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - Crystal from binary, Pre-Dashing from source
- Version or commit hash:
  - 1766de05961a7a08a77f446a22fca0abdd6d0552
- DDS implementation:
  - Fast-RTPS
  - rclpy

#### Steps to reproduce issue

```
[...]:~/ros2_dashing_ws$ ros2 lifecycle get
/damped_pendulum/damped_pendulum: inactive [2]
/transport_delay/transport_delay: unconfigured [1]
[...]:~/ros2_dashing_ws$ ros2 lifecycle get /damped_pendulum/damped_pendulum 
inactive [2]
unconfigured [1]
[...]:~/ros2_dashing_ws$ 
```

#### Expected behavior
Only the state of the selected node is shown.

#### Actual behavior
States of all nodes are shown - without names.

jacobperron:
Previously, the 'get' verb was returning the state for all lifecycle nodes.

Fixes #233

### DIFF
--- a/ros2lifecycle/ros2lifecycle/verb/get.py
+++ b/ros2lifecycle/ros2lifecycle/verb/get.py
@@ -49,7 +49,7 @@ class GetVerb(VerbExtension):
                 return 'Node not found'
 
         with DirectNode(args) as node:
-            states = call_get_states(node=node, node_names=node_names)
+            states = call_get_states(node=node, node_names=[node_name])
 
             # output exceptions
             for node_name in sorted(states.keys()):
--- a/ros2lifecycle/ros2lifecycle/verb/get.py
+++ b/ros2lifecycle/ros2lifecycle/verb/get.py
@@ -25,12 +25,13 @@ from ros2node.api import NodeNameCompleter
 
 
 class GetVerb(VerbExtension):
-    """Get lifecycle state."""
+    """Get lifecycle state for one or more nodes."""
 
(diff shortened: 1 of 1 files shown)


ros2lifecycle/ros2lifecycle/verb/get.py
```diff
@@ -49,7 +49,7 @@ def main(self, *, args):  # noqa: D102
                 return 'Node not found'
 
         with DirectNode(args) as node:
-            states = call_get_states(node=node, node_names=node_names)
+            states = call_get_states(node=node, node_names=[node_name])
```
dirk-thomas: The argument is optional so this won't work when nothing is passed.

jacobperron: Huh, I guess the behavior should be documented for what occurs when no node name is provided. I assume it should output the state for all nodes.

jacobperron: Fixed in 0c60f98 and clarified documentation in 080f249.

Issue #260, opened 2019-05-29
[ros2component] standalone verb creates zombie containers


jacobperron:
## Bug report

**Required Info:**

- Operating System:
  - macOS 
- Installation type:
  - Binaries 
- Version or commit hash:
  - 
- DDS implementation:
  - Fast-RTPS 
- Client library (if applicable):
  - N/A 

#### Steps to reproduce issue

Run the `ros2 component standalone` with invalid arguments, for example:
```
ros2 component standalone foo bar
```

#### Expected behavior

The command fails and no component container is running.

#### Actual behavior

The command fails, as expected, with the output:

```
$ ros2 component standalone foo bar
[ERROR] [standalone_container_d28808be9256]: Could not find requested resource in ament index
Failed to load component: Could not find requested resource in ament index
```

But, we can see a container left running:

```
$ ros2 component list
/standalone_container_d28808be9256
```

Multiple invalid calls will result in multiple zombie containers.

mjcarroll:
I can confirm this behavior on `Bionic` source build.

jubeira:
This doesn't seem to happen in Dashing testing binaries for Linux :+1:

PR #269, opened 2019-05-29
[ros2component] Stop the standalone container when load fails

mjcarroll:
This prevents zombie container processes from hanging around when a
component fails to load.

Closes: #260

### DIFF
--- a/ros2component/ros2component/verb/standalone.py
+++ b/ros2component/ros2component/verb/standalone.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import signal
 import uuid
 
 from ros2cli.node.direct import DirectNode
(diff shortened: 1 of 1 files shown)


ros2component/ros2component/verb/standalone.py
```diff
@@ -39,13 +41,21 @@ def main(self, *, args):
...
+                load_component_into_container(
+                    node=node, remote_container_node_name=args.container_node_name,
+                    package_name=args.package_name, plugin_name=args.plugin_name,
+                    node_name=args.node_name, node_namespace=args.node_namespace,
+                    log_level=args.log_level, remap_rules=args.remap_rules,
+                    parameters=args.parameters, extra_arguments=args.extra_arguments
+                )
+            except RuntimeError as ex:
+                # In case the component fails to load, kill the container.
+                print("{}, terminating container.".format(ex))
```
dirk-thomas: Since the code only sends `SIGINT`s the message should mentioned "interrupting" or similar since the process is not being "terminated" (with a `SIGTERM`).

mjcarroll: 3b9f983

mjcarroll:
@jacobperron I have verified this on Linux and Mac, do you want to check if it resolved the issue for you?

jacobperron:
Fixes the issue for me!

PR #271, opened 2019-06-07
Add 'topic find' verb

artivis:
Add the `find` verb which lists all topics of a given type.
E.g.  
```terminal
$ ros2 topic find rcl_interfaces/msg/Log
/rosout
```
or with `count-topics` option,  
```terminal
$ ros2 topic find rcl_interfaces/msg/Log -c
1
```

### DIFF
--- a/ros2topic/ros2topic/verb/find.py
+++ b/ros2topic/ros2topic/verb/find.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 from ros2cli.node.strategy import NodeStrategy
-from ros2topic.api import TopicTypeCompleter
+from ros2msg.api import message_type_completer
 from ros2topic.api import get_topic_names_and_types
 from ros2topic.verb import VerbExtension
 
@@ -25,7 +25,7 @@ class FindVerb(VerbExtension):
         arg = parser.add_argument(
             'topic_type',
             help="Name of the ROS topic type to filter for (e.g. 'std_msg/msg/String')")
-        arg.completer = TopicTypeCompleter()
+        arg.completer = message_type_completer
         parser.add_argument(
             '-c', '--count-topics', action='store_true',
             help='Only display the number of topics discovered')
--- a/ros2topic/ros2topic/verb/find.py
+++ b/ros2topic/ros2topic/verb/find.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 from ros2cli.node.strategy import NodeStrategy
+from ros2topic.api import TopicTypeCompleter
 from ros2topic.api import get_topic_names_and_types
 from ros2topic.verb import VerbExtension
 
@@ -21,9 +22,10 @@ class FindVerb(VerbExtension):
     """Output a list of available topics of a given type."""
 
     def add_arguments(self, parser, cli_name):
-        parser.add_argument(
+        arg = parser.add_argument(
             'topic_type',
             help="Name of the ROS topic type to filter for (e.g. 'std_msg/msg/String')")
+        arg.completer = TopicTypeCompleter()
         parser.add_argument(
             '-c', '--count-topics', action='store_true',
             help='Only display the number of topics discovered')
--- a/ros2topic/ros2topic/verb/find.py
+++ b/ros2topic/ros2topic/verb/find.py
@@ -1,0 +1,50 @@
+# Copyright 2019 Canonical Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


ros2topic/ros2topic/verb/find.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def main(self, *, args):
+        with NodeStrategy(args) as node:
+            topic_names_and_types = get_topic_names_and_types(
+                node=node,
+                include_hidden_topics=args.include_hidden_topics)
+
+        filtered_topics = []
+        for (topic_name, topic_type) in topic_names_and_types:
+            if args.topic_type in topic_type:
```
artivis: Not sure if there should be a special consideration in case there are several `topic_type` ?

ros2topic/ros2topic/verb/find.py
```diff
@@ -0,0 +1,50 @@
...
+from ros2topic.verb import VerbExtension
+
+
+class FindVerb(VerbExtension):
+    """Output a list of available topics of a given type."""
+
+    def add_arguments(self, parser, cli_name):
+        parser.add_argument(
+            'topic_type',
+            help="Name of the ROS topic type to filter for (e.g. 'std_msg/msg/String')")
```
hidmic: @artivis consider adding [autocompletion support](https://github.com/ros2/ros2cli/blob/master/ros2topic/ros2topic/api/__init__.py#L75) for topic types.

ros2topic/ros2topic/verb/find.py
```diff
@@ -0,0 +1,50 @@
...
+        parser.add_argument(
+            '-c', '--count-topics', action='store_true',
```
hidmic: @artivis isn't `--count-topics` somewhat redundant? Why not just `--count`?

ros2topic/setup.py
```diff
@@ -40,6 +40,7 @@
             'info = ros2topic.verb.info:InfoVerb',
             'list = ros2topic.verb.list:ListVerb',
             'pub = ros2topic.verb.pub:PubVerb',
+            'find = ros2topic.verb.find:FindVerb',
```
dirk-thomas: Please insert the new verb in alphabetical order.

ros2topic/ros2topic/verb/find.py
dirk-thomas: Please add a completer for this argument. See other verbs for an example, e.g. https://github.com/ros2/ros2cli/blob/e5b0ceb73b5380a745a3dccba17c078ce42aa2c1/ros2topic/ros2topic/verb/info.py#L27-L28

ros2topic/ros2topic/verb/find.py
hidmic: What situation do you envision?

ros2topic/ros2topic/verb/find.py
hidmic: You beat me to it :)

artivis: I guess `TopicNameCompleter` autocompletes topics' names, what about type ?

ros2topic/ros2topic/verb/find.py
artivis: Actually none.

ros2topic/ros2topic/verb/find.py
artivis: I'm actually following what was done for [`list` ](https://github.com/ros2/ros2cli/blob/master/ros2topic/ros2topic/verb/list.py#L30)

ros2topic/ros2topic/verb/find.py
artivis: Nevermind, found `TopicTypeCompleter` :+1:

dirk-thomas: When looking for the `TopicNameCompleter` there is a similar completer for types: `TopicTypeCompleter`. It finds only the type for a given topic though (if a topic is given). Of no topic is given it uses `ros2msg.api.message_type_completer` which you can use directly for this use case.

artivis: If no topic is passed to `TopicNameCompleter` it lists all topics types. That works too.

artivis: Resolving since this point was raised by @dirk-thomas.

dirk-thomas: Since you don't need the logic of `TopicNameCompleter` please call `ros2msg.api.message_type_completer` directly.

ros2topic/ros2topic/verb/find.py
artivis: It's just that `topic_names_and_types` is a weird structure ;)

ros2topic/ros2topic/verb/find.py
artivis: Done :+1:

dirk-thomas:
Thank you for the improvement and for iterating on it.

CI build to check linters: [Build Status](https://ci.ros2.org/job/ci_linux/7316/)

PR #272, opened 2019-06-07
add 'topic type' verb

artivis:
Add the `type` verb which return the type of a given topic.
E.g.
```
$ ros2 topic type /rosout
rcl_interfaces/msg/Log
```

### DIFF
--- a/ros2topic/ros2topic/verb/type.py
+++ b/ros2topic/ros2topic/verb/type.py
@@ -1,0 +1,39 @@
+# Copyright 2019 Canonical Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from ros2cli.node.direct import DirectNode
+from ros2topic.api import get_topic_names_and_types
+from ros2topic.api import TopicNameCompleter
+from ros2topic.verb import VerbExtension
+
+
+class TypeVerb(VerbExtension):
+    """Print a topic's type."""
+
+    def add_arguments(self, parser, cli_name):
+        arg = parser.add_argument(
+            'topic_name',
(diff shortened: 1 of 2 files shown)


artivis:
Came out of discussing https://github.com/ros2/ros2cli/pull/204.

ros2topic/ros2topic/verb/type.py
```diff
@@ -0,0 +1,39 @@
...
+from ros2topic.verb import VerbExtension
+
+
+class TypeVerb(VerbExtension):
+    """Print a topic's type."""
+
+    def add_arguments(self, parser, cli_name):
+        arg = parser.add_argument(
+            'topic_name',
+            help="Name of the ROS topic to get info (e.g. '/chatter')")
```
dirk-thomas: The help text should mention the type instead of "info".

ros2topic/ros2topic/verb/type.py
```diff
@@ -0,0 +1,39 @@
...
+            include_hidden_topics_key='include_hidden_topics')
+
+    def main(self, *, args):
+        with DirectNode(args) as node:
+            topic_names_and_types = get_topic_names_and_types(
+                node=node,
+                include_hidden_topics=args.include_hidden_topics)
+
+        for (topic_name, topic_types) in topic_names_and_types:
+            if args.topic_name == topic_name:
```
dirk-thomas: If the topic name is not found the function should return a non-zero exit code to indicate that it failed to find the given topic.

ros2topic/ros2topic/verb/type.py
```diff
@@ -0,0 +1,39 @@
...
+                print(topic_types[0])
```
dirk-thomas: By definition there can't be two tuples with the same topic name. So ince the topic is found the loop can return early to improve performance.

ros2topic/ros2topic/verb/type.py
artivis: Fixed.

ros2topic/ros2topic/verb/type.py
artivis: Went with a plain `return 1`, is that what you meant ?

ros2topic/ros2topic/verb/type.py
artivis: Went with a plain `return 0`, in mirror of the "non-zero exit code".

dirk-thomas:
Thanks for the patch and the quick iterations!

Linux CI: [Build Status](https://ci.ros2.org/job/ci_linux/7315/)

PR #273, opened 2019-06-07
Add 'service type' verb

artivis:
Add the `type` verb to `service` which return the type of a given service.
E.g.
given `ros2 run demo_nodes_cpp talker`
```terminal
$ ros2 service type /talker/describe_parameters
rcl_interfaces/srv/DescribeParameters
```

### DIFF
--- a/ros2service/ros2service/verb/type.py
+++ b/ros2service/ros2service/verb/type.py
@@ -36,7 +36,8 @@ class TypeVerb(VerbExtension):
 
         for (service_name, service_types) in service_names_and_types:
             if args.service_name == service_name:
-                print(service_types[0])
+                for service_type in service_types:
+                    print(service_type)
                 return 0
 
         return 1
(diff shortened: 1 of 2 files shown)


ros2service/ros2service/verb/type.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def main(self, *, args):
+        with NodeStrategy(args) as node:
+            service_names_and_types = get_service_names_and_types(
+                node=node,
+                include_hidden_services=args.include_hidden_services)
+
+        for (service_name, service_types) in service_names_and_types:
+            if args.service_name == service_name:
+                print(service_types[0])
```
dirk-thomas: Why should this only print the first type? (I guess the same question applies to the previous topic PR.)

artivis: First, can it really have several types ?? If so then this is a mistake.  
I only print the first one for a formatting purpose,  I could do the same with a loop I guess.

dirk-thomas: > can it really have several types ??

Yes, it can have multiple types. That is why the API returns a list of type names.

> I could do the same with a loop I guess.

:+1: one type per line sounds good.

artivis: Done.

dirk-thomas:
Thanks for the feature addition.

[Build Status](https://ci.ros2.org/job/ci_linux/7326/)

Issue #277, opened 2019-06-08
ros2 topic echo does not support messages generated directly from IDL


davidhodo:
## Bug report
- Operating System:
  -MacOS 10.14.5 (Mojave)
- Installation type:
  - Binaries
- Version or commit hash:
  - Dashing
- DDS implementation:
  - Fast-RTPS

#### Steps to reproduce issue
ROS2 topic echo currently hardcodes '.msg' for importing message python modules.  Message packages created directly from idl files as supported in Dashing use a '.idl' name.  Running 'ros2 topic echo' on a topic that uses a message generated from idl fails.

An example package that creates a simple string message using idl can be found at: [ros2_idl_demo](https://github.com/davidhodo/ros2_idl_demo).  Building the message (idl_msgs) and sample publisher (idl_pub) packages and then running `ros2 run idl_pub idl_pub` and `ros2 topic echo /chatter` results in a crash.

#### Expected behavior
```
---
var_string: 'ROS2: Hello'
---
var_string: 'ROS2: Hello'

```

#### Actual behavior

```
davidhodo@Davids-MacBook-Pro:idl_test_ws∫ ros2 topic echo /chatter
Traceback (most recent call last):
  File "/opt/ros/dashing/bin/ros2", line 11, in <module>
    load_entry_point('ros2cli==0.7.0', 'console_scripts', 'ros2')()
  File "/opt/ros/dashing/lib/python3.7/site-packages/ros2cli/cli.py", line 69, in main
    rc = extension.main(parser=parser, args=args)
  File "/opt/ros/dashing/lib/python3.7/site-packages/ros2topic/command/topic.py", line 43, in main
    return extension.main(args=args)
  File "/opt/ros/dashing/lib/python3.7/site-packages/ros2topic/verb/echo.py", line 69, in main
    return main(args)
  File "/opt/ros/dashing/lib/python3.7/site-packages/ros2topic/verb/echo.py", line 79, in main
    node.node, args.topic_name, args.message_type, callback)
  File "/opt/ros/dashing/lib/python3.7/site-packages/ros2topic/verb/echo.py", line 106, in subscriber
    msg_module = import_message_type(topic_name, message_type)
  File "/opt/ros/dashing/lib/python3.7/site-packages/ros2topic/api/__init__.py", line 71, in import_message_type
    module = importlib.import_module(package_name + '.' + middle_module)
[...]

Issue #223, opened 2019-04-17
Add ros2interface


jacobperron:
#### Feature description

Originally proposed in https://github.com/ros2/ros2cli/issues/202#issuecomment-472540339.

Replace `ros2 msg` and `ros2 srv` commands with `ros2 interface`, which also can handle action definitions (and possible more general IDL definitions).

Example usage:

```sh
# List all .msg, .srv, and .action types available
 $ ros2 interface list
 Messages:
     # Message types listed here
 Services:
     # Service types listed here
 Actions:
     # Action types listed here

 # Only list .msg types
 $ ros2 interface list --only-msg
 Messages:
     # Message types listed here

 # List all packages that generate messages, services, or actions
 $ ros2 interface packages
 action_msgs
 builtin_interfaces
 diagnostic_msgs
 # etc ...

 # Only list packages that generate actions
 $ ros2 interface packages --only-actions
 action_msgs
 example_interfaces
 test_msgs
 # etc ...
 
 # Show an interface definition
 $ ros2 interface show example_interfaces/Fibonacci
 Action:
     # Goal
     int32 order
     ---
     # Result
     int32[] sequence
     ---
     # Feedback
     int32[] sequence
 
 # In case of ambiguity (e.g. message and service with same name), show both
 $ ros2 interface show test_msgs/Primitives
 Message:
     bool bool_value
     byte byte_value
     char char_value
     # etc ...
 Service:
     bool bool_value
     byte byte_value
     char char_value
     # etc ...
     ---
     bool bool_value
     byte byte_value
     char char_value
     # etc ...
```

artivis:
After the merge of #288, will `ros2 msg` and `ros2 srv` be deprecated or are they all going to live side by side ?

jacobperron:
> After the merge of #288, will ros2 msg and ros2 srv be deprecated or are they all going to live side by side ?

IMO they could be deprecated for Eloquent, but open for discussion.

jacobperron:
Closed by #288

PR #278, opened 2019-06-08
parse middle_module from message type in ros2topic

davidhodo:
Attempts to solve issue #277.  The middle module name is parsed from the message type rather than being hardcoded as .msg.  I have tested it with a few message types generated from .msg and from .idl but not exhaustively.  Is there a case where the message_type split could result in more than 3 strings?

### DIFF
--- a/ros2topic/ros2topic/api/__init__.py
+++ b/ros2topic/ros2topic/api/__init__.py
@@ -58,7 +58,9 @@ def import_message_type(topic_name, message_type):
     # TODO(dirk-thomas) this logic should come from a rosidl related package
     try:
         package_name, middle_module, message_name = message_type.split('/')
-        if not package_name or not middle_module or not message_name:
+        if not middle_module:
+            middle_module = 'msg'
+        if not package_name  or not message_name:
             raise ValueError()
     except ValueError:
         raise RuntimeError('The passed message type is invalid')
--- a/ros2topic/ros2topic/api/__init__.py
+++ b/ros2topic/ros2topic/api/__init__.py
@@ -57,19 +57,18 @@ class TopicNameCompleter:
 def import_message_type(topic_name, message_type):
     # TODO(dirk-thomas) this logic should come from a rosidl related package
     try:
-        package_name, *message_name = message_type.split('/')
-        if not package_name or not message_name or not all(message_name):
+        package_name, middle_module, message_name = message_type.split('/')
+        if not package_name or not middle_module or not message_name:
             raise ValueError()
     except ValueError:
         raise RuntimeError('The passed message type is invalid')
 
     # TODO(sloretz) node API to get topic types should indicate if action or msg
-    middle_module = 'msg'
     if topic_name.endswith('/_action/feedback'):
         middle_module = 'action'
 
     module = importlib.import_module(package_name + '.' + middle_module)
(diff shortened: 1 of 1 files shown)


jacobperron:
It's undecided if we'll allow message types with more than three parts. There's an open proposal that there could be more than three: https://github.com/ros2/design/pull/220
Until it is decided, I don't think we need to worry about supporting that use-case.

davidhodo:
Ok.  I updated the request to default to 'msg' if the type only contains two parts.  Can this be merged or is it a wontfix in light of the planned new features listed in #223 mentioned by Dirk in #277?

ros2topic/ros2topic/api/__init__.py
```diff
@@ -57,19 +57,20 @@ def __call__(self, prefix, parsed_args, **kwargs):
 def import_message_type(topic_name, message_type):
     # TODO(dirk-thomas) this logic should come from a rosidl related package
     try:
-        package_name, *message_name = message_type.split('/')
-        if not package_name or not message_name or not all(message_name):
+        package_name, middle_module, message_name = message_type.split('/')
+        if not middle_module:
```
dirk-thomas: This logic doesn't seem to make any sense. In which case would `not middle_module` be `True`?

ros2topic/ros2topic/api/__init__.py
```diff
@@ -57,19 +57,20 @@ def __call__(self, prefix, parsed_args, **kwargs):
...
             raise ValueError()
     except ValueError:
         raise RuntimeError('The passed message type is invalid')
 
     # TODO(sloretz) node API to get topic types should indicate if action or msg
-    middle_module = 'msg'
     if topic_name.endswith('/_action/feedback'):
         middle_module = 'action'
 
     module = importlib.import_module(package_name + '.' + middle_module)
```
jacobperron: Perhaps we should just support any number of parts, e.g. `'.'.join([package_name] + message_name[:-1])`.

Karsten1987: Does it not generally make sense to unify this logic and have it in a central place? I know of at least one more other location where we duplicate the same logic:
https://github.com/ros2/ros2cli/blob/master/ros2service/ros2service/verb/call.py#L61-L72

jacobperron: Yes. There's a TODO at the top of this function. I'm not sure what package it should belong in (or if a new package needs to be created.

nuclearsandwich:
@davidhodo will you be able to address the feedback on this PR?

davidhodo:
Yes. Sorry for the long delay.  I will try to get it done this week.

nuclearsandwich:
@davidhodo are you still interested in updating this PR?

jacobperron:
I think this PR is obsolete by #415.

I'll check if #277 is resolved by #415 and if not, we should probably target a PR against https://github.com/ros2/rosidl_runtime_py.

jacobperron:
If we want this fix for Dashing, then I suggest re-opening this PR and changing the base branch to `dashing`. Trying to backport #415 and it's dependencies is much more work.

davidhodo:
Sorry.  I am still interested, but have been swamped and haven't been able to get back to it.  I'll rebase on Dashing and re-open as soon as I get some time.

PR #280, opened 2019-06-11
Add pkg xml verb

artivis:
Add the `xml` verb to `pkg`.  
By default it prints the whole xml manifest of a given package. User can request for a particular tag with the `--tag` option.

E.g.
```terminal
$ ros2 pkg xml demo_nodes_cpp -t version
0.7.6
$
$ ros2 pkg xml demo_nodes_cpp -t test_depend
ament_cmake_pytest
ament_lint_auto
ament_lint_common
launch
launch_testing
launch_testing_ament_cmake
launch_testing_ros
```

### DIFF
--- a/ros2pkg/ros2pkg/verb/xml.py
+++ b/ros2pkg/ros2pkg/verb/xml.py
@@ -35,9 +35,8 @@ class XmlVerb(VerbExtension):
             'package_name',
             help='The package name')
         arg.completer = package_name_completer
-        arg = parser.add_argument(
+        parser.add_argument(
             '-t', '--tag',
-            default='None',
             help="The package's xml's tag to print")
 
     def main(self, *, args):
@@ -47,17 +46,16 @@ class XmlVerb(VerbExtension):
             return PACKAGE_NOT_FOUND
 
         package_xml = os.path.join(package_share_dir, 'package.xml')
-        if os.path.isfile(package_xml):
-            tree = ET.parse(package_xml)
-            if args.tag == 'None':
-                ET.dump(tree)
-                return 0
-            else:
-                elements = tree.getroot().findall(args.tag)
-                if len(elements) > 0:
-                    for element in elements:
-                        print(element.text)
-                else:
-                    return PACKAGE_XML_TAG_NOT_FOUND
-        else:
+        if not os.path.isfile(package_xml):
             return PACKAGE_XML_NOT_FOUND
+
+        tree = ET.parse(package_xml)
+        if args.tag is None:
+            ET.dump(tree)
+            return 0
+        else:
+            elements = tree.getroot().findall(args.tag)
+            if len(elements) == 0:
+                return PACKAGE_XML_TAG_NOT_FOUND
+            for element in elements:
+                    print(element.text)
--- a/ros2pkg/ros2pkg/verb/xml.py
+++ b/ros2pkg/ros2pkg/verb/xml.py
@@ -1,0 +1,57 @@
+# Copyright 2017 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+
+import xml.etree.ElementTree as ET
+
+from ament_index_python import get_package_share_directory
+from ament_index_python import PackageNotFoundError
+from ros2pkg.api import package_name_completer
+from ros2pkg.verb import VerbExtension
+
+
+PACKAGE_NOT_FOUND = 'Package not found'
+PACKAGE_XML_TAG_NOT_FOUND = 'XML tag not found'
+
+
+class XmlVerb(VerbExtension):
+    """Output the package xml manifest or an optional tag."""
+
+    def add_arguments(self, parser, cli_name):
+        arg = parser.add_argument(
+            'package_name',
+            help='The package name')
+        arg = parser.add_argument(
+            '-t', '--tag',
+            default='root',
+            help="The package's xml's tag to print")
+        arg.completer = package_name_completer
+
+    def main(self, *, args):
+        try:
+            package_share_dir = get_package_share_directory(args.package_name)
+            package_xml = os.path.join(package_share_dir, 'package.xml')
+            if os.path.isfile(package_xml):
+                tree = ET.parse(package_xml)
+                if args.tag == 'root':
+                    ET.dump(tree)
+                    return 0
+                else:
+                    for elements in tree.getroot().findall(args.tag):
+                        print(elements.text)
+                    return 0
+                return PACKAGE_XML_TAG_NOT_FOUND
+        except PackageNotFoundError:
+            return PACKAGE_NOT_FOUND
--- a/ros2pkg/ros2pkg/verb/xml.py
+++ b/ros2pkg/ros2pkg/verb/xml.py
@@ -27,7 +27,7 @@ PACKAGE_XML_TAG_NOT_FOUND = 'XML tag not found'
 
 
 class XmlVerb(VerbExtension):
-    """Output the package xml manifest or an optional tag."""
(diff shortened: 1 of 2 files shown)


artivis:
I did not find the use of any particular xml parser lib thus I used `xml.etree.ElementTree`. Let me know if there is actually one lying around.

ros2pkg/ros2pkg/verb/xml.py
```diff
@@ -0,0 +1,57 @@
...
+from ros2pkg.api import package_name_completer
+from ros2pkg.verb import VerbExtension
+
+
+PACKAGE_NOT_FOUND = 'Package not found'
+PACKAGE_XML_TAG_NOT_FOUND = 'XML tag not found'
+
+
+class XmlVerb(VerbExtension):
+    """Output the package xml manifest or an optional tag."""
```
dirk-thomas: Wording: I would suggest "... or a specific tag.". "Optional tag" has an ambiguous meaning.

ros2pkg/ros2pkg/verb/xml.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def add_arguments(self, parser, cli_name):
+        arg = parser.add_argument(
+            'package_name',
+            help='The package name')
+        arg = parser.add_argument(
+            '-t', '--tag',
+            default='root',
+            help="The package's xml's tag to print")
+        arg.completer = package_name_completer
```
dirk-thomas: The completer should be added to the `package_name` argument - not the `--tag`.

ros2pkg/ros2pkg/verb/xml.py
```diff
@@ -0,0 +1,57 @@
...
+                tree = ET.parse(package_xml)
+                if args.tag == 'root':
+                    ET.dump(tree)
+                    return 0
+                else:
+                    for elements in tree.getroot().findall(args.tag):
+                        print(elements.text)
+                    return 0
+                return PACKAGE_XML_TAG_NOT_FOUND
+        except PackageNotFoundError:
```
dirk-thomas: Please keep the scope of the `try` / `except` block as small as possible - only around the lines which can actually raise that exception.

ros2pkg/ros2pkg/verb/xml.py
```diff
@@ -0,0 +1,57 @@
...
+            if os.path.isfile(package_xml):
+                tree = ET.parse(package_xml)
+                if args.tag == 'root':
+                    ET.dump(tree)
+                    return 0
+                else:
+                    for elements in tree.getroot().findall(args.tag):
+                        print(elements.text)
+                    return 0
+                return PACKAGE_XML_TAG_NOT_FOUND
```
dirk-thomas: This line isn't reachable?

ros2pkg/ros2pkg/verb/xml.py
```diff
@@ -0,0 +1,57 @@
...
+            '-t', '--tag',
+            default='root',
+            help="The package's xml's tag to print")
+        arg.completer = package_name_completer
+
+    def main(self, *, args):
+        try:
+            package_share_dir = get_package_share_directory(args.package_name)
+            package_xml = os.path.join(package_share_dir, 'package.xml')
+            if os.path.isfile(package_xml):
```
dirk-thomas: In this condition is `False` the script returns with no output and no error code.

ros2pkg/ros2pkg/verb/xml.py
```diff
@@ -0,0 +1,57 @@
...
+class XmlVerb(VerbExtension):
+    """Output the package xml manifest or an optional tag."""
+
+    def add_arguments(self, parser, cli_name):
+        arg = parser.add_argument(
+            'package_name',
+            help='The package name')
+        arg = parser.add_argument(
+            '-t', '--tag',
+            default='root',
```
dirk-thomas: Maybe use a different default - e.g. the default default `None`. That way if in the future the manifest would contain `root` tag this doesn't fail.

ros2pkg/ros2pkg/verb/xml.py
artivis: Done.

ros2pkg/ros2pkg/verb/xml.py
artivis: Done.

ros2pkg/ros2pkg/verb/xml.py
artivis: Fixed the overall logic.

ros2pkg/ros2pkg/verb/xml.py
artivis: Done.

ros2pkg/ros2pkg/verb/xml.py
artivis: Fixed.

ros2pkg/ros2pkg/verb/xml.py
artivis: Fixed the overall logic.

ros2pkg/ros2pkg/verb/xml.py
```diff
@@ -0,0 +1,63 @@
...
+
+class XmlVerb(VerbExtension):
+    """Output the package xml manifest or a specific tag."""
+
+    def add_arguments(self, parser, cli_name):
+        arg = parser.add_argument(
+            'package_name',
+            help='The package name')
+        arg.completer = package_name_completer
+        arg = parser.add_argument(
```
dirk-thomas: This variable `arg` isn't used anywhere.

ros2pkg/ros2pkg/verb/xml.py
```diff
@@ -0,0 +1,63 @@
...
+            help="The package's xml's tag to print")
+
+    def main(self, *, args):
+        try:
+            package_share_dir = get_package_share_directory(args.package_name)
+        except PackageNotFoundError:
+            return PACKAGE_NOT_FOUND
+
+        package_xml = os.path.join(package_share_dir, 'package.xml')
+        if os.path.isfile(package_xml):
```
dirk-thomas: May I suggest to invert the logic to return early (like above when the package is not found). That will avoid all the remaining code to be indented:

```
if not os.path.isfile(package_xml):
    return PACKAGE_XML_NOT_FOUND

<...>
```

The same pattern could be applied to the following conditions to avoid the `else` case (not necessary though - just an idea).

ros2pkg/ros2pkg/verb/xml.py
```diff
@@ -0,0 +1,63 @@
...
+    """Output the package xml manifest or a specific tag."""
+
+    def add_arguments(self, parser, cli_name):
+        arg = parser.add_argument(
+            'package_name',
+            help='The package name')
+        arg.completer = package_name_completer
+        arg = parser.add_argument(
+            '-t', '--tag',
+            default='None',
```
dirk-thomas: The default values shouldn't be the string `'None'` but the type `None`. If you leave the `default` keyword argument out `None` is already the default value.

ros2pkg/ros2pkg/verb/xml.py
```diff
@@ -0,0 +1,63 @@
...
+    def main(self, *, args):
+        try:
+            package_share_dir = get_package_share_directory(args.package_name)
+        except PackageNotFoundError:
+            return PACKAGE_NOT_FOUND
+
+        package_xml = os.path.join(package_share_dir, 'package.xml')
+        if os.path.isfile(package_xml):
+            tree = ET.parse(package_xml)
+            if args.tag == 'None':
```
dirk-thomas: This check should then be `args.tag is None:`.

ros2pkg/ros2pkg/verb/xml.py
artivis: Removed.

ros2pkg/ros2pkg/verb/xml.py
artivis: Fixed.

ros2pkg/ros2pkg/verb/xml.py
artivis: Fixed.

ros2pkg/ros2pkg/verb/xml.py
artivis: The exit condition is now checked first.

dirk-thomas:
Thanks for the quick updates. I made a few minor tweaks in 180fb5c0634be1975657a89a801edb2f24120bcb:
* capitalize the word "XML"
* update the help text for "--tag" to include an example
* avoid the `else` if the `if` block calls `return`
* fix the indentation in the `for` loop

This looks great - thank you for contributing this feature.

PR #282, opened 2019-06-12
add --ttl option to multicast send

dirk-thomas:
Replaces #224.

### DIFF
--- a/ros2multicast/ros2multicast/api/__init__.py
+++ b/ros2multicast/ros2multicast/api/__init__.py
@@ -19,8 +19,11 @@ DEFAULT_GROUP = '225.0.0.1'
 DEFAULT_PORT = 49150
 
 
-def send(data, *, group=DEFAULT_GROUP, port=DEFAULT_PORT):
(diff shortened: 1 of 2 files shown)


ros2multicast/ros2multicast/api/__init__.py
```diff
@@ -19,8 +19,11 @@
 DEFAULT_PORT = 49150
 
 
-def send(data, *, group=DEFAULT_GROUP, port=DEFAULT_PORT):
+def send(data, *, group=DEFAULT_GROUP, port=DEFAULT_PORT, ttl=None):
     s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM, socket.IPPROTO_UDP)
+    if ttl is not None:
+        packed_ttl = struct.pack('b', ttl)
```
mjcarroll: Is this required for a particular platform?  I found some examples where the packing was not used.

dirk-thomas: It is the recommended approach: https://docs.python.org/3/library/socket.html#socket.socket.setsockopt

An integer might work too. I just used the same approach as in the original PR.

PR #284, opened 2019-06-14
Make daemon “reset” itself when the IP address changes

ivanpauno:
The problem is that when the IP address changes (e.g.: by DHCP), discovery doesn't work all the time.

This is a first proposal for solving the problem. 
I'm doing the following:
  The methods registered in the `XMLRPC` server now check if one IP has changed, and in that case the `DirectNode` used by the Daemon is reset.

When the `DirectNode` is reset, the requested method is immediately call. Maybe an `spin_time` should be added (for ensuring enough discovery time).

With "checking that one IP has changed", I'm referring to the following -> If the IP address of any interface that has a gateway has changed, the `DirectNode` is reset.

As far as I understand, it's not possible to know what interface is the `rmw` implementation using.

Python doesn't have in the standard library a package for listing network interfaces, etc (in a platform independent way).
I'm using [netifaces](https://pypi.org/project/netifaces/) for that. I don't know if we want to add a new dependency or not.
In case that we don't want to add it, I can maybe add this as an optional plugin in another package.

For trying this, you have to install netifaces: `python3 -m pip install -U netifaces`

### DIFF
--- a/ros2cli/ros2cli/daemon/__init__.py
+++ b/ros2cli/ros2cli/daemon/__init__.py
@@ -115,7 +115,7 @@ class _DirectNode:
 
         def wrapper():
             self.reset_if_addresses_changed()
-            return self.node.__getattr__(name)
+            return self.node.__getattr__(name)()
         wrapper.__name__ = attr.__name__
 
         if inspect.ismethod(attr):
--- a/ros2cli/ros2cli/daemon/__init__.py
+++ b/ros2cli/ros2cli/daemon/__init__.py
@@ -14,11 +14,15 @@
 
 import argparse
 from collections import namedtuple
+import inspect
 import os
 from xmlrpc.server import SimpleXMLRPCRequestHandler
 from xmlrpc.server import SimpleXMLRPCServer
 
+import netifaces
+
 import rclpy
+
 from ros2cli.node.direct import DirectNode
 
 
@@ -48,7 +52,7 @@ def main(*, script_name='_ros2_daemon', argv=None):
     node_args = NodeArgs(
         node_name_suffix='_daemon_%d' % args.ros_domain_id,
         start_parameter_services=False)
-    with DirectNode(node_args) as node:
+    with _DirectNode(node_args) as node:
         server = LocalXMLRPCServer(
             addr, logRequests=False, requestHandler=RequestHandler,
             allow_none=True)
@@ -90,6 +94,69 @@ def main(*, script_name='_ros2_daemon', argv=None):
                 pass
         finally:
             server.server_close()
+
+
+class _DirectNode:
+    """A direct node, that resets itself when a network interface changes."""
+
+    def __init__(self, args):
+        self.args = args
+        self.node = DirectNode(args)
+        self.addresses_at_start = self.interfaces_ip_addresses()
+
+    def __enter__(self):
+        self.node.__enter__()
+        return self
+
+    def __getattr__(self, name):
+        attr = self.node.__getattr__(name)
+
+        def wrapper():
+            self.reset_if_addresses_changed()
+            return self.node.__getattr__(name)
+        wrapper.__name__ = attr.__name__
+
+        if inspect.ismethod(attr):
+            return wrapper
+        self.reset_if_addresses_changed()
+        return attr
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.node.__exit__(exc_type, exc_value, traceback)
+
+    def interfaces_ip_addresses(self):
+        addresses_by_interfaces = {}
+        for (kind, info) in netifaces.gateways().items():
+            if kind not in (netifaces.AF_INET, netifaces.AF_INET6):
+                continue
+            if isinstance(info, dict):
+                continue
+            info_list = info
+            if isinstance(info, tuple):
+                info_list = [info]
+            addresses_by_interfaces[kind] = {}
+            for info in info_list:
+                # info[1] is the interface name
+                addresses_by_interfaces[kind][info[1]] = netifaces.ifaddresses(info[1])[kind][0]['addr']
+        return addresses_by_interfaces
+
+    def reset_if_addresses_changed(self):
+        def have_addresses_changed(new_addresses):
+            for (kind, addresses_by_interfaces) in self.addresses_at_start.items():
+                for (interface, addr) in addresses_by_interfaces.items():
+                    try:
+                        if new_addresses[kind][interface] != addr:
+                            return True
+                    except KeyError:
+                        return True
+            return False
+        new_addresses = self.interfaces_ip_addresses()
+        if have_addresses_changed(new_addresses):
+            self.addresses_at_start = new_addresses
+            self.node.destroy_node()
+            rclpy.shutdown()
+            self.node = DirectNode(self.args)
+            self.node.__enter__()
 
 
 class LocalXMLRPCServer(SimpleXMLRPCServer):
--- a/ros2cli/ros2cli/daemon/__init__.py
+++ b/ros2cli/ros2cli/daemon/__init__.py
@@ -14,6 +14,7 @@
 
 import argparse
 from collections import namedtuple
+import functools
 import inspect
 import os
 from xmlrpc.server import SimpleXMLRPCRequestHandler
@@ -96,6 +97,27 @@ def main(*, script_name='_ros2_daemon', argv=None):
             server.server_close()
 
 
+def get_interfaces_ip_addresses():
+    addresses_by_interfaces = {}
+    for (kind, info) in netifaces.gateways().items():
+        if kind not in (netifaces.AF_INET, netifaces.AF_INET6):
+            continue
+        print('Interface kind: {}, info: {}'.format(kind, info))
+        if isinstance(info, dict):
+            continue
+        info_list = info
+        if isinstance(info, tuple):
+            info_list = [info]
+        addresses_by_interfaces[kind] = {}
+        for item in info_list:
+            interface_name = info[1]
+            addresses_by_interfaces[kind][interface_name] = (
+                netifaces.ifaddresses(interface_name)[kind][0]['addr']
+            )
+    print('Addresses by interfaces: {}'.format(addresses_by_interfaces))
+    return addresses_by_interfaces
+
+
 class _DirectNode:
     """A direct node, that resets itself when a network interface changes."""
 
@@ -104,21 +126,21 @@ class _DirectNode:
         # TODO(ivanpauno): A race condition is possible here, since it isn't possible to know
         # exactly which interfaces were available at node creation.
         self.node = DirectNode(args)
-        self.addresses_at_start = self.interfaces_ip_addresses()
+        self.addresses_at_start = get_interfaces_ip_addresses()
 
     def __enter__(self):
         self.node.__enter__()
         return self
 
     def __getattr__(self, name):
-        attr = self.node.__getattr__(name)
-
-        def wrapper():
-            self.reset_if_addresses_changed()
-            return self.node.__getattr__(name)()
-        wrapper.__name__ = attr.__name__
+        attr = getattr(self.node, name)
 
         if inspect.ismethod(attr):
+            @functools.wraps(attr)
+            def wrapper(*args, **kwargs):
+                self.reset_if_addresses_changed()
+                return getattr(self.node, name)(*args, **kwargs)
+            wrapper.__signature__ = inspect.signature(attr)
             return wrapper
         self.reset_if_addresses_changed()
         return attr
@@ -126,26 +148,8 @@ class _DirectNode:
     def __exit__(self, exc_type, exc_value, traceback):
         self.node.__exit__(exc_type, exc_value, traceback)
 
-    def interfaces_ip_addresses(self):
-        addresses_by_interfaces = {}
-        for (kind, info) in netifaces.gateways().items():
-            if kind not in (netifaces.AF_INET, netifaces.AF_INET6):
-                continue
-            print('Interface kind: {}, info: {}'.format(kind, info))
-            if isinstance(info, dict):
-                continue
-            info_list = info
-            if isinstance(info, tuple):
-                info_list = [info]
-            addresses_by_interfaces[kind] = {}
-            for item in info_list:
-                # info[1] is the interface name
-                addresses_by_interfaces[kind][item[1]] = netifaces.ifaddresses(item[1])[kind][0]['addr']
-        print('Addresses by interfaces: {}'.format(addresses_by_interfaces))
-        return addresses_by_interfaces
-
     def reset_if_addresses_changed(self):
-        new_addresses = self.interfaces_ip_addresses()
+        new_addresses = get_interfaces_ip_addresses()
         if new_addresses != self.addresses_at_start:
             self.addresses_at_start = new_addresses
             self.node.destroy_node()
--- a/ros2cli/ros2cli/daemon/__init__.py
+++ b/ros2cli/ros2cli/daemon/__init__.py
@@ -1,4 +1,4 @@
-# Copyright 2017 Open Source Robotics Foundation, Inc.
+# Copyright 2017-2019 Open Source Robotics Foundation, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


wjwwood:
@codebot can you please ask someone on the SOSS project to try this out? Maybe Grey?

ivanpauno:
@codebot I will try to reproduce a similar scenario here.
Does the interface address of Machine A change too?

ivanpauno:
@codebot I'm not being able to reproduce your problem.
I ran in one machine a node. In another machine I ran different `ros2` commands and verbs (node list, topic list), and I haven't seen a "discovery" problem. In the middle of running the commands, I have manually switched the IP of the computer a bunch of times. Both machines were connected using WiFi.
I also tried with the `master` branch of this repo, which doesn't reset the daemon when an interface changes, and I couldn't reproduce the problem either.

Could you provide me some other details?:
- rmw implementation you're using
- Operative system

And anything else you think is useful for reproducing the issue correctly.

dirk-thomas:
@codebot Just checking if you came up with any further information on this topic.

dirk-thomas:
@codebot Friendly ping.

claireyywang:
OS: machine A (linux desktop), machine B (linux VM on laptop), ubuntu 18.04
RMW: fastrtps
ROS distro: Eloquent 
I'm getting slightly different behaviour when testing this branch. I was not able to discover the topic publishing on machine A from machine B, but was able to use multicast both ways. I also encountered the same `ros2 topic list` returning blank behaviour, which was temporarily fixed by `ros2 daemon stop`. I found a thread https://answers.ros.org/question/304373/ros2-nodes-can-not-see-one-another-via-network/ discussing the same issues I ran into. It seems like they are both related to `ros2 daemon`.

ivanpauno:
@claireyywang It would be good to have some additional information:

- > I was not able to discover the topic publishing on machine A from machine B, but was able to use multicast both ways

  I guess that you were initially able to discover topics on machine A from machine B, but after a network change it stopped working. In that case, how did you trigger the network change?

- I haven't been able to reproduce the problem in our network. I'm not to sure what can be the difference, but if you could provide detailed information on how the computers were connected, it may be useful.

My setup: I have two machines in the same LAN, using static IPs. I run a node in machine A (e.g. `ros2 run demo_nodes_cpp talker`) and in machine B I run `ros2 topic list`, the node is listed correctly.
I then manually change the IP of MACHINE B, and run `ros2 topic list` again, having the same result.
I didn't need to run `ros2 daemon stop`, and I was using the `master` branch of this repo (it also works using this branch).

- Does this happen using other DDS vendors?

----

I think there are two possible reasons why doing `ros2 daemon stop` solves the problem and this patch doesn't:

- An error in the detection of a network interface change. That's part of this PR ([this piece of code](https://github.com/ros2/ros2cli/blob/b8c9a7556ac535d7cac7949b5276c5b2c6275e2b/ros2cli/ros2cli/daemon/__init__.py#L129-L164)).
- This patch is currently just destroying a node, shutting done rclpy, and creating a new node when a network interface change is detected. That may not be enough, as the DDS implementation may have some global state.
  Instead of destroying the node and creating a new one, I could run a new daemon process and stop the original one, which is equivalent to manually run `ros2 daemon stop`.

I don't believe it's the first problem, as I have checked and tested the logic a bunch of times.
To be sure that's not the case, I added some debug messages.
For seeing them, you have to use `ros2 daemon start --debug` (just in cast, first run `ros2 daemon stop`).
Each time you run `ros2 topic list`, you should see something like this:
```
get_topic_names_and_types()
Interface kind: 2, info: [('10.34.0.1', 'enx70886b89a852', True), ('10.34.0.1', 'wlp5s0', False)]
Addresses by interfaces: {2: {'enx70886b89a852': '10.34.0.67', 'wlp5s0': '10.34.0.104'}}
__LIST_OF_NODES__ (it may be empty)
```
And if some network interface changed, you should see:
```
get_topic_names_and_types()
Interface kind: 2, info: [('10.34.0.1', 'enx70886b89a852', True)]
Addresses by interfaces: {2: {'enx70886b89a852': '10.34.0.67'}}
Daemon node was reset
__LIST_OF_NODES__ (it may be empty)
```

@claireyywang Could you test this and check if you see the `Daemon node was reset` message when you run a cli command (e.g.: `ros2 topic list`) after a change happened in a network interface?
If you're seeing that message, I can discard the first problem and I'm sure it's the second.

If you're not seeing that, please copy all your logs here (the ones when you were able to discover topics in the other machine, and the ones after you stopped being able).

claireyywang:
Thanks for the suggestions @ivanpauno! Here are some updates.
machine A (Linux VM on laptop), machine B (Linux desktop)
`master` branch:
Machine B publish to topic `/testingmaster` the whole time. Machine A and B start with being on the same wifi network. 
Machine A was able to discover the topic published by B. After machine A was switched to a mobile hotspot network, A still lists `/testingmaster` in `ros2 topic list`, and only removed it after I ran `ros2 daemon stop`. Same thing happened after I switched A back to the same wifi network as B; A was not able to discover `/testingmaster` until daemon was restarted. 

`reset-daemon` branch: 
(followed your steps and saw the `Daemon node was reset` message)
Machine B publishing to topic `/testingmaster`. A and B on the same wifi network.
Machine A was not able to discover the topic (could also be experiencing the same delay as in master branch). I ran `ros2 daemon stop` and `ros2 daemon start --debug`, then `ros2 topic list` and got this output
```

```
After switching to a different wifi network, I ran `ros2 topic list`, got this output 
```

```

ivanpauno:
Thanks for checking @claireyywang !

> Machine A was able to discover the topic published by B. After machine A was switched to a mobile hotspot network, A still lists /testingmaster in ros2 topic list, and only removed it after I ran ros2 daemon stop.

Though participant removal is detected fast (e.g.: when you `ctrl-c` a node), dropping a participant (i.e. when not being able to communicate with it for a while) is inherently slow. If you wait some minutes, `/testingmaster` should stop appearing (I have tested that, and it finally disappears).
Resetting the daemon node should make the process faster.

>  Same thing happened after I switched A back to the same wifi network as B; A was not able to discover /testingmaster until daemon was restarted.

I was finally able to reproduce this.
Originally, I was just trying to change the IP and not to switch between networks.
AFAIU, network interfaces are checked when a Participant is created, and the `Participant` doesn't detect a network change (at least, that seems to be the case for `Fast-RTPS`).
Now I'm forcing to reset the daemon node when a new interface is detected (before, I was just resetting it when an interface disappeared).

> reset-daemon branch:

I corrected some bugs in https://github.com/ros2/ros2cli/pull/284/commits/9767d0f0e0a5af51e734b8c737f3ab2e35fd2b64 and forced to reset the daemon node when a new network interface is detected in https://github.com/ros2/ros2cli/pull/284/commits/d7fd7388fdedbeebe8b074a2d6cbb9bb9535fd14.
After that, both problems are solved.
@claireyywang could you please check if this is working for you?

----

I wonder if this only happens with `Fast-RTPS` or if it also happens with other `rmw` implementations.
I will check that on Monday.

claireyywang:
I confirm it is working. I switched machine between two different networks and the delay is significantly shorter; topics are correctly discovered without me doing anything to the daemon. Seems to me that both problems are solved indeed.

ros2cli/ros2cli/daemon/__init__.py
```diff
@@ -92,6 +96,65 @@ def shutdown_handler():
...
+
+        if inspect.ismethod(attr):
+            return wrapper
+        self.reset_if_addresses_changed()
+        return attr
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.node.__exit__(exc_type, exc_value, traceback)
+
+    def interfaces_ip_addresses(self):
```
hidmic: @ivanpauno it seems like this method could be a free function.

ros2cli/ros2cli/daemon/__init__.py
```diff
@@ -92,6 +96,65 @@ def shutdown_handler():
...
+                continue
+            print('Interface kind: {}, info: {}'.format(kind, info))
+            if isinstance(info, dict):
+                continue
+            info_list = info
+            if isinstance(info, tuple):
+                info_list = [info]
+            addresses_by_interfaces[kind] = {}
+            for item in info_list:
+                # info[1] is the interface name
```
hidmic: @ivanpauno why not `interface_name = info[1]` instead of a comment?

ros2cli/ros2cli/daemon/__init__.py
```diff
@@ -92,6 +96,65 @@ def shutdown_handler():
...
+        # exactly which interfaces were available at node creation.
+        self.node = DirectNode(args)
+        self.addresses_at_start = self.interfaces_ip_addresses()
+
+    def __enter__(self):
+        self.node.__enter__()
+        return self
+
+    def __getattr__(self, name):
+        attr = self.node.__getattr__(name)
```
hidmic: @ivanpauno nit: consider 
```suggestion
        attr = getattr(self.node, name)
```

ros2cli/ros2cli/daemon/__init__.py
```diff
@@ -92,6 +96,65 @@ def shutdown_handler():
...
+
+        def wrapper():
+            self.reset_if_addresses_changed()
+            return self.node.__getattr__(name)()
+        wrapper.__name__ = attr.__name__
+
+        if inspect.ismethod(attr):
+            return wrapper
```
hidmic: @ivanpauno consider:
```python
if inspect.ismethod(attr):
    @functools.wraps(attr)
    def wrapper(*args, **kwargs):
        self.reset_if_addresses_changed()
        return getattr(self.node, name)(*args, **kwargs)
    wrapper.__signature__ = inspect.signature(attr)
```
instead

ros2cli/ros2cli/daemon/__init__.py
```diff
@@ -92,6 +96,65 @@ def shutdown_handler():
...
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.node.__exit__(exc_type, exc_value, traceback)
+
+    def interfaces_ip_addresses(self):
+        addresses_by_interfaces = {}
+        for (kind, info) in netifaces.gateways().items():
+            if kind not in (netifaces.AF_INET, netifaces.AF_INET6):
+                continue
+            print('Interface kind: {}, info: {}'.format(kind, info))
+            if isinstance(info, dict):
```
hidmic: @ivanpauno what are we leaving out with this `if` clause?

ros2cli/ros2cli/daemon/__init__.py
```diff
@@ -92,6 +96,65 @@ def shutdown_handler():
             server.server_close()
 
 
+class _DirectNode:
```
hidmic: @ivanpauno `DaemonNode` ?

ros2cli/ros2cli/daemon/__init__.py
ivanpauno: Is this needed?
`wrapper.__signature__ = inspect.signature(attr)`

ros2cli/ros2cli/daemon/__init__.py
ivanpauno:
```python3
import netifaces
netifaces.gateways().items()
```

output:
```
dict_items([('default', {2: ('172.17.0.1', 'eth0')}), (2, [('172.17.0.1', 'eth0', True)])])
```

As you can see, the default interface value is a dictionary like `{kind: info_tuple}`,
though the others interface values are a list of `info_tuple`.
I think it would be clearer if I do:
```python3
if 'default' == kind:
    continue
```

do you agree?

ivanpauno: Also, I'm checking, but I think this other if wasn't needed:
https://github.com/ros2/ros2cli/blob/106b33459f79d2a883999b15f3df3485fadab762/ros2cli/ros2cli/daemon/__init__.py#L138-L139

ros2cli/ros2cli/daemon/__init__.py
ivanpauno: Can be confused with https://github.com/ros2/ros2cli/blob/3feb662510fc2b84c9f6aed8ee7af2f6aa82b8f8/ros2cli/ros2cli/node/daemon.py#L96.

I have some options:
- AutoResettingNode
- NetworkAutoResettingNode
- NetworkAwareNode
- DaemonDirectNode
- NodeWithAutoresettingPower

ros2cli/ros2cli/daemon/__init__.py
ivanpauno: Also, I think I can delete this if https://github.com/ros2/ros2cli/blob/106b33459f79d2a883999b15f3df3485fadab762/ros2cli/ros2cli/daemon/__init__.py#L138-L139.

Will double check.

ros2cli/ros2cli/daemon/__init__.py
ivanpauno: AFAIK, `functools.wraps` does that automatically.

hidmic: It is needed if we want to keep the decorator as transparent as possible, and no, `functools.wraps` does not force any function signature, only naming.

ivanpauno: See https://github.com/ros2/launch/pull/279#discussion_r335518599.

hidmic: See https://github.com/ros2/launch/pull/279#discussion_r335626803, though it's arguably a detail that has no impact on this particular code (at least for the time being).

ros2cli/ros2cli/daemon/__init__.py
ivanpauno: Ok, I simplified the logic.
`'default'` was already ruled out in:
```python3
if kind not in (netifaces.AF_INET, netifaces.AF_INET6):
    continue
```

> Is it guaranteed it'll be the only interface value that is a dict?

Yes. For further reference, see https://pypi.org/project/netifaces/.

ros2cli/ros2cli/node/daemon.py
```diff
@@ -72,11 +72,14 @@ def spawn_daemon(args, wait_until_spawned=None):
...
+    cmd.extend([
         # the arguments are only passed for visibility in e.g. the process list
         '--rmw-implementation', rmw_implementation_identifier,
-        '--ros-domain-id', str(ros_domain_id)],
-        stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL, **kwargs)
+        '--ros-domain-id', str(ros_domain_id)])
+    if not hasattr(args, 'debug') or not args.debug:
+        kwargs['stdout'] = subprocess.DEVNULL
+        kwargs['stderr'] = subprocess.DEVNULL
+    subprocess.Popen(cmd, **kwargs)
```
hidmic: @ivanpauno meta: I had not noticed, and it predates this PR, but this is a peculiar way to demonize a process.

ros2cli/ros2cli/daemon/__init__.py
ivanpauno: As a proof, you run do the following:
```python3
import functools
import inspect

def asd(bsd, csd):
    print('asd')

@functools.wraps(asd)
def wrapper(*args, **kwargs):
    asd(*args, **kwargs)

print(inspect.signature(wrapper) == inspect.signature(asd))
```

The result is `True`.

ros2cli/ros2cli/node/daemon.py
```diff
@@ -72,11 +72,14 @@ def spawn_daemon(args, wait_until_spawned=None):
...
             'try specifying the implementation to use via the '
             "'RMW_IMPLEMENTATION' environment variable")
-    subprocess.Popen(cmd + [
+    cmd.extend([
         # the arguments are only passed for visibility in e.g. the process list
         '--rmw-implementation', rmw_implementation_identifier,
-        '--ros-domain-id', str(ros_domain_id)],
-        stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL, **kwargs)
+        '--ros-domain-id', str(ros_domain_id)])
+    if not hasattr(args, 'debug') or not args.debug:
```
hidmic: @ivanpauno hmm all CLIs that use `NodeStrategy` and happen to have a `debug` option may unintentionally hook up the daemon to their standard streams. Consider adding another argument to the `spawn_daemon` function instead.

ros2cli/ros2cli/node/daemon.py
ivanpauno: I don't know what is the correct way, but in any case, I would make that modification in a follow-up PR.

ros2cli/ros2cli/node/daemon.py
ivanpauno: See https://github.com/ros2/ros2cli/pull/284/commits/e33912323089920a28a992bc1410d56d1ee014c1.

ivanpauno:
CI in https://github.com/ros2/ci/pull/346#issuecomment-543346038.
Going in!

PR #288, opened 2019-06-24
Added ros2interface to replace ros2 msg/srv

skucheria:
Addressing https://github.com/ros2/ros2cli/issues/223


skucheria:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7422/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3573/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6090/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7245/)

ros2interface/ros2interface/verb/show.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def add_arguments(self, parser, cli_name):
+        arg = parser.add_argument(
+                'type',
+                help='Show an interface definition (e.g. "std_msgs/msg/String")')
+        arg.completer = type_completer
+
+    def main(self, *, args):
+        try:
+            parts = args.type.split('/')
```
jacobperron: Since the format of types is decided by rosidl, I would add a TODO that this logic should be moved to a rosidl related package.

ros2interface/ros2interface/verb/show.py
```diff
@@ -0,0 +1,56 @@
...
+            if len(parts) < 3:
```
jacobperron: I think it's possible that a type has 2 parts (though I don't any of our examples do).

ros2interface/ros2interface/verb/show.py
```diff
@@ -0,0 +1,56 @@
...
+                raise ValueError()
+            if not parts[0] or not parts[1] or not parts[2]:
```
jacobperron: There could be more than 3 parts. You can check if each part is non-empty with `all(parts)`.
Having whitespace is also not valid (e.g. "pkg/my namespace/MsgType"), so it would be nice to guard against that as well.

ros2interface/ros2interface/verb/show.py
```diff
@@ -0,0 +1,56 @@
...
+                raise ValueError()
+            # now have to check if msg or srv
```
jacobperron: I think this comment can be removed or moved below?

ros2interface/ros2interface/verb/show.py
```diff
@@ -0,0 +1,56 @@
...
+        except ValueError:
+            raise RuntimeError('The passed interface type is invalid')
+
+        try:
```
jacobperron: Since this try block is catching/raising the same error as the previous it can be combined with previous try block.

ros2interface/ros2interface/verb/show.py
```diff
@@ -0,0 +1,56 @@
...
+            if parts[1] == 'msg':
+                path = get_message_path(parts[0], parts[2])
+            elif parts[1] == 'srv':
+                path = get_service_path(parts[0], parts[2])
+            elif parts[1] == 'action':
+                path = get_action_path(parts[0], parts[2])
+            else:
+                raise ValueError()
```
jacobperron: Since we want to support generic IDL files, it would be better to have a generic `get_interface_path` function. I don't think it would need any special logic based on the middle namespace.

ros2interface/package.xml
```diff
@@ -0,0 +1,30 @@
...
+  <version>0.0.0</version>
+  <description>TODO: Package description</description>
+  <maintainer email="dthomas@osrfoundation.org">Dirk Thomas</maintainer>
+  <license>TODO: License declaration</license>
+
+
+  <depend>ros2cli</depend>
+  <depend>ros2action</depend>
+  <depend>ros2msg</depend>
+  <depend>ros2srv</depend>
```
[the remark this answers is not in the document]
skucheria: I'm using functions from those packages api's so I think I need them as dependencies. Unless I move those over to the ros2interface api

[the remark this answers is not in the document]
jacobperron: I would copy/make variants of those functions in this package (I expect most of them will have TODOs for moving the implementation to a rosidl package anyways).

skucheria:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7441/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3586/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6103/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7278/)

ros2interface/ros2interface/api/__init__.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+def package_name_completer(**kwargs):
+    """Callable returning a list of package name containing messages, services, and action."""
+    return get_all_types()
+
+
+def type_completer(**kwargs):
+    """Callable returning a list of message, service, and action types."""
+    types = []
```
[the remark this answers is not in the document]
jacobperron: My comment still stands.

ros2interface/ros2interface/api/__init__.py
```diff
@@ -0,0 +1,150 @@
...
+        content, _ = get_resource('rosidl_interfaces', package_name)
+    except LookupError:
+        return []
+    interface_names = content.splitlines()
+    # TODO(jacobperron) this logic should come from a rosidl related package
+    # Only return actions in action folder
+    return list(sorted({
+        n[7:-7]
+        for n in interface_names
+        if n.startswith('action/') and n[-7:] in ('.idl', '.action')}))
```
jacobperron: The condition `n[-7:] in ('.idl')` looks troublesome. It would only be true the full name is `.idl`, which is not going to happen. This should probably be broken up into two separate conditions.

ros2interface/ros2interface/verb/list.py
```diff
@@ -0,0 +1,70 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from ros2interface.api import get_all_action_types
+from ros2interface.api import get_all_message_types
+from ros2interface.api import get_all_service_types
+from ros2interface.verb import VerbExtension
+
+
+def printMessages():
```
jacobperron: style nit: use snake_case for function names

ros2interface/ros2interface/verb/list.py
```diff
@@ -0,0 +1,70 @@
...
+    print('Messages: ')
+    message_types = get_all_message_types()
+    for package_name in sorted(message_types.keys()):
```
jacobperron: I don't think `.key()` is necessary.

ros2interface/ros2interface/api/__init__.py
```diff
@@ -0,0 +1,150 @@
...
+        raise LookupError('Unknown package name')
+    try:
+        content, _ = get_resource('rosidl_interfaces', package_name)
+    except LookupError:
+        return []
+    interface_names = content.splitlines()
+    # TODO(dirk-thomas) this logic should come from a rosidl related package
+    # Only return services in srv folder
+    return list(sorted({
+        n[4:-4]
```
jacobperron: AFAICT, wherever we use `get_service_types`, we are re-inserting the the `srv/` namespace.
Unless there is a reason to not include the namespace in the names returned by this function, I would include it.
Same for the other `get_*_types()` functions.

ros2interface/ros2interface/verb/list.py
```diff
@@ -0,0 +1,70 @@
...
+def printActions():
+    print('Actions: ')
+    action_types = get_all_action_types()
+    for package_name in sorted(action_types.keys()):
+        for action_name in sorted(action_types[package_name]):
+            print('{package_name}/action/{action_name}'.format_map(locals()))
+
+
+class ListVerb(VerbExtension):
+    """List all .msg and .srv types available."""
```
jacobperron: This docstring should be updated to something like "List all interfaces types available."

ros2interface/ros2interface/verb/package.py
```diff
@@ -0,0 +1,35 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from ros2interface.api import get_types
+from ros2interface.api import package_name_completer
+from ros2interface.verb import VerbExtension
+
+
+class PackageVerb(VerbExtension):
+    """Output a list of available msg/srv types within one package."""
```
jacobperron: This docstring should be updated.

ros2interface/ros2interface/verb/packages.py
```diff
@@ -0,0 +1,46 @@
...
+from ros2interface.verb import VerbExtension
+
+
+def printPackages(packs):
+    for name in sorted(packs):
+        print((name))
+
+
+class PackagesVerb(VerbExtension):
+    """Output a list of packages which contain msgs and srvs."""
```
jacobperron: This docstring should be updated.

ros2interface/ros2interface/verb/packages.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def add_arguments(self, parser, cli_name):
+        parser.add_argument('-m', '--only-msgs', action='count',
+                            help='Only list packages that generate messages')
+
+        parser.add_argument('-s', '--only-srvs', action='count',
+                            help='Only list packages that generate services')
```
jacobperron: Add an option `--only-actions`.

ros2interface/ros2interface/verb/packages.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def main(self, *, args):
+        packages = set()
+        all_interface_types = get_all_types()
```
jacobperron: We don't need to make this query for two of the three cases below. It can be moved into the `else:` block.

ros2interface/ros2interface/verb/list.py
skucheria: Why would it not be? Since message_types is a dict with the keys being the package names

ros2interface/ros2interface/api/__init__.py
```diff
@@ -0,0 +1,78 @@
...
+    for item in msgsList:
+        typeList.append(item)
+    for item in srvsList:
+        typeList.append(item)
+    return typeList
+
+
+def package_name_completer(**kwargs):
+    """Callable returning a list of package name containing messages, services, and action."""
+    return get_all_types()
```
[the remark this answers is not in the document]
skucheria: That function returns a list of package names, its a misleading function name. Will change it.

ros2interface/ros2interface/verb/list.py
jacobperron: Be default, a for-in loop will iterate overs the keys of a dict.

skucheria:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7450/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3595/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6112/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7287/)

ros2interface/test/test_api.py
```diff
@@ -0,0 +1,53 @@
...
+    interface_names = get_interface('std_srvs')
+    assert len(interface_names) == 3
+
+
+def test_get_interface_path():
+    # check known package name
+    get_interface('std_msgs')
+    path = ['std_msgs', 'msg', 'Empty.msg']
+    interface_path = get_interface_path(path)
+    assert os.path.isfile(interface_path)
```
jacobperron: Add a test that for an invalid path.

ros2interface/test/test_api.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def test_get_all_interface_pacakges():
+    # testing get_all_interface_packages with 'packages' verb
```
jacobperron: From an organization view, I think this file should only be testing functions in `ros2interface.api`. I would move the verb test to it's own file or `test_cli.py`. In fact, it looks like it's already being smoke-tested in `test_cli.py`.

ros2interface/test/test_api.py
```diff
@@ -0,0 +1,53 @@
...
+    packages_cmd = ['ros2', 'interface', 'packages', '-a']
+    packages_result = subprocess.run(packages_cmd, stdout=subprocess.PIPE, check=True)
+    package_names = packages_result.stdout.decode().splitlines()
+    assert len(package_names) == 4
+
+    # check all packages being found
+    interface_packages = get_all_interface_packages()
+    assert len(interface_packages) == 26
```
jacobperron: This is a brittle assertion. If a new package containing an interface is added to the dependency tree, then this test will start failing. It would be better to assert about known interface packages that we are explicitly depending on,

```python
assert 'std_msgs' in interface_packages
assert 'std_srvs' in interface_packages
```

ros2interface/ros2interface/verb/packages.py
```diff
@@ -0,0 +1,52 @@
...
+        packages = set()
+        if args.only_msgs:
+            print_packages(get_all_message_types().keys())
+        elif args.only_srvs:
+            print_packages(get_all_service_types().keys())
+        elif args.only_actions:
+            print_packages(get_all_action_types().keys())
+        else:
+            all_interface_packages = get_all_interface_packages()
+            for package in all_interface_packages:
```
jacobperron: This for-loop looks redundant. Can we just pass `all_interface_packages` directly to the print function? Then we don't need the `packages` variable on L41.

ros2interface/ros2interface/api/__init__.py
```diff
@@ -0,0 +1,152 @@
...
+        content, _ = get_resource('rosidl_interfaces', package_name)
+    except LookupError:
+        return []
+    interface_names = content.splitlines()
+    # TODO(jacobperron) this logic should come from a rosidl related package
+    # Only return actions in action folder
+    return list(sorted({
+        n[7:-7]
+        for n in interface_names
+        if n.startswith('action/') and n[-7:] in ('.action')}))
```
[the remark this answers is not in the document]
skucheria: This doesn't work. It adds both .idl. and .action types. Checking it against just .action removes the duplicates and prints it out properly without the extension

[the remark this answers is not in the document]
jacobperron: I think it's because of L90 (which is assuming the extension is `.action`. I think it's better to keep support for `.idl` files in case someone defines an action in pure IDL.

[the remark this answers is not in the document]
skucheria: Ok. I was thinking of changing that too, but wasn't sure what would be better.

ros2interface/test/test_api.py
```diff
@@ -13,41 +13,39 @@
...
-    path = ['std_msgs', 'msg', 'Empty.msg']
+    # check for existing path
+    path = ['std_msgs', 'msg', 'Empty']
     interface_path = get_interface_path(path)
     assert os.path.isfile(interface_path)
 
+    # check for nonexistent path
+    try:
+        path = ['std_msgs', 'msg', 'InvalidPath']
+        get_interface_path(path)
```
jacobperron: Similarly, this can be tested in a `pytest.raises(FileNotFoundError)` context.

ros2interface/ros2interface/verb/package.py
```diff
@@ -0,0 +1,35 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from ros2interface.api import get_interface
+from ros2interface.api import package_name_completer
+from ros2interface.verb import VerbExtension
+
+
+class PackageVerb(VerbExtension):
+    """Output a list of available msg/srv types within one package."""
```
jacobperron:
```suggestion
    """Output a list of available interface types within one package."""
```

ros2interface/ros2interface/verb/packages.py
```diff
@@ -0,0 +1,49 @@
...
+from ros2interface.verb import VerbExtension
+
+
+def print_packages(packs):
+    for name in sorted(packs):
+        print((name))
+
+
+class PackagesVerb(VerbExtension):
+    """Output a list of packages which contain msgs, srvs, and actions."""
```
jacobperron:
```suggestion
    """Output a list of packages that provide interfaces."""
```

ros2interface/package.xml
```diff
@@ -0,0 +1,29 @@
...
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>ros2interface</name>
+  <version>0.7.4</version>
+  <description>
+    The interface command for ROS 2 command line tools
+  </description>
+  <maintainer email="jacob@osrfoundation.org">Jacob Perron</maintainer>
+  <license>Apache License 2.0</license>
+
```
jacobperron: nitpick: one empty line is sufficient

ros2interface/package.xml
```diff
@@ -0,0 +1,29 @@
...
+  <exec_depend>ament_index_python</exec_depend>
+
+  <test_depend>ament_copyright</test_depend>
+  <test_depend>ament_flake8</test_depend>
+  <test_depend>ament_pep257</test_depend>
+  <test_depend>ament_xmllint</test_depend>
+  <test_depend>python3-pytest</test_depend>
+  <test_depend>std_msgs</test_depend>
+  <test_depend>std_srvs</test_depend>
+
```
jacobperron: ditto

ros2interface/ros2interface/verb/list.py
```diff
@@ -0,0 +1,70 @@
...
+from ros2interface.api import get_all_service_types
+from ros2interface.verb import VerbExtension
+
+
+def print_messages():
+    print('Messages: ')
+    message_types = get_all_message_types()
+    for package_name in sorted(message_types):
+        for message_name in sorted(message_types[package_name]):
+            print('{package_name}/msg/{message_name}'.format_map(locals()))
```
jacobperron: IMO, indenting the messages will look nicer
```suggestion
            print('    {package_name}/msg/{message_name}'.format_map(locals()))
```

Same for service and actions.

ros2interface/ros2interface/verb/show.py
```diff
@@ -0,0 +1,43 @@
...
+            if len(parts) < 2:
+                raise ValueError()
+            if not all(parts):
+                raise ValueError()
+            path = get_interface_path(parts)
+        except ValueError:
+            raise RuntimeError('The passed interface type is invalid')
+        except LookupError as e:
+            return str(e)
+        with open(path, 'r') as h:
```
jacobperron: If the path does not exist, we should print a nicer error message. Right now we get an ugly stack trace:

```
$ ros2 interface show std_msgs/msg/Foo                                                                      1
  File "/home/jacob/ws/ros2cli_ws/install/ros2cli/bin/ros2", line 11, in <module>
    load_entry_point('ros2cli==0.7.0', 'console_scripts', 'ros2')()
  File "/home/jacob/ws/ros2cli_ws/install/ros2cli/lib/python3.6/site-packages/ros2cli/cli.py", line 69, in main
  File "/home/jacob/ws/ros2cli_ws/install/ros2interface/lib/python3.6/site-packages/ros2interface/command/interface.py", line 35, in main
  File "/home/jacob/ws/ros2cli_ws/install/ros2interface/lib/python3.6/site-packages/ros2interface/verb/show.py", line 42, in main
    with open(path, 'r') as h:
FileNotFoundError: [Errno 2] No such file or directory: '/opt/ros/dashing/share/std_msgs/msg/Foo.idl'
```

I would first check if the file exists and return a nice error message if it does not, e.g. `return 'Could not find the interface {!r}'.format('/'.join(parts)))`.

skucheria:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7455/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3599/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6116/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7292/)

ros2interface/ros2interface/api/__init__.py
```diff
@@ -0,0 +1,153 @@
...
+from ament_index_python import get_resource
+from ament_index_python import get_resources
+from ament_index_python import has_resource
+
+
+def get_all_interface_packages():
+    return get_resources('rosidl_interfaces')
+
+
+def get_interface(package_name):
```
dirk-thomas: Since the function returns a list the name should be plural: `get_interfaces`.

ros2interface/ros2interface/api/__init__.py
```diff
@@ -0,0 +1,153 @@
...
+    interface_names = content.splitlines()
+    return list(sorted({
+        n.rsplit('.', 1)[0]
+        for n in interface_names
+        if '_' not in n}))
+
+
+def get_interface_path(parts):
+    prefix_path = has_resource('packages', parts[0])
+    joined = '/'.join(parts)
```
dirk-thomas: This should be a platform specific separator.

ros2interface/ros2interface/api/__init__.py
```diff
@@ -0,0 +1,153 @@
...
+    return interface_path
+
+
+def package_name_completer(**kwargs):
+    """Callable returning a list of types containing messages, services, and action."""
+    return get_all_interface_packages()
+
+
+def type_completer(**kwargs):
+    """Callable returning a list of message, service, and action types."""
```
dirk-thomas: Where are the action types in this function?

ros2interface/ros2interface/api/__init__.py
```diff
@@ -0,0 +1,153 @@
...
+        content, _ = get_resource('rosidl_interfaces', package_name)
+    except LookupError:
+        return []
+    interface_names = content.splitlines()
+    # TODO(dirk-thomas) this logic should come from a rosidl related package
+    # Only return services in srv folder
+    return list(sorted({
+        n[4:-4]
+        for n in interface_names
+        if n.startswith('srv/') and n[-4:] in ('.idl', '.srv')}))
```
dirk-thomas: Several of these function look like copy-n-paste with only minimal differences. Please refactor to avoid the duplication. Also we do have a `rosidl_runtime_py` now so all the TODOs can be addressed.

ros2interface/ros2interface/verb/list.py
```diff
@@ -0,0 +1,70 @@
...
+# limitations under the License.
+
+from ros2interface.api import get_all_action_types
+from ros2interface.api import get_all_message_types
+from ros2interface.api import get_all_service_types
+from ros2interface.verb import VerbExtension
+
+
+def print_messages():
+    print('Messages: ')
```
dirk-thomas: Why a trailing space?

Same below.

ros2interface/ros2interface/verb/list.py
```diff
@@ -0,0 +1,70 @@
...
+        for message_name in sorted(message_types[package_name]):
+            print('    {package_name}/msg/{message_name}'.format_map(locals()))
+
+
+def print_services():
+    print('Services: ')
+    service_types = get_all_service_types()
+    for package_name in sorted(service_types):
+        for service_name in sorted(service_types[package_name]):
+            print('{package_name}/srv/{service_name}'.format_map(locals()))
```
dirk-thomas: Why are services and actions not indented but messages are?

ros2interface/ros2interface/verb/list.py
```diff
@@ -0,0 +1,70 @@
...
+    for package_name in sorted(action_types):
+        for action_name in sorted(action_types[package_name]):
+            print('{package_name}/action/{action_name}'.format_map(locals()))
+
+
+class ListVerb(VerbExtension):
+    """List all interface types available."""
+
+    def add_arguments(self, parser, cli_name):
+        parser.add_argument('-m', '--only-msgs', action='count',
```
dirk-thomas: Please wrap after the open parenthesis to follow style guide.

Same below.

ros2interface/ros2interface/verb/list.py
```diff
@@ -0,0 +1,70 @@
...
+                            help='Print out only the message types')
+
+        parser.add_argument('-s', '--only-srvs', action='count',
+                            help='Print out only the srvs types')
+
+        parser.add_argument('-a', '--only-actions', action='count',
+                            help='Print out only the action types')
+
+    def main(self, *, args):
+        if(args.only_msgs):
```
dirk-thomas: Nitpick: please don't use parenthesis around conditions if not necessary.

Same below.

ros2interface/ros2interface/verb/list.py
```diff
@@ -0,0 +1,70 @@
...
+            print_messages()
+            return
+        if(args.only_srvs):
+            print_services()
+            return
+        if(args.only_actions):
+            print_actions()
+            return
```
dirk-thomas: What if the user wants to see messages and services but no actions? That should be a supported use case of the CLI.

Same below.

ros2interface/ros2interface/verb/show.py
```diff
@@ -0,0 +1,43 @@
...
+            if len(parts) < 2:
+                raise ValueError()
+            if not all(parts):
+                raise ValueError()
+            file_path = get_interface_path(parts)
+        except ValueError:
+            raise RuntimeError('The passed interface type is invalid')
+        except LookupError as e:
+            return str(e)
+        with open(file_path, 'r') as h:
```
dirk-thomas: This must specify an `encoding='utf-8'`. And before the print it probably needs to encode into the encoding of the output stream.

ros2interface/setup.py
```diff
@@ -0,0 +1,42 @@
...
+        'ros2cli.command': [
+            'interface = ros2interface.command.interface:InterfaceCommand',
+        ],
+        'ros2cli.extension_point': [
+            'ros2interface.verb = ros2interface.verb:VerbExtension',
+        ],
+        'ros2interface.verb': [
+            'list = ros2interface.verb.list:ListVerb',
+            'packages = ros2interface.verb.packages:PackagesVerb',
+            'package = ros2interface.verb.package:PackageVerb',
```
dirk-thomas: Nitpick: swap order to be alphabetical.

ros2interface/ros2interface/verb/list.py
skucheria: This was resolved in a follow-up PR

dirk-thomas: > This was resolved in a follow-up PR

Please reference the follow-up PR / the change within it in case it is bigger.

skucheria: https://github.com/ros2/ros2cli/pull/289

ros2interface/ros2interface/command/interface.py
```diff
@@ -0,0 +1,35 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from ros2cli.command import add_subparsers
+from ros2cli.command import CommandExtension
+from ros2cli.verb import get_verb_extensions
+
+
+class InterfaceCommand(CommandExtension):
```
dirk-thomas: This class doesn't have a docblock and as a consquence their is no description (which is the first line of the docblock) when calling `ros2 --help`.

jacobperron: Added in https://github.com/ros2/ros2cli/pull/434

PR #294, opened 2019-07-10
Create README

maryaB-osr:
issue #157 
Would appreciate review for correct use of language, relevance of links, content, etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,24 @@
+# ros2cli
+
+This repository contains the source code for ROS 2 command line interface tools included with a standard install of any ROS 2 distro. 
+
+## Usage
+
+Run `ros2 --help` to see all available verbs.
+
+Run `ros2 <verb> --help` for more information on individual command usage.
+
+Read [Introspection with command line tools](https://index.ros.org/doc/ros2/Tutorials/Introspection-with-command-line-tools/) on ROS Index for more information and an example.
+
+### Cheat Sheet
+
+This [cheat sheet](https://github.com/artivis/ros2_cheats_sheet/blob/master/cli/cli_cheats_sheet.pdf) provides examples for verbs and their sub-verbs (some of which rely on the [ROS 2 demos package](https://github.com/ros2/demos)).
+
+## Add New Verbs
+
+You can use [Python entry points](https://setuptools.readthedocs.io/en/latest/pkg_resources.html#entry-points) to create new verbs and sub-verbs for the cli. 
+[Here's an example](https://github.com/ros2/ros2cli/pull/273/files).
+
+## Background
+
+You can find a [historical discussion](https://discourse.ros.org/t/ros-graph-information-tools-implementation-discussion/674/34) on this subject on Discourse.
--- a/README.md
+++ b/README.md
@@ -1,23 +1,26 @@
 # ros2cli
 
-This repository contains the source code for ROS 2 command line interface tools included with a standard install of any ROS 2 distro. 
+This repository contains the source code for ROS 2 command line interface tools included with a standard install of any ROS 2 distro.
 
 ## Usage
 
-Run `ros2 --help` to see all available verbs.
+Run `ros2 --help` to see all available commands.
 
-Run `ros2 <verb> --help` for more information on individual command usage.
+Run `ros2 <command> --help` for more information on individual command usage.
+
+Run `ros2 <command> <verb> --help` for even more usage information on a specific command's verbs.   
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -0,0 +1,24 @@
+# ros2cli
+
+This repository contains the source code for ROS 2 command line interface tools included with a standard install of any ROS 2 distro. 
+
+## Usage
+
+Run `ros2 --help` to see all available verbs.
+
+Run `ros2 <verb> --help` for more information on individual command usage.
```
ivanpauno: It's possible to run either:
```
ros2 <command> --help
```
and:
```
ros2 <command> <verb> --help
```

Maybe, it would be good to explicitly add both options here.

README.md
```diff
@@ -0,0 +1,24 @@
...
+
+Read [Introspection with command line tools](https://index.ros.org/doc/ros2/Tutorials/Introspection-with-command-line-tools/) on ROS Index for more information and an example.
+
+### Cheat Sheet
+
+This [cheat sheet](https://github.com/artivis/ros2_cheats_sheet/blob/master/cli/cli_cheats_sheet.pdf) provides examples for verbs and their sub-verbs (some of which rely on the [ROS 2 demos package](https://github.com/ros2/demos)).
```
ivanpauno: Nice! I didn't know about it.

ivanpauno: I think we use the words `command` and `verb` (and not `verb` and `subverb`) throughout the code. I don't know what the correct words are (the cheatsheet also use verb and subverb).

README.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Add New Verbs
+
+You can use [Python entry points](https://setuptools.readthedocs.io/en/latest/pkg_resources.html#entry-points) to create new verbs and sub-verbs for the cli. 
+[Here's an example](https://github.com/ros2/ros2cli/pull/273/files).
```
ivanpauno: We would need also an example on how to add a command. It's not exactly the same.
Here is an [example](https://github.com/ros2/ros2cli/pull/288/files), but it's too long.

README.md
ivanpauno: Do someone here has write access on that repo? If not, it would be good to have a local copy of that here (if it can be copied), or generate something similar locally.

mjcarroll: I think that @artivis would probably be open to collaborating on it, if we wanted to do PRs there.  I don't know where we would bring it in locally?

mjcarroll: I also don't know what the correct words are, so I would say pick something and be consistent with it?

artivis: I'm definitely open to collaboration/PRs or even move it as you see fit !
The Latex source is hosted on the repo after all ;)

artivis: Let me know about the appropriate wording so that I can update the cheats sheet if necessary :+1:

README.md
```diff
@@ -0,0 +1,24 @@
+# ros2cli
+
+This repository contains the source code for ROS 2 command line interface tools included with a standard install of any ROS 2 distro.
```
artivis: Well except for [`sros2`](https://github.com/ros2/sros2)-related commands. Maybe there are others ??
CLI tools that are part of the standard install but not held here should be listed.

README.md
artivis: I guess we could put together a `ros2 hello world`

README.md
mjcarroll: And it's already Apache-2, how about that :)

ivanpauno: I think that using `command` and `verb` is more consistent with the existing code.

README.md
ivanpauno: That would be great!

README.md
ivanpauno: [ros2launch](https://github.com/ros2/launch_ros/tree/master/ros2launch), [ros2test](https://github.com/ros2/ros_testing/tree/master/ros2test) too. I don't know if I'm missing another extension.

clalancette: [ros2bag](https://github.com/ros2/rosbag2/tree/master/ros2bag)

README.md
maryaB-osr: `ros2 -h` uses `<command>` for both the first and the second argument, e.g. when I run `ros2 bag -h` it says `call 'ros2 bag <command> -h' for more detailed usage`. Any thoughts about that?

maryaB-osr: I just noticed it also says `Various rosbag related sub-commands`. So it seems like the code says uses different terminology in different places

README.md
artivis: FYI, wip [ros2hellocli](https://github.com/artivis/ros2hellocli).

PR #296, opened 2019-07-13
Add Pkg create for ament python

### LINKED ISSUE #125: pkg create for ament_python
There is currently no support for `ament_python` in `ros2 pkg create`. It defaults to `ament_cmake`, but there is no automated way to create python packages.

jpace121:
# Overview
This PR adds the build type `ament_python` to the list of build types that pkg create understands how to make template packages for. 

This is my first PR to this project, I'll be happy to take any suggestions.

# Related
Resolves #125


ros2pkg/ros2pkg/resource/ament_python/main.py.em
```diff
@@ -0,0 +1,7 @@
+import rclpy
+
+def main():
+    print("Hi from @project_name.")
+
+if __name__ == "__main__":
+    main()
```
Karsten1987: wrong return line
same below.

ros2pkg/ros2pkg/resource/ament_python/main.py.em
```diff
@@ -0,0 +1,7 @@
+import rclpy
```
Karsten1987: these imports should maybe be considered from the dependencies, but I think it's not required to have `rclpy` every time.

ros2pkg/ros2pkg/resource/ament_python/setup.py.em
```diff
@@ -0,0 +1,28 @@
...
+setup(
+    name=package_name,
+    version='0.0.0',
+    packages=[package_name],
+    data_files=[
+        ('share/ament_index/resource_index/packages',
+            ['resource/' + package_name]),
+        ('share/' + package_name, ['package.xml']),
+    ],
+    install_requires=['setuptools'@[for dep in dependencies]@ , '@dep'@[end for]@ ],
```
Karsten1987: I am not an expert on this, but I believe the `install_requires` are not for all dependencies, but only for the setuptools.

ros2pkg/ros2pkg/verb/create.py
```diff
@@ -28,6 +28,7 @@
 from ros2pkg.api.create import populate_cmake
 from ros2pkg.api.create import populate_cpp_library
 from ros2pkg.api.create import populate_cpp_node
+from ros2pkg.api.create import populate_ament_python
```
Karsten1987: this should be ordered alphabetically

ros2pkg/ros2pkg/verb/create.py
```diff
@@ -101,12 +105,24 @@ def main(self, *, args):
...
+                cpp_library_name = args.cpp_library_name
+            if args.cpp_node_name:
+                cpp_node_name = args.cpp_node_name
+                if args.cpp_node_name == args.cpp_library_name:
+                    cpp_node_name = args.cpp_node_name + '_node'
+                    print('[WARNING] node name can not be equal to the library name',
+                          file=sys.stderr)
+                    print('[WARNING] renaming node to %s' % cpp_node_name, file=sys.stderr)
+
+        python_node_name = None
```
Karsten1987: I could see that we can rename `cpp_node_name` to only `node_name` and recycle the same variable for a python and cpp project.

ros2pkg/ros2pkg/resource/ament_python/main.py.em
```diff
@@ -0,0 +1,7 @@
...
+
```
Karsten1987: should have a double empty line

ros2pkg/ros2pkg/resource/ament_python/main.py.em
```diff
@@ -0,0 +1,7 @@
...
+def main():
+    print("Hi from @project_name.")
+
```
Karsten1987: double empty line

ros2pkg/ros2pkg/resource/ament_python/main.py.em
```diff
@@ -0,0 +1,7 @@
+import rclpy
+
+def main():
+    print("Hi from @project_name.")
```
Karsten1987: single quotes

ros2pkg/ros2pkg/resource/ament_python/main.py.em
```diff
@@ -0,0 +1,7 @@
...
+
+if __name__ == "__main__":
```
Karsten1987: single quotes

ros2pkg/ros2pkg/resource/ament_python/main.py.em
jpace121: Thanks for the feedback. All of the formatting issues in the template should be fixed now. I was able to create an example project and then run flake8 on each file and have the example pass.

ros2pkg/ros2pkg/resource/ament_python/setup.py.em
jpace121: Ooops!

 You are correct. Should be fixed now. I was able to make a project with a rclpy dependency and build and run it.

ros2pkg/ros2pkg/resource/ament_python/main.py.em
jpace121: Removed rclpy in the last commit. This makes it more consistent with the ament_cmake template.

ros2pkg/ros2pkg/verb/create.py
jpace121: This is the only recommendation I haven't resolved. I will look into it when I have a few more cycles.

@Karsten1987 Thanks for the review.

Karsten1987: sounds good. I'd hold this PR back until this is resolved then.
We could maybe address that in a different, follow up PR, but I rather have that all settled in one PR than introducing new CLIs and then break them again :)

jpace121: Should be handled in d4d90b.

Thanks again for the review.

jpace121:
@Karsten1987 , sorry it took me a while for me to get back at this. Would you be willing to take another look?

ros2pkg/ros2pkg/verb/create.py
```diff
@@ -100,13 +101,23 @@ def main(self, *, args):
...
-        if args.cpp_node_name:
-            cpp_node_name = args.cpp_node_name
-            if args.cpp_node_name == args.cpp_library_name:
-                cpp_node_name = args.cpp_node_name + '_node'
-                print('[WARNING] node name can not be equal to the library name', file=sys.stderr)
-                print('[WARNING] renaming node to %s' % cpp_node_name, file=sys.stderr)
+        node_name = None
+        cpp_library_name = None
+        if args.build_type == 'ament_cmake' or args.build_type == 'cmake':
+            if args.cpp_library_name:
```
Karsten1987: when generalizing `cpp_library_name` to a general `library_name` this check can be performed in any case.

Karsten1987:
Also, before running CI, this PR has to be rebased on top of master.

piraka9011:
Was just running into this issue recently. Thanks @jpace121 for this.
One suggestion is to add a `data_files` entry for launch files to `ros2pkg/ros2pkg/resource/ament_python/setup.py.em` since it's really not explained _anywhere_ in the wiki or docs. 
(Opening another PR for that...)
```python
('share/' + package_name + '/launch', glob(os.path.join('launch', '*.launch.py')))
```
The end result would be: 
```python
data_files=[
    ('share/ament_index/resource_index/packages', ['resource/' + package_name]),
    ('share/' + package_name, ['package.xml']),
    ('share/' + package_name + '/launch', glob(os.path.join('launch', '*.launch.py')))
]
```

@Karsten1987 thoughts?

Karsten1987:
@piraka9011 I guess that having support for launch files should be definitely documented!

I am happy to review an additional PR to support launch files - it is outside of the scope of this PR though. Also when dealing with launch files, there should be support for it in Python as well as CMake then.

jpace121:
@Karsten1987 This should be done now. Can you kick off CI?

Thanks for all of the help!

jpace121:
One thing I'm not handling right now is if a user tries to name the python package "test". In that case the directory for the tests and the directory for the source code will conflict.

piraka9011:
@jpace121 Maybe adding a quick check [here](https://github.com/ros2/ros2cli/pull/296/files#diff-97bb991599247d807fe7cf4dad543f62R111) to raise a `ValueError`?
```python
bad_package_names = ['test',...]
if package.name in bad_package_names:
    raise ValueError('Package name cannot be {}'.format(package.name))
```

dirk-thomas:
Since the generated package wouldn't work correctly when named `test` the logic should protect the user from passing `test`. Therefore it shouldn't only be a warning but actually fail in that case.

ros2pkg/ros2pkg/verb/create.py
```diff
@@ -100,21 +103,35 @@ def main(self, *, args):
...
 
         test_dependencies = []
         if args.build_type == 'ament_cmake':
             test_dependencies = ['ament_lint_auto', 'ament_lint_common']
+        if args.build_type == 'ament_python':
+            test_dependencies = ['ament_copyright', 'ament_flake8', 'ament_pep257',
+                                 'python3-pytest']
+
+        if args.build_type == 'ament_python':
+            if args.package_name == 'test':
```
dirk-thomas: Please merge these two lines into a single one.

ros2pkg/ros2pkg/verb/create.py
```diff
@@ -100,21 +103,35 @@ def main(self, *, args):
...
+                # If the package name is 'test', there will be a conflict between
+                # the directory the source code for the package goes in and the
+                # directory the tests for the package go in.
+                return '\nAborted!\nament_python packages can not be named "test". \n' + \
```
dirk-thomas: Please avoid a leading newline as well as excessive use of newlines. I would suggest:

```
return "Aborted since 'ament_python' packages can't be named 'test'. Please choose a different package name."
```

ros2pkg/ros2pkg/resource/ament_python/test_copyright.py.em
```diff
@@ -0,0 +1,23 @@
+# Copyright 2015 Open Source Robotics Foundation, Inc.
```
emersonknapp: should we template these years / copyright owners? for all of these python source files

emersonknapp:
Other than dirk's comments and that tiny question about copyright notices, this looks great to me!

ros2pkg/ros2pkg/resource/ament_python/test_copyright.py.em
jpace121: For context, I copied the test_* files from [here](https://github.com/ros2/demos/tree/master/demo_nodes_py/test) with the only change being adding an extra `@` symbol in front of the decorators so Empy wouldn't try to expand them.

ros2pkg/ros2pkg/verb/create.py
jpace121: Fixed in 002ee04

ros2pkg/ros2pkg/verb/create.py
jpace121: Fixed in 002ee40.

ros2pkg/ros2pkg/verb/create.py
dirk-thomas: Please keep the suggested single quotes around `ament_package` as well as `test` (rather than none or double quotes).

ros2pkg/ros2pkg/resource/ament_python/test_copyright.py.em
dirk-thomas: These test templates can stay as is since they are copied from somewhere with that year and copyright name. For other Python modules being generated the current year and the actual copyright holder (calling the generation script) would be great.

ros2pkg/ros2pkg/verb/create.py
jpace121: Sorry about that. Should be really fixed in 4449c48

ros2pkg/ros2pkg/resource/ament_python/test_copyright.py.em
jpace121: @dirk-thomas The current generated files are fairly barren, matching the generated files for C++. Do you think it's worth a separate PR to add this to the C++ generated files as well?

emersonknapp: It's definitely outside the scope of this PR, and I think we can expect in this tooling for the user to pick their own license. Not every user is going to license their library as Apache 2, so I don't think we need to add any licensing header to the template files for user code in this PR.

jpace121:
@dirk-thomas Would you be willing to take another look at this?

dirk-thomas:
> Would you be willing to take another look at this?

:+1: The patch looks fine to me - I haven't tried to use it though.

Karsten1987:
Alright. Going to merge this.

@jpace121 thanks for iterating over this! Appreciated.

sybohy:
@Karsten1987 do you know if this PR made it into [Dashing Diademata - Patch Release 4](https://github.com/ros2/ros2/releases/tag/release-dashing-20191018)? The git trail seems to indicate so but when I use this release locally, and run `ros2 pkg create --build-type ament_python ...`, I still get the following: 

```
ros2 pkg create: error: argument --build-type: invalid choice: 'amend_python' (choose from 'cmake', 'ament_cmake')
```

dirk-thomas:
@sybohy please read the wire message again:

```
invalid choice: 'amend_python' (choose from 'cmake', 'ament_cmake')
```

And now the spelling.

dirk-thomas:
It seems to not be included in the Dashing patch release 4 (even though it was released before that date). It should be included in the next patch release.

PR #300, opened 2019-07-17
Add service call prototype completer

artivis:
Add a service request prototype completer to `ros2 service call`.  
E.g.
```terminal
$ ros2 service call /add_two_ints example_interfaces/srv/AddTwoInts [tab][tab]
a:\ 0\^Jb:\ 0\  -r              --rate
```
Not very pretty but one can at least figure out the first few characters of the request body.
Then,
```terminal
$ ros2 service call /add_two_ints example_interfaces/srv/AddTwoInts "a [tab]
```
resolves to,
```terminal
ros2 service call /add_two_ints example_interfaces/srv/AddTwoInts "a: 0
b: 0"
```

This follows the work initiated in https://github.com/ros2/ros2cli/pull/298.

Signed-off-by: artivis <jeremie.deray@canonical.com>

PR #299, opened 2019-07-17
Add topic pub prototype completer

artivis:
Add a message prototype completer to `ros2 topic pub`.  
E.g.
```terminal
$ ros2 topic pub /chatter geometry_msgs/msg/Pose [tab][tab]
-1
-n
--node-name
--once
-p
position:^J  x: 0.0^J  y: 0.0^J  z: 0.0^Jorientation:^J  x: 0.0^J  y: 0.0^J  z: 0.0^J  w: 0.0
--print
--qos-durability
--qos-profile
--qos-reliability
-r
--rate
```
Not very pretty but one can at least figure out the first few characters of the message body.
Then,
```terminal
$ ros2 topic pub /chatter geometry_msgs/msg/Pose "p [tab]
```
resolves to,
```terminal
ros2 topic pub /chatter geometry_msgs/msg/Pose "position:
  x: 0.0
  y: 0.0
  z: 0.0
orientation:
  x: 0.0
  y: 0.0
  z: 0.0
  w: 0.0"
```

This follows the work initiated in https://github.com/ros2/ros2cli/pull/298.

Signed-off-by: artivis <jeremie.deray@canonical.com>

PR #301, opened 2019-07-17
Add action send_goal prototype completer

artivis:
Add an action goal prototype completer to `ros2 action send_goal`.  
E.g.
```terminal
$ ros2 action send_goal /fibonacci action_tutorials/action/Fibonacci [tab][tab]
-f          --feedback  order:\ 0\ 
```
Not very pretty but one can at least figure out the first few characters of the goal body.
Then,
```terminal
$ ros2 action send_goal /fibonacci action_tutorials/action/Fibonacci "o [tab]
```
resolves to,
```terminal
$ ros2 action send_goal /fibonacci action_tutorials/action/Fibonacci "order: 0"
```

This follows the work initiated in https://github.com/ros2/ros2cli/pull/298.

Signed-off-by: artivis <jeremie.deray@canonical.com>

### DIFF
--- a/ros2action/ros2action/api/__init__.py
+++ b/ros2action/ros2action/api/__init__.py
@@ -22,6 +22,8 @@ import rclpy.action
 from rclpy.expand_topic_name import expand_topic_name
 from rclpy.validate_full_topic_name import validate_full_topic_name
 from ros2cli.node.direct import DirectNode
+from rosidl_runtime_py.convert import message_to_yaml
+from rosidl_runtime_py.utilities import get_action
 
 
 def _is_action_status_topic(topic_name, action_name):
@@ -139,3 +141,14 @@ class ActionTypeCompleter:
                 if n == action_name:
                     return t
         return []
+
+
+class ActionGoalPrototypeCompleter:
+    """Callable returning an action goal prototype."""
+
+    def __init__(self, *, action_type_key=None):
+        self.action_type_key = action_type_key
+
+    def __call__(self, prefix, parsed_args, **kwargs):
+        action = get_action(getattr(parsed_args, self.action_type_key))
+        return [message_to_yaml(action.Goal())]
--- a/ros2action/ros2action/verb/send_goal.py
+++ b/ros2action/ros2action/verb/send_goal.py
@@ -18,6 +18,7 @@ from action_msgs.msg import GoalStatus
 import rclpy
 from rclpy.action import ActionClient
 from ros2action.api import action_name_completer
+from ros2action.api import ActionGoalPrototypeCompleter
 from ros2action.api import ActionTypeCompleter
 from ros2action.verb import VerbExtension
 from ros2cli.node import NODE_NAME_PREFIX
@@ -39,9 +40,10 @@ class SendGoalVerb(VerbExtension):
             'action_type',
             help="Type of the ROS action (e.g. 'example_interfaces/action/Fibonacci')")
         arg.completer = ActionTypeCompleter(action_name_key='action_name')
-        parser.add_argument(
+        arg = parser.add_argument(
             'goal',
             help="Goal request values in YAML format (e.g. '{order: 10}')")
+        arg.completer = ActionGoalPrototypeCompleter(action_type_key='action_type')
         parser.add_argument(
             '-f', '--feedback', action='store_true',
             help='Echo feedback messages for the goal')


ros2action/ros2action/api/__init__.py
```diff
@@ -139,3 +142,23 @@ def __call__(self, prefix, parsed_args, **kwargs):
...
+    def __init__(self, *, action_type_key=None):
+        self.action_type_key = action_type_key
+
+    def __call__(self, prefix, parsed_args, **kwargs):
+        full_action_type = getattr(parsed_args, self.action_type_key)
+        try:
+            parts = full_action_type.split('/')
+            package_name = parts[0]
+            interface_type = parts[1]
+            action_type = parts[-1]
```
ivanpauno: `action_tutorials/Fibonacci` and `action_tutorials/action/Fibonacci` should be both supported (as this is `ros2 action` command, in the first case `action` is implicit).
Prefer checking for `len(parts)` than using a `try` `except` block.

artivis: Updated this chunk on the model of  [`send_goal`](https://github.com/ros2/ros2cli/blob/master/ros2action/ros2action/verb/send_goal.py#L84).

artivis:
@ivanpauno I'll wait on your feedback and final approval of this PR before updating #299 & #300.

ros2action/ros2action/api/__init__.py
```diff
@@ -139,3 +142,29 @@ def __call__(self, prefix, parsed_args, **kwargs):
...
+                parts = [parts[0], 'action', parts[1]]
+            if not all(parts):
+                raise ValueError()
+            package_name = parts[0]
+            interface_type = parts[1]
+            action_type = parts[-1]
+        except ValueError:
+            raise RuntimeError('The passed action type is invalid')
+        action = import_message_from_namespaced_type(
+            NamespacedType([package_name, interface_type], action_type))
```
dirk-thomas: Why is the type first split, three parts explicitly extracted and then put back together? I would be good to rewrite this part to work for any number of namespaces (e.g. more than 2).

Also as noted in various other cases where this logic is present in this repo it should be considered to not copy-n-paste another instance / variation of it her but to put a generic function in `rosidl_runtime_py` instead.

ivanpauno: In the case of an action, the second item of the namespace should always be `action`, right?
e.g.: `std_msgs/msg/String` shouldn't be accepted as a valid action.
We are not checking that in other places of this repo (see https://github.com/ros2/ros2cli/pull/304#discussion_r308410962).

artivis: Do you mean `package_name, interface_type, action_type` ? Those variables exists here mostly for readability. Notice that they are not fully put back together. I can remove them if you prefer.

How would it work with more than 2 namespaces ? What would be those other namespaces and what would the pattern be ? I could try to put a generic function together with those info.

artivis: @ivanpauno I was actually thinking of adding a check on `action`, was surprised there was none in [send_goal](https://github.com/ros2/ros2cli/blob/master/ros2action/ros2action/verb/send_goal.py#L84).
I'll do so for this chunk of code in this PR and consider another PR for `send_goal`.

dirk-thomas: `NamespacedType` takes two arguments: the first one are the namespaces in a list - the list can have any length, the second is the name of the type. To from the `split()` point of view `[-1]` becomes the second argument and `[0:-1]` the first argument.

Injecting `action` when the split length is only 2 is syntactic sugar for backward compatibility.

dirk-thomas: At the moment the second part is `action`. But there is no reason to limit the type name to that convention. Therefore I don't think it should be enforced here.

artivis: > At the moment the second part is `action`. But there is no reason to limit the type name to that convention. Therefore I don't think it should be enforced here.

@dirk-thomas I don't understand why you ':+1:' @ivanpauno comment then. From the comment linked it seems like to good idea to enforce `action` here.

artivis: @dirk-thomas I removed `package_name, interface_type, action_type` and called directly `NamespacedType(parts[0:-1], parts[-1])`.

dirk-thomas: There is a big difference between "the second item of the namespace should be `action`" and "enforcing that the second item is `action`". As I said above: atm the second item is `action` for all defined interfaces but that might not we the case in the future when we allow more flexibility / users to diverge from conventions.

artivis: There sure is, but then it makes it difficult not to accept e.g. `std_msgs/msg/String` or whatnot.

dirk-thomas: > There sure is, but then it makes it difficult not to accept e.g. `std_msgs/msg/String` or whatnot.

You could check the returned instance and check if it looks like an action or not?

artivis: Something more explicit than `return [message_to_yaml(action.Goal())]` ?  
`if not hasattr(action, 'Goal'): raise RuntimeError('Expected an action')` ?

artivis: Updated with a `hasattr(action, 'Goal')` check.

ros2action/ros2action/api/__init__.py
```diff
@@ -139,3 +142,27 @@ def __call__(self, prefix, parsed_args, **kwargs):
...
+    def __init__(self, *, action_type_key=None):
+        self.action_type_key = action_type_key
+
+    def __call__(self, prefix, parsed_args, **kwargs):
+        full_action_type = getattr(parsed_args, self.action_type_key)
+        try:
+            parts = full_action_type.split('/')
+            if len(parts) == 1:
+                raise ValueError()
+            if 'action' not in parts:
```
ivanpauno: I think this is not strictly correct.
I'm adding an utilities module to `rosidl_runtime_py` for this (https://github.com/ros2/rosidl_python/pull/78/files).
You can wait until it gets merged, or do a similar logic here (in that case, I will later use the new methods through this repo in a followup PR).

artivis: A quick try seemed not to work if `action` was not specified in the full name, thus this check/insert.

I'll rather wait for https://github.com/ros2/rosidl_python/pull/78 to be merged, it seems small enough to move fast.

artivis: maybe the following would be better (?)
```python
if parts[-1] != 'action':
```

artivis: @ivanpauno I updated this PR using the utilities from https://github.com/ros2/rosidl_python/pull/78

ivanpauno:
> * Linux: [Build Status](https://ci.ros2.org/job/ci_linux/7816/)

@artivis There's a linter failure, could you address it? Thanks!

artivis:
> @artivis There's a linter failure, could you address it? Thanks!

Sure thing !

artivis:
@ivanpauno Should be fixed now. 
I didn't realized that I had a broken version of `flake8-docstrings` .

ivanpauno:
> @ivanpauno Should be fixed now.
> I didn't realized that I had a broken version of `flake8-docstrings` .

I tested locally, and linter is passing (and auto-completion continues working well).
I'm merging this, thanks for the PR @artivis.

ivanpauno:
@artivis another minimal comment: try to avoid force pushing if it's not completely necessary, it does the review process harder.

PR #305, opened 2019-07-29
Alphasort ros2 node list output.

hidmic:
For node name order to be deterministic. Connected to #304.


hidmic:
AFAIK there are no tests in `master` for this section of the codebase. It seems to me that running CI is pointless.

hidmic:
Ok going in!

PR #306, opened 2019-07-29
Fix ros2 topic bw output units.

hidmic:
Precisely what the title says. `ros2 topic bw` output units are all bytes per second now. Connected to #304.


hidmic:
AFAIK there are no tests for this section of the codebase, so running CI seems pointless to me.

hidmic:
Ok, going in!

PR #308, opened 2019-07-30
ros2interface fixes/changes

skucheria:
Addressing comments: https://github.com/ros2/ros2cli/pull/288#discussion_r308473106

### DIFF
--- a/ros2interface/ros2interface/verb/list.py
+++ b/ros2interface/ros2interface/verb/list.py
@@ -19,7 +19,7 @@ from ros2interface.verb import VerbExtension
 
 
 def print_messages():
-    print('Messages: ')
+    print('Messages:')
     message_types = get_all_message_types()
     for package_name in sorted(message_types):
         for message_name in sorted(message_types[package_name]):
@@ -27,7 +27,7 @@ def print_messages():
 
 
 def print_services():
-    print('Services: ')
+    print('Services:')
     service_types = get_all_service_types()
     for package_name in sorted(service_types):
         for service_name in sorted(service_types[package_name]):
@@ -35,7 +35,7 @@ def print_services():
 
 
 def print_actions():
-    print('Actions: ')
+    print('Actions:')
     action_types = get_all_action_types()
     for package_name in sorted(action_types):
         for action_name in sorted(action_types[package_name]):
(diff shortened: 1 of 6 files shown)


ros2interface/ros2interface/verb/list.py
```diff
@@ -46,24 +46,26 @@ class ListVerb(VerbExtension):
     """List all interface types available."""
 
     def add_arguments(self, parser, cli_name):
-        parser.add_argument('-m', '--only-msgs', action='count',
-                            help='Print out only the message types')
+        parser.add_argument(
+            '-m', '--only-msgs', action='count',
```
dirk-thomas: The `count` action doesn't seem to be used for counting the number of times the option is passed. Just removing it will make it a simple boolean flag.

Same below.

ros2interface/ros2interface/verb/list.py
```diff
@@ -46,24 +46,26 @@ class ListVerb(VerbExtension):
...
+            help='Print out only the message types')
 
-        parser.add_argument('-s', '--only-srvs', action='count',
-                            help='Print out only the srvs types')
+        parser.add_argument(
+            '-s', '--only-srvs', action='count',
+            help='Print out only the srvs types')
```
dirk-thomas: More similar to "message types" and "action types" would be: "service types".

ros2interface/ros2interface/verb/list.py
skucheria: I used the ```count``` action so that I'd be able to call ```ros2 interface -m/s/a``` without having another argument follow the call

Otherwise usage changes and I get this error:
```usage: ros2 interface list [-h] [-m ONLY_MSGS] [-s ONLY_SRVS] [-a ONLY_ACTIONS] ```
```ros2 interface list: error: argument -s/--only-srvs: expected one argument```

dirk-thomas: You want `action='store_true'` for that.

dirk-thomas:
After making changes to the PR please comment on the ticket - otherwise there is no notification email sent so it won't be looked at again.

skucheria:
> After making changes to the PR please comment on the ticket - otherwise there is no notification email sent so it won't be looked at again.

Should I run CI for these changes or is it good to go?

dirk-thomas:
> Should I run CI for these changes or is it good to go?

Your call. If you are very confident it passes cleanly across all platforms you can go ahead without. The safer bet would be to run CI for the affected packages to make sure it does.

skucheria:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7850/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3915/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6405/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7699/)

PR #311, opened 2019-08-08
Add distribution check and report feature to `ros2 doctor` command  

ghost:
`ros2doctor` or `ros2wtf` (alias) is a tool that checks ROS system setup and other potential issues. This PR includes ROS distribution status check, platform and ROS version report (print to terminal using `ros2 doctor -r`).


clalancette:
Personally, I don't really like the name `debug`; that implies to me that it is going to help me debug my packages (i.e. with `gdb`).  The ROS 1 equivalent of this was `roswtf`; there is some value in keeping with that nomenclature for continuity.

If we don't like `roswtf` anymore, something more like `ros2systemcheck` (though wordy) is more descriptive of what this package does.

clairewangosrf:
I see how the name `debug` can cause confusion. This package includes some features in `roswtf` but will extend to checks for pub/sub, multicast, tf etc. as well. The usage is very open-ended. While `ros2systemcheck` is fairly descriptive for the `ros2 debug setup` command, it may sound restricting for other features. 

We can discuss it during next ROS2 meeting to gather more ideas and feedback, and hopefully decide on a name that people are happy about. I'm gonna keep it as `debug` for now out of convenience in dev.

clairewangosrf:
windows [Build Status](https://ci.ros2.org/job/ci_windows/7700/)

OSX [Build Status](https://ci.ros2.org/job/ci_osx/6404/)

claireyywang:
Linux [Build Status](https://ci.ros2.org/job/ci_linux/7860/)
OSX [Build Status](https://ci.ros2.org/job/ci_osx/6414/)
Windows [Build Status](https://ci.ros2.org/job/ci_windows/7721/)

ros2debug/ros2debug/api/__init__.py
```diff
@@ -0,0 +1,142 @@
...
+                                   'release': ['14.0.0', '14.5.0', '15.0.0',
+                                               '15.6.0', '16.0.0', '16.5.0',
+                                               '16.6.0', '17.0.0', '17.5.0',
+                                               '17.6.0', '17.7.0'],
+                                   'processor': ''},
+                        'windows': {'version': '10',
+                                    'release': [],
+                                    'processor': 'AMD64'}
+                        }
+            }
```
dirk-thomas: Embedding this kind of information in the code requires it to be constantly updated. I am not sure that is somthing we want to do. Some information (like the status of a ROS distribution) can be queries through `rosdistro` - e.g. https://github.com/ros/rosdistro/blob/87bd1a05b0294b55554d9e339e178e166227c7ee/index-v4.yaml#L29. Atm we only store the Debian platforms (https://github.com/ros/rosdistro/blob/87bd1a05b0294b55554d9e339e178e166227c7ee/dashing/distribution.yaml#L5) but maybe it would be a good idea to enumerate the other platforms too to make them machine readable.

claireyywang: I will take a look and change that part.

claireyywang:
Linux [Build Status](https://ci.ros2.org/job/ci_linux/7879/)
OSX [Build Status](https://ci.ros2.org/job/ci_osx/6428/)
Windows [Build Status](https://ci.ros2.org/job/ci_windows/7742/)

claireyywang:
Linux [Build Status](https://ci.ros2.org/job/ci_linux/7882/)
OSX [Build Status](https://ci.ros2.org/job/ci_osx/)
Windows [Build Status](https://ci.ros2.org/job/ci_windows/7751/)

claireyywang:
Read for review

ros2doctor/ros2doctor/api/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+
+import rosdistro
+
+
+def print_platform_info():
+    """Print out platform information."""
+    # platform info
+    print('System Information')
+    print('system   : ', platform.system())
+    print('Linux distribution   : ', platform.dist())
```
sloretz: The documentation says `platform.dist()` [will be removed in python 3.8](https://docs.python.org/3/library/platform.html#platform.dist). I recommend removing this check for now

ros2doctor/ros2doctor/api/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+    print('compiler     : ', platform.python_compiler())
+    print('build        : ', platform.python_build())
+    print('\n')
+
+
+def print_ros2_reqs():
+    """Print out ROS2 distribution info using `rosdistro`."""
+    distro_name = os.environ.get('ROS_DISTRO').lower()
+    u = rosdistro.get_index_url()
+    i = rosdistro.get_index(u)
```
sloretz: Does this raise any exceptions if it can't access the rosdistro index? If so, maybe this should have a fallback where it says it was unable to get the distribution information.

ros2doctor/ros2doctor/api/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+    distro_info = i.distributions.get(distro_name)
```
sloretz: What happens if the `distro_name` is not  in the list of known distributions?

ros2doctor/ros2doctor/api/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+    distro_data = rosdistro.get_distribution(i, distro_name).get_data()
+
+    print('ROS Information')
+    print('distribution name    : ', distro_name)
+    print('distribution type    : ', distro_info.get('distribution_type'))
+    print('distribution status  : ', distro_info.get('distribution_status'))
```
sloretz: It's possible someone set an environment variable `ROSDISTRO_INDEX_URL` to the non-v4 version of the index file, which doesn't have the type or status attributes. I think `rosdistro` will use that URL instead. What do these lines return if this check in `ros2 doctor` is invoked with?

```
ROSDISTRO_INDEX_URL https://raw.githubusercontent.com/ros/rosdistro/master/index.yaml ros2 doctor
```

ros2doctor/ros2doctor/api/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+    print('release platforms    : ', distro_data.get('release_platforms'))
+    print('\n')
+
+
+def check_platform():
+    """Check platform information against ROS2 requirements."""
+    distro_name = os.environ.get('ROS_DISTRO').lower()
```
sloretz: Same comment about handling the case where `ROS_DISTRO` is unset.

ros2doctor/ros2doctor/api/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+            To get more stable features,\
+                Download a stable version at\
+                    https://index.ros.org/doc/ros2/Installation/')
+    elif distro_info.get('distribution_status') == 'end-of-life':
+        print('WARNING: Distribution is no longer supported or deprecated.\
+            To get the latest features,\
+                Download the latest version at\
+                    https://index.ros.org/doc/ros2/Installation/')
+    else:
+        pass
```
sloretz: nit, the `else` statement is unnecessary

ros2doctor/ros2doctor/api/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+    """Check platform information against ROS2 requirements."""
+    distro_name = os.environ.get('ROS_DISTRO').lower()
+    u = rosdistro.get_index_url()
+    i = rosdistro.get_index(u)
+    distro_info = i.distributions.get(distro_name)
+    distro_data = rosdistro.get_distribution(i, distro_name).get_data()
+
+    # check distro status
+    if distro_info.get('distribution_status') == 'prerelease':
+        print('WARNING: Distribution is not fully supported or tested.\
```
sloretz: Recommend `sys.stderr.write()` instead of `print()` for warnings.

ros2doctor/ros2doctor/api/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+    # check system platform and distro release platforms
+    supported_platforms = distro_data.get('release_platforms')
+    if platform.system() == 'Linux':
+        releases = supported_platforms.get(platform.dist()[0].lower())
+        if not releases or platform.dist()[2].lower() not in releases:
+            print('WARNING: Current system platform is not supported\
+                by this ROS distribution.\
+                    User `ros2 doctor platform -r` to check report for detail.')
+        else:
+            pass
```
sloretz: Nit, unnecessary `else`

ros2doctor/ros2doctor/api/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+                Download the latest version at\
+                    https://index.ros.org/doc/ros2/Installation/')
+    else:
+        pass
+
+    # check system platform and distro release platforms
+    supported_platforms = distro_data.get('release_platforms')
+    if platform.system() == 'Linux':
+        releases = supported_platforms.get(platform.dist()[0].lower())
+        if not releases or platform.dist()[2].lower() not in releases:
```
sloretz: same comment about `dist()` being removed in Python 3.8

ros2doctor/ros2doctor/api/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+            print('WARNING: Current system platform is not supported\
+                by this ROS distribution.\
+                    User `ros2 doctor platform -r` to check report for detail.')
+        else:
+            pass
+    else:
+        print('WARNING:\
```
sloretz: I would recommend not printing this. If a check can't be done on a platform, then it should be silently skipped since there's nothing that can be done by the user to fix it.

ros2doctor/ros2doctor/command/doctor.py
```diff
@@ -0,0 +1,39 @@
...
+        self._subparser = parser
+        verb_extensions = get_verb_extensions('ros2doctor.verb')
+        add_subparsers(
+            parser, cli_name, '_verb', verb_extensions, required=False)
+
+    def main(self, *, parser, args):
+        if not hasattr(args, '_verb'):
+            # in case no verb was passed
+            self._subparser.print_help()
+            return 0
```
sloretz: `roswtf` when invoked on its own would run all checks. I recommend keeping the same behavior with `ros2 doctor` running every check by default. If someone calls `roswtf`, it's because they don't know what the problem is, and the the tool tells them what looks broken.

ros2doctor/ros2doctor/verb/platform.py
```diff
@@ -0,0 +1,40 @@
...
+            help='Display platform information and ROS 2 requirements'
+        )
+
+    def main(self, *, args):
+        """Conduct platform checks on machine against ROS2 distro reqs."""
+        if args.report:
+            print_platform_info()
+            print_ros2_reqs()
+            print('To check platform requirements use `ros2 doctor platform`.')
+        else:
```
sloretz: It looks like this function is doing two things: reporting info or checking status. It seems like this could be repeated in many checks. Maybe split this into two APIs?

Maybe `--report` could be an option on `ros2 doctor`? If `ros2 doctor --report` then it would call all the extensions that report system info, otherwise it would call all the extensions that check for system issues.

ros2doctor/ros2doctor/verb/platform.py
```diff
@@ -0,0 +1,40 @@
...
+            check_platform()
+            print('Platform check completed!')
+            print('\n')
```
sloretz: Nit, could do one call `print('Platform check completed!\n')`

ros2doctor/package.xml
```diff
@@ -0,0 +1,24 @@
...
+  <name>ros2doctor</name>
+  <version>0.0.0</version>
+  <description>A command line tool to check potential issues in ROS2 system</description>
+  <maintainer email="clairewang@openrobotics.org">Claire Wang</maintainer>
+  <license>Apache License 2.0</license>
+
+  <depend>ros2cli</depend>
+
+  <exec_depend>python3-numpy</exec_depend>
+  <exec_depend>python3-rosdistro</exec_depend>
```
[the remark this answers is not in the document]
dirk-thomas: This is not going to work. `python3-rosdistro` will conflict with `python-rosdistro` which would result in uninstalling ROS 1 stuff when installing this Debian package.

Instead this package should depend on `python3-rosdistro-modules` only (similar to `python3-catkin-pkg-modules`). This will still require a new `rosdep` key (similar to this one: https://github.com/ros/rosdistro/blob/0ea30fed0bcb564b9e4d5b2d2efd8783e8e3afa9/rosdep/python.yaml#L4913).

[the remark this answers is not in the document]
claireyywang: Does something like this look right? 
`python3-rosdistro-modules:`
   `debian: [python3-rosdistro-modules]`
   `fedora: [python3-rosdistro]`
   `gentoo: [dev-python/rosdistro]`
   `openembedded: [python3-rosdistro@meta-ros]`
   `rhel: ['python%{python3_pkgversion}-rosdistro']`
   `ubuntu: [python3-rosdistro-modules]`
I'm gonna add this to `python.yaml` in my forked repo, test then create a pull request.

ros2doctor/ros2doctor/api/__init__.py
claireyywang: If I use `platform.platform()` instead, it will give me this output ''Linux-5.0.0-23-generic-x86_64-with-Ubuntu-18.04-bionic'', which contains the same information as platform.dist() with additional parsing. Do you think parsing the string to keep this check is a worthy effort?

sloretz: I don't think parsing is worth the effort. The [docs say it's "human readable"](https://docs.python.org/3.7/library/platform.html#platform.platform), which sounds like it could change at any time. It might be useful to output all of `platform.platform()` anyways with a different label.

ros2doctor/ros2doctor/api/__init__.py
claireyywang: ROS Information
distribution name    :  dashing
distribution type    :  None
distribution status  :  None
release platforms    :  {'ubuntu': ['bionic']}

In this case,  status check is not invoked. I could add a warning regarding the two missing values if you think setting URL to non-v4 version presents significance in checks; otherwise, it seems ok to me since both distro_name and platform are still there.

ros2doctor/ros2doctor/command/doctor.py
claireyywang: I'm actually not sure how to add alias to a command. In this case I only added `wtf` to `entry_points` in `setup.py`, and that resulted in behaviours I didn't expect. But yes, `ros2 wtf` is supposed to behave the same way as `ros2 doctor`. Do you know a cli command that also has alias set up? Maybe I could use that repo as reference.

ros2doctor/ros2doctor/api/__init__.py
sloretz: The output seems ok to me too. I was mostly wondering if `distro_info.get()` would raise an exception.

ros2doctor/ros2doctor/command/doctor.py
sloretz: I was referring to the ros1 tool `roswtf`; I wasn't suggesting adding an alias. Rephrasing:

Does `ros2 doctor` run all checks when called without any arguments, just like the ros 1 tool `roswtf` does?

ros2doctor/ros2doctor/api/__init__.py
claireyywang: If I used `distro_info['distribution_type']`, it would raise an exception. `.get()` just returns None if there's no such key and doesn't break anything, which is nice :)

ros2doctor/package.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>ros2doctor</name>
+  <version>0.7.4</version>
+  <description>A command line tool to check potential issues in ROS2 system</description>
```
sloretz: nit `in ROS2 system` -> `in a ROS 2 system`

ros2doctor/ros2doctor/api/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+# limitations under the License.
+
+from ros2doctor.api.platform import check_platform
+from ros2doctor.api.platform import print_platform_info
+from ros2doctor.api.platform import print_ros2_reqs
+
+
+def run_checks():
+    """Run all checks when `ros2 doctor/wtf` is called."""
+    check_platform()
```
sloretz: I think this function should use `iter_entry_points()` to get all checks rather than hard coding a list of them. That gives downstream users a way to add new checks specific to their packages without needing to change and make a new release of `ros2doctor`.

ros2doctor/ros2doctor/api/platform.py
```diff
@@ -0,0 +1,91 @@
...
+    print('ROS Information')
+    print('distribution name    : ', distro_name)
+    print('distribution type    : ', distro_info.get('distribution_type'))
+    print('distribution status  : ', distro_info.get('distribution_status'))
+    print('release platforms    : ', distro_data.get('release_platforms'))
+
+
+def check_platform():
+    """Check platform information against ROS2 requirements."""
+    if not check_platform_helper():
```
sloretz: Nit, could store result and check if it's `None` rather than calling the function twice.

ros2doctor/ros2doctor/api/platform.py
```diff
@@ -0,0 +1,91 @@
...
+        sys.stderr.write('WARNING: Distribution is not fully supported or tested.\
+            To get more stable features,\
+                Download a stable version at\
+                    https://index.ros.org/doc/ros2/Installation/')
+    elif distro_info.get('distribution_status') == 'end-of-life':
+        sys.stderr.write('WARNING: Distribution is no longer supported or deprecated.\
+            To get the latest features,\
+                Download the latest version at\
+                    https://index.ros.org/doc/ros2/Installation/')
+    print('Platform checks completed.')
```
sloretz: I recommend not printing anything in a specific check function if no issues are found.

sloretz: Another recommendation, this check could return a value indicating if discovered any issues. `ros2 doctor` could use that result to conditionally output a summary of how many checks failed.

ros2doctor/setup.py
```diff
@@ -0,0 +1,34 @@
...
+    ],
+    description='The doctor command for ROS 2 command line tools',
+    long_description="""\
+    The package provides a cli tool to check potential issues in ROS2 system""",
+    license='Apache License, Version 2.0',
+    tests_require=['pytest'],
+    entry_points={
+        'ros2cli.command': [
+            'doctor = ros2doctor.command.doctor:DoctorCommand',
+            'wtf = ros2doctor.command.wtf:WtfCommand',
```
sloretz: Does `wtf = ros2doctor.command.doctor:DoctorCommand` work? It looks like the code is identical.

If that doesn't work, can code duplication be avoided by adding a global in `doctor.py` called `WtfCommand`?

```python3
WtfCommand = DoctorCommand
```

then

```
wtf = ros2doctor.command.doctor:WtfCommand
```

claireyywang: `ros2 doctor` stops working after adding global. Only `wtf` works. Seems to need duplicate command classes to make both work.

sloretz: What about a subclass?

```python3
class WtfCommand(DoctorCommand):
    pass
```

ros2doctor/ros2doctor/api/__init__.py
claireyywang: Tried 
```suggestion
    for check in iter_entry_points('ros2doctor.api'):
        check.load()
```
load() doesn't seem to be invoking function call. Not sure how to call function within script after getting the iterator.

sloretz: I suspect `load()` returns the method, so `check.load()()` would call it with no arguments.

ros2doctor/ros2doctor/api/platform.py
```diff
@@ -0,0 +1,95 @@
...
+
+import rosdistro
+
+
+def print_platform_info():
+    """Print out platform information."""
+    # platform info
+    print('System Information')
+    print('system   : ', platform.system())
+    print('Platform Info   : ', platform.platform())
```
sloretz: nit, the spacing looks a little strange to me since the `:` don't line up. I recommend either removing the spaces before the `:` or making sure they're all aligned.

Something like 
```
System Information
system:  Linux
Platform Info:  Linux-4.15.0-55-generic-x86_64-with-Ubuntu-18.04-bionic
Mac OS version:  ('', ('', '', ''), '')
release:  4.15.0-55-generic
processor:  x86_64
Python
version:  3.6.8
compiler:  GCC 8.0.1 20180414 (experimental) [trunk revision 259383
build:  ('default', 'Jan 14 2019 11:02:34')
ROS Information
distribution name:  dashing
distribution type:  ros2
distribution status:  active
release platforms:  {'ubuntu': ['bionic']}
```

or
```
System Information
system               :  Linux
Platform Info        :  Linux-4.15.0-55-generic-x86_64-with-Ubuntu-18.04-bionic
Mac OS version       :  ('', ('', '', ''), '')
release              :  4.15.0-55-generic
processor            :  x86_64
Python
version              :  3.6.8
compiler             :  GCC 8.0.1 20180414 (experimental) [trunk revision 259383
build                :  ('default', 'Jan 14 2019 11:02:34')
ROS Information
distribution name    :  dashing
distribution type    :  ros2
distribution status  :  active
release platforms    :  {'ubuntu': ['bionic']}
```

claireyywang:
Linux [Build Status](https://ci.ros2.org/job/ci_linux/7926/)
OSX [Build Status](https://ci.ros2.org/job/ci_osx/6453/)
Windows [Build Status](https://ci.ros2.org/job/ci_windows/7777/)

ros2doctor/ros2doctor/api/platform.py
```diff
@@ -0,0 +1,98 @@
...
+    if platform_name == 'Darwin':
+        print('Mac OS version       : ', platform.mac_ver())
+    print('release              : ', platform.release())
+    print('processor            : ', platform.processor())
+
+    # python info
+    print('PYTHON INFORMATION')
+    print('version              : ', platform.python_version())
+    print('compiler             : ', platform.python_compiler())
+    print('build                : ', platform.python_build())
```
sloretz: The `ros2 doctor --report` output looks great!

```
$ ros2 doctor --report
PLATFORM INFORMATION
system               :  Linux
Platform Info        :  Linux-4.15.0-55-generic-x86_64-with-Ubuntu-18.04-bionic
release              :  4.15.0-55-generic
processor            :  x86_64
PYTHON INFORMATION
version              :  3.6.8
compiler             :  GCC 8.0.1 20180414 (experimental) [trunk revision 259383
build                :  ('default', 'Jan 14 2019 11:02:34')
ROS INFORMATION
distribution name    :  dashing
distribution type    :  ros2
distribution status  :  active
release platforms    :  {'ubuntu': ['bionic']}
```

People that add report entry points will want to match the indentation to keep a consistent look. They'll have to make their reports pad spaces out to 21 characters before a `:` and then two spaces after the `:` before the content. If we want to make that longer in the future it might be a minor burden to update the spacing everywhere.

It might be a good idea to separate the concern of presenting data from getting the data. If the report entry points returned their data in a common format, then `generate_report()` could be made responsible for figuring out how to display it.

Design-wise I think it's fine as it is for now. This is just something to think about as an enhancement.

claireyywang: Good point! Added to my features list.

Issue #313, opened 2019-08-09
[test_ros2cli] Action CLI test failures on Windows


jacobperron:
## Bug report

**Required Info:**

- Operating System:
  - Windows 10
- Installation type:
  - source
- Version or commit hash:
  - master

Follow-up from #310 

E.g. https://ci.ros2.org/job/ci_windows/7715/

ivanpauno:
I think something like https://github.com/ros2/ros2cli/blob/0ce8efa618c31a82f4f49f152a7347de9d234b04/ros2run/ros2run/api/__init__.py#L37-L43 is missing in https://github.com/ros2/launch_ros/blob/7e4f43156c729839cc8e6dd986f32bcac169a25b/launch_ros/launch_ros/substitutions/executable_in_package.py#L77-L82.

Probably, an utility providing that functionality could be added to `ament_index_python`.

ivanpauno:
> I think something like
> 
> https://github.com/ros2/ros2cli/blob/0ce8efa618c31a82f4f49f152a7347de9d234b04/ros2run/ros2run/api/__init__.py#L37-L43
> 
> is missing in https://github.com/ros2/launch_ros/blob/7e4f43156c729839cc8e6dd986f32bcac169a25b/launch_ros/launch_ros/substitutions/executable_in_package.py#L77-L82.
> Probably, an utility providing that functionality could be added to `ament_index_python`.

I'm wrong. `Node` action was handling this correctly.

PR #315, opened 2019-08-12
Change action_tutorials_py for action_tutorials_cpp

ivanpauno:
Fixes https://github.com/ros2/ros2cli/issues/313.

For a reason I don't understand, after we split `action_tutorials` in `action_tutorials_py` and `action_tutorials_interfaces`, launch_testing can't kill `fibonacci_action_server` and the test fails (before we were installing a script and running it with `python ...`. Now we are installing it using console scripts entry points, that already installs an executable).
I switched to `action_tutorials_cpp` executable and it works well now.

I will dig in the real problem later, this is just for having green CI.

### DIFF
--- a/test_ros2cli/package.xml
+++ b/test_ros2cli/package.xml
@@ -10,7 +10,7 @@
   <license>Apache License 2.0</license>
 
   <test_depend>action_tutorials_interfaces</test_depend>
-  <test_depend>action_tutorials_py</test_depend>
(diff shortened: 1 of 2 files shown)


ivanpauno:
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/7731/)

ivanpauno:
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/7733/)

PR #319, opened 2019-08-21
Add network configuration check and report to ros2doctor

claireyywang:
Most changes happen in `ros2doctor/api/network.py`. Network configuration is checked and printed using `ifcfg` python-pip package. `ros2doctor/api/format.py` is created to format the report. Current padding is 20 chars, but will be changed in the future to dynamically allocate padding based on string length. I have also modified print statements in `ros2doctor/api/platform.py` to reflect new report format.

### DIFF
--- a/ros2doctor/package.xml
+++ b/ros2doctor/package.xml
@@ -9,7 +9,7 @@
 
   <depend>ros2cli</depend>
 
-  <exec_depend>python3-ifcfg</exec_depend>
+  <exec_depend>ifcfg-pip</exec_depend>
   <exec_depend>python3-rosdistro-modules</exec_depend>
   
   <test_depend>ament_copyright</test_depend>
--- a/ros2doctor/package.xml
+++ b/ros2doctor/package.xml
@@ -9,7 +9,7 @@
 
   <depend>ros2cli</depend>
 
-  <exec_depend>python3-numpy</exec_depend>
+  <exec_depend>python3-ifcfg</exec_depend>
   <exec_depend>python3-rosdistro-modules</exec_depend>
   
   <test_depend>ament_copyright</test_depend>
--- a/ros2doctor/ros2doctor/api/format.py
+++ b/ros2doctor/ros2doctor/api/format.py
@@ -1,0 +1,20 @@
+# Copyright 2019 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+def print_term(k, v):
+    """Print term only if it exists."""
+    if v:
+        # TODO: 20 padding needs to be dynamically set
+        print('{:20}: {}'.format(k,v))
--- a/ros2doctor/ros2doctor/api/format.py
+++ b/ros2doctor/ros2doctor/api/format.py
@@ -17,4 +17,4 @@ def print_term(k, v):
     """Print term only if it exists."""
     if v:
         # TODO: 20 padding needs to be dynamically set
-        print('{:20}: {}'.format(k,v))
+        print('{:20}: {}'.format(k, v))
--- a/ros2doctor/ros2doctor/api/network.py
+++ b/ros2doctor/ros2doctor/api/network.py
@@ -90,4 +90,6 @@ def check_network():
 
 def print_network():
     """Print all system and ROS network information."""
+    local_addrs = get_local_addresses()
+    print('Local addresses: ', *local_addrs)
     pass
--- a/ros2doctor/ros2doctor/api/network.py
+++ b/ros2doctor/ros2doctor/api/network.py
@@ -1,0 +1,91 @@
+# Copyright 2019 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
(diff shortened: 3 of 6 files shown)


claireyywang:
Linux [Build Status](https://ci.ros2.org/job/ci_linux/7961/)
Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64)
OSX [Build Status](https://ci.ros2.org/job/ci_osx)
Windows [Build Status](https://ci.ros2.org/job/ci_windows)

ros2doctor/package.xml
```diff
@@ -9,7 +9,7 @@
 
   <depend>ros2cli</depend>
 
-  <exec_depend>python3-numpy</exec_depend>
+  <exec_depend>ifcfg-pip</exec_depend>
```
sloretz: Is there a Debian release of this package? Packages released into a ROS distro (like the packages in this repo) can't use packages from `PyPi` because `.deb`s can only depend on other Debian packages.

If there isn't a Debian release, an alternative is to release our own `deb`, which means creating a vendor package for `ifcfg`.

claireyywang: vendor package for `ifcfg` at https://github.com/ros2/ifcfg_vendor

sloretz: need to change this dependency to `ifcfg_vendor`.

ros2doctor/ros2doctor/api/format.py
```diff
@@ -0,0 +1,20 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+def print_term(k, v):
+    """Print term only if it exists."""
+    if v:
+        # TODO: 20 padding needs to be dynamically set
```
sloretz: nit, add parenthesis with your github username. `TODO(claireyywang)`. The name in the parenthesis is the person who can answer questions about what needs to be done. Most of the time it's the person who created the comment.

ros2doctor/ros2doctor/api/network.py
```diff
@@ -0,0 +1,63 @@
...
+import sys
+
+import ifcfg
+
+from ros2doctor.api.format import print_term
+
+
+def _is_unix_like_platform():
+    """Return True if conforms to UNIX/POSIX-style APIs."""
+    return platform.system() in ['Linux', 'FreeBSD', 'Darwin']
```
sloretz: Recommend `'posix' == os.name` instead

ros2doctor/ros2doctor/api/network.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def check_sys_ips():
```
sloretz: This function name doesn't seem very descriptive. It's a little more confusing in the context of `ros2doctor` since it's not one of the `ros2doctor.checks`.  I don't have an alternative though.

ros2doctor/ros2doctor/api/network.py
```diff
@@ -0,0 +1,63 @@
...
+    """Check if loopback and multicast IP addresses are found."""
+    has_loopback, has_others, has_multicast = False, False, False
+    for name, iface in ifcfg.interfaces().items():
+        flags = iface.get('flags')
+        if 'LOOPBACK' in flags:
+            has_loopback = True
+        else:
+            has_others = True
```
sloretz: recommend `has_non_loopback` instead

ros2doctor/ros2doctor/api/network.py
```diff
@@ -0,0 +1,63 @@
...
+    """Conduct network checks and output error/warning messages."""
+    result = False
+    has_loopback, has_others, has_multicast = check_sys_ips()
+    if not has_loopback:
+        sys.stderr.write('ERROR: No loopback IP address is found.\n')
+    if not has_others:
+        sys.stderr.write('WARNING: Only loopback IP address is found.\n')
+    if not has_multicast:
+        sys.stderr.write('WARNING: No multicast IP address is found.\n')
+    if has_loopback and has_others and has_multicast:
```
sloretz: nit, can eliminate `result` with `return has_loopback and has_others and has_multicast`

ros2doctor/ros2doctor/api/platform.py
```diff
@@ -24,37 +26,38 @@ def print_platform_info():
...
+    print_term('processor', platform.processor())
 
     # python info
     print('PYTHON INFORMATION')
-    print('version              : ', platform.python_version())
-    print('compiler             : ', platform.python_compiler())
-    print('build                : ', platform.python_build())
+    print_term('version', platform.python_version())
+    print_term('compiler', platform.python_compiler())
+    print_term('build', platform.python_build())
```
sloretz: This is definitely an improvement. I still think it would be good if the `reports` just returned information and the command was responsible for outputting it.

ros2doctor/package.xml
claireyywang: do I also need to change to ifcfg_vendor on rosdistro rosdep key?

sloretz: > do I also need to change to ifcfg_vendor on rosdistro rosdep key?

I don't think so. `ifcfg-pip` is the right name for the rosdistro key (the key isn't needed for this PR since there's a vendor package). The package.xml should be changed to depend on the new package `ifcfg_vendor`.

dirk-thomas:
Since adding a vendor package for `ifcfg` is quite some overhead and maintenance burden moving forward I would like to double check on this. What exact API / functionality is being used from that package? What other Python packages have been considered / checked if they could provide the same functionality?

ros2doctor/ros2doctor/api/network.py
claireyywang: changing to `check_network_config_helper`

claireyywang:
> Since adding a vendor package for `ifcfg` is quite some overhead and maintenance burden moving forward I would like to double check on this. What exact API / functionality is being used from that package? What other Python packages have been considered / checked if they could provide the same functionality?

`ifcfg` provides the same information as cmd `ifconfig` on terminal. It provides 'flags' that show whether an IP is used for loopback, broadcast or multicast etc. which I believe is important for ros2 checks. I also tried `netifaces` and `socket` but could not find enough information to check for multicast network interface.

ros2doctor/ros2doctor/api/platform.py
claireyywang: I was thinking about making that change as a feature in the next PR. Like you suggested last time, I'd like to add arguments like `--report_all` or `--report_failed` to let users control what they would like to see based on the check results.

claireyywang:
Linux: [Build Status](https://ci.ros2.org/job/ci_linux/)
Linux-aarch64: [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/)
OSX: [Build Status](https://ci.ros2.org/job/ci_osx)
Windows: [Build Status](https://ci.ros2.org/job/ci_windows/)

PR #324, opened 2019-08-28
Update report feature with new argument, add temp fix for ifcfg module 

claireyywang:
Import error added in `network.py` as a temporary fix for `ifcfg` module. Will remove and update with build dependency once a `deb` package is ready.

This is an enhancement to `report` feature based on previous discussion https://github.com/ros2/ros2cli/pull/319#discussion_r317217343 

This PR adds `--report_failed` argument, which allows user to only print report of failed checks. Major changes happen in `format.py` and `doctor.py` files. 

Report content is output from each check module and indexed by their module names in `generate_report()`. They are printed out by calling `format_print(modules, report)`. The function takes module names and print out their indexed report, with dynamically allocated space padding computed by `compute_padding`.

### DIFF
--- a/ros2doctor/package.xml
+++ b/ros2doctor/package.xml
@@ -9,7 +9,6 @@
 
   <depend>ros2cli</depend>
 
-  <exec_depend>ifcfg_vendor</exec_depend>
   <exec_depend>python3-rosdistro-modules</exec_depend>
   
   <test_depend>ament_copyright</test_depend>
--- a/ros2doctor/package.xml
+++ b/ros2doctor/package.xml
@@ -9,7 +9,7 @@
 
   <depend>ros2cli</depend>
 
-  <exec_depend>python3-ifcfg</exec_depend>
+  <exec_depend>ifcfg-pip</exec_depend>
   <exec_depend>python3-rosdistro-modules</exec_depend>
   
   <test_depend>ament_copyright</test_depend>
--- a/ros2doctor/package.xml
+++ b/ros2doctor/package.xml
@@ -9,7 +9,7 @@
 
   <depend>ros2cli</depend>
 
-  <exec_depend>ifcfg-pip</exec_depend>
+  <exec_depend>ifcfg_vendor</exec_depend>
   <exec_depend>python3-rosdistro-modules</exec_depend>
   
   <test_depend>ament_copyright</test_depend>
--- a/ros2doctor/package.xml
+++ b/ros2doctor/package.xml
@@ -9,7 +9,7 @@
 
   <depend>ros2cli</depend>
 
-  <exec_depend>python3-numpy</exec_depend>
+  <exec_depend>python3-ifcfg</exec_depend>
   <exec_depend>python3-rosdistro-modules</exec_depend>
   
   <test_depend>ament_copyright</test_depend>
--- a/ros2doctor/ros2doctor/api/__init__.py
+++ b/ros2doctor/ros2doctor/api/__init__.py
@@ -77,7 +77,6 @@ def run_checks() -> Tuple[Set[str], int, int]:
             result = check_class.check()
         except ValueError:
             doctor_warn('Check class from %s failed to load.' % check_entry_pt.name)
-            pass
         if result is False:
             fail += 1
             failed_cats.add(cat)
@@ -99,7 +98,6 @@ def generate_reports(*, categories=None) -> List[Report]:
             report = report_class.report()
         except ValueError:
             doctor_warn('Report class from %s failed to load.' % report_entry_pt.name)
-            pass
         if categories:
             if report_category in categories:
                 reports.append(report)
--- a/ros2doctor/ros2doctor/api/__init__.py
+++ b/ros2doctor/ros2doctor/api/__init__.py
@@ -94,13 +94,13 @@ def generate_reports(*, categories=None) -> List[Report]:
     reports = []
     for report_entry_pt in iter_entry_points('ros2doctor.report'):
         report_class = report_entry_pt.load()()
+        try:
+            report_category = report_class.category()
+            report = report_class.report()
+        except ValueError:
+            doctor_warn('Report class failed to load.')
+            pass
         if categories:
-            try:
-                report_category = report_class.category()
-                report = report_class.report()
-            except ValueError:
-                doctor_warn('Report class failed to load.')
-                pass
             if report_category in categories:
                 reports.append(report)
         else:
--- a/ros2doctor/ros2doctor/api/__init__.py
+++ b/ros2doctor/ros2doctor/api/__init__.py
@@ -12,11 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from pkg_resources import iter_entry_points
 from typing import List
 from typing import Set
 from typing import Tuple
-import warnings
+
+from pkg_resources import iter_entry_points
 
 from ros2doctor.api.format import doctor_warn
 
@@ -25,15 +25,11 @@ class DoctorCheck:
     """Abstract base class of ros2doctor check."""
 
     def category(self) -> str:
-        """
-        :return: string linking checks and reports
-        """
+        """:return: string linking checks and reports."""
         raise NotImplementedError
 
     def check(self) -> bool:
-        """
-        :return: boolean indicating result of checks
-        """
+        """:return: boolean indicating result of checks."""
         raise NotImplementedError
 
 
@@ -41,15 +37,11 @@ class DoctorReport:
     """Abstract base class of ros2doctor report."""
 
     def category(self) -> str:
-        """
-        :return: string linking checks and reports
-        """
+        """:return: string linking checks and reports."""
         raise NotImplementedError
 
     def report(self) -> 'Report':  # using str as wrapper for custom class Report
-        """
-        :return: Report object storing report content
-        """
+        """:return: Report object storing report content."""
         raise NotImplementedError
 
 
@@ -71,6 +63,7 @@ class Report:
 def run_checks() -> Tuple[Set[str], int, int]:
     """
     Run all checks and return check results.
+
     :return: 3-tuple (categories of failed checks, number of failed checks,
              total number of checks)
     """
@@ -95,6 +88,7 @@ def run_checks() -> Tuple[Set[str], int, int]:
 def generate_reports(*, categories=None) -> List[Report]:
     """
     Print all reports or reports of failed checks to terminal.
+
     :return: list of Report objects
     """
     reports = []
--- a/ros2doctor/ros2doctor/api/__init__.py
+++ b/ros2doctor/ros2doctor/api/__init__.py
@@ -41,7 +41,9 @@ class DoctorReport:
 
 class Report:
     """Contains report name and content."""
+
     __slots__ = ['name', 'items']
+
     def __init__(self, name):
         self.name = name
         self.items = []
--- a/ros2doctor/ros2doctor/api/__init__.py
+++ b/ros2doctor/ros2doctor/api/__init__.py
@@ -19,11 +19,11 @@ class DoctorCheck:
     """Abstract base class of ros2doctor check."""
 
     def category(self):
-        """Return: string linking checks and reports."""
+        """Return: String linking checks and reports."""
         raise NotImplementedError
 
     def check(self):
-        """Return: boolean indicating result of checks."""
+        """Return: Boolean indicating result of checks."""
         raise NotImplementedError
 
 
@@ -31,16 +31,16 @@ class DoctorReport:
     """Abstract base class of ros2doctor report."""
 
     def category(self):
-        """Return: a string linking checks and reports."""
+        """Return: String linking checks and reports."""
         raise NotImplementedError
 
     def report(self):
-        """Return: a Report object containing report content."""
+        """Return: Report object containing report content."""
         raise NotImplementedError
 
 
 class Report:
-    """Contains report name and content."""
+    """Stores report name and content."""
 
     __slots__ = ['name', 'items']
 
@@ -56,23 +56,45 @@ def run_checks():
     """
     Run all checks and return check results.
 
-    Return: list of tuple (string, boolean) => (category, check result)
+    Return: list of String, Int, Int
     """
-    results = []
+    failed_cats = []
+    fail = 0
+    total = 0
     for check_entry_pt in iter_entry_points('ros2doctor.checks'):
         check_class = check_entry_pt.load()()
-        results.append((check_class.category(), check_class.check()))
-    return results
+        cat = check_class.category()
+        result = check_class.check()
+        if result is False:
+            fail += 1
+            if cat not in failed_cats:
+                failed_cats.append(cat)
+        total += 1
+    return failed_cats, fail, total
 
 
-def generate_report():
+def generate_all_reports():
     """
-    Print report to terminal.
+    Print all reports to terminal.
 
-    Return: list of tuple (string, list of tuple) => (category, report items)
+    Return: list of Report objects
     """
     reports = []
     for report_entry_pt in iter_entry_points('ros2doctor.report'):
         report_class = report_entry_pt.load()()
-        reports.append((report_class.category(), report_class.report()))
+        reports.append(report_class.report())
+    return reports
+
+
+def generate_fail_reports(failed_cats):
+    """
+    Print reports of failed checks to terminal.
+
+    Return: list of Report objects
+    """
+    reports = []
+    for report_entry_pt in iter_entry_points('ros2doctor.report'):
+        report_class = report_entry_pt.load()()
+        if report_class.category() in failed_cats:
+            reports.append(report_class.report())
     return reports
--- a/ros2doctor/ros2doctor/api/__init__.py
+++ b/ros2doctor/ros2doctor/api/__init__.py
@@ -18,14 +18,15 @@ from pkg_resources import iter_entry_points
 def run_checks():
     """Run all checks when `ros2 doctor/wtf` is called."""
     all_results = []
-    failed_names = []
+    failed_checks = []
+    failed_modules = []
     for check in iter_entry_points('ros2doctor.checks'):
         result = check.load()()  # load() returns method
         all_results.append(result)
-        print(check.module_name)
         if result is False:
-            failed_names.append(check.module_name)
-    return all_results, failed_names
+            failed_checks.append(check.name)
+            failed_modules.append(check.module_name)
+    return all_results, failed_checks, failed_modules
 
 
 def generate_report():
--- a/ros2doctor/ros2doctor/api/__init__.py
+++ b/ros2doctor/ros2doctor/api/__init__.py
@@ -16,8 +16,13 @@ from pkg_resources import iter_entry_points
 
 
 class DoctorCheck:
+    """
+        Abstract based class of ros2doctor check modules.
+        category method returns a string that includes the module name.
+        check method returns a boolean value based on check result.
+    """
 
-    def target(self):
+    def category(self):
         raise NotImplementedError
 
     def check(self):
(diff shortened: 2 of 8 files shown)


claireyywang:
Linux [Build Status](https://ci.ros2.org/job/ci_linux/8026/)
Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/4036/)
OSX [Build Status](https://ci.ros2.org/job/ci_osx/6536/)
Windows [Build Status](https://ci.ros2.org/job/ci_windows/7884/)

ros2doctor/ros2doctor/api/__init__.py
```diff
@@ -18,16 +18,23 @@
...
 
 
 def generate_report():
     """Print report to terminal when `-r/--report` is attached."""
-    for report in iter_entry_points('ros2doctor.report'):
-        report.load()()  # load() returns method
+    report = {}
+    for r in iter_entry_points('ros2doctor.report'):
+        if r.module_name in report:
+            report[r.module_name].extend(r.load()())
```
sloretz: Recommend using something other than the module name since reports in the same module could check very different things. Maybe the name of the entry point?

```python3
report[r.name] = r.load()()
```


```Python3
>>> list(pr.iter_entry_points('ros2doctor.report'))[0]
EntryPoint.parse('report_network_config = ros2doctor.api.network:print_network')
>>> list(pr.iter_entry_points('ros2doctor.report'))[0].module_name
'ros2doctor.api.network'
>>> list(pr.iter_entry_points('ros2doctor.report'))[0].name
'report_network_config'
```

ros2doctor/ros2doctor/api/format.py
```diff
@@ -12,9 +12,33 @@
...
-        print('{:20}: {}'.format(k, v))
+
+def format_print(modules, report):
+    """Format print report according to module names."""
+    padding_num = compute_padding(modules, report)
+    for m in modules:
+        if m in report:
+            module_report = report.get(m)
+            for k, v in module_report:
+                if k == 'NAME':
```
sloretz: Recommend having reports return a class instance (`class Report:`?) instead of a dictionary. It's hard to document that `NAME` is a special key reports can return; where does that documentation go? A class definition with docstrings would be a spot that can be pointed to when answering future questions about how to write a report for `ros2doctor`.

ros2doctor/ros2doctor/api/format.py
```diff
@@ -12,9 +12,33 @@
...
+                    print(v)
+                else:
+                    print('{:{padding}}: {}'.format(k, v, padding=padding_num))
+        else:
+            # print warning if no report available for a particular check
+            sys.stderr.write('WARNING: No report available for this check.')
```
sloretz: Missing `\n` at the end. Alternatively use

```Python3
print('WARNING: No report available for this check.', file=sys.sterr)
```

ros2doctor/ros2doctor/api/network.py
```diff
@@ -51,10 +53,11 @@ def check_network_config():
     return has_loopback and has_non_loopback and has_multicast
 
 
-def print_network():
+def report_network():
     """Print all system and ROS network information."""
-    print('NETWORK CONFIGURATION')
+    network_info = [('NAME', 'NETWORK CONFIGURATION')]
     for name, iface in ifcfg.interfaces().items():
```
sloretz: I think this will raise `AttributeError` if the `ifcfg` import above failed.

ros2doctor/ros2doctor/command/doctor.py
```diff
@@ -23,21 +24,31 @@ class DoctorCommand(CommandExtension):
     def add_arguments(self, parser, cli_name):
         parser.add_argument(
             '--report', '-r', action='store_true',
-            help='Print out doctor report.'
+            help='Print out all report info.'
+        )
+        parser.add_argument(
+            '--report_failed', '-rf', action='store_true',
```
sloretz: nit, I think `-` instead of `_` is more consistent with options on other CLI tools, so `--report-failed`.

ros2doctor/ros2doctor/command/doctor.py
```diff
@@ -23,21 +24,31 @@ class DoctorCommand(CommandExtension):
...
+            help='Print out report info on failed checks.'
         )
 
     def main(self, *, parser, args):
+        # print report only if no failed checks requested
         if args.report:
-            generate_report()
+            report = generate_report()
+            format_print(report.keys(), report)
```
sloretz: `(report.keys(), report)` is passing the same information twice. Can this be just `(report)`?

ros2doctor/ros2doctor/command/doctor.py
```diff
@@ -23,21 +24,31 @@ class DoctorCommand(CommandExtension):
     def add_arguments(self, parser, cli_name):
         parser.add_argument(
             '--report', '-r', action='store_true',
-            help='Print out doctor report.'
+            help='Print out all report info.'
+        )
+        parser.add_argument(
+            '--report_failed', '-rf', action='store_true',
+            help='Print out report info on failed checks.'
```
sloretz: It looks like `--report` and `--report_failed` are mutually exclusive options. Recommend using an [argparse mutually exclusive group](https://docs.python.org/3/library/argparse.html#mutual-exclusion) so the user gets an error when `--report` and `--report_failed` are both passed.

ros2doctor/package.xml
```diff
@@ -8,7 +8,7 @@
   <license>Apache License 2.0</license>
 
   <depend>ros2cli</depend>
-
```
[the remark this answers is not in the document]
claireyywang: I think that's where I deleted the `ifcfg_vendor` build dependency. Good to know about convention on whitespace changes.

ros2doctor/ros2doctor/api/__init__.py
claireyywang: `module_name` is the only direct correlation between checks and reports at this point. I use it as key to report any failed check. The design goal of this part is to make check and report inside the same module serve one specific area like `platform`, `network`, `topic` etc. I can see when there could be a lot of checks inside one module, but to tailor each report around a single check would require some string manipulation and strict rules on report function names.

claireyywang:
Linux [Build Status](https://ci.ros2.org/job/ci_linux)
Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64)
OSX [Build Status](https://ci.ros2.org/job/ci_osx)
Windows [Build Status](https://ci.ros2.org/job/ci_windows)

ros2doctor/ros2doctor/api/__init__.py
```diff
@@ -15,19 +15,68 @@
 from pkg_resources import iter_entry_points
 
 
+class DoctorCheck:
+    """
+    Abstract based class of ros2doctor check modules.
```
sloretz: nit `based` -> `base`

ros2doctor/ros2doctor/api/__init__.py
```diff
@@ -15,19 +15,68 @@
...
+
+    category method returns a string that includes the module name.
+    check method returns a boolean value based on check result.
```
sloretz: Recommend making these notes docstrings on the methods themselves.

ros2doctor/ros2doctor/api/__init__.py
```diff
@@ -15,19 +15,68 @@
 from pkg_resources import iter_entry_points
 
 
+class DoctorCheck:
+    """
+    Abstract based class of ros2doctor check modules.
+
+    category method returns a string that includes the module name.
```
sloretz: This should say what the returned string is for: for linking reports and checks.

ros2doctor/ros2doctor/api/__init__.py
```diff
@@ -15,19 +15,68 @@
...
+    def category(self):
+        raise NotImplementedError
+
+    def check(self):
+        raise NotImplementedError
+
+
+class DoctorReport:
+    """
+    Abstract based class of ros2doctor report modules.
```
sloretz: nit `based` -> `base`

ros2doctor/ros2doctor/api/__init__.py
```diff
@@ -15,19 +15,68 @@
...
+    def report(self):
+        raise NotImplementedError
+
+
+class Report:
+    """Contains report name and content."""
+
+    def __init__(self, name):
+        self.name = name
+        self.items = []
```
sloretz: Recommend using `__slots__` since the number and names of the attributes are fixed.

ros2doctor/ros2doctor/api/format.py
```diff
@@ -12,9 +12,25 @@
...
-def print_term(k, v):
-    """Print term only if it exists."""
-    if v:
-        # TODO(clairewang): 20 padding needs to be dynamically set
-        print('{:20}: {}'.format(k, v))
+
+def format_print(report):
+    """Format print report content."""
+    if report is None:
+        sys.stderr.write('WARNING: No report found. Skip print...\n')
```
sloretz: When can `format_print(None)` be called? It seems like that should raise an exception instead of printing a warning.

ros2doctor/ros2doctor/api/network.py
```diff
@@ -15,19 +15,31 @@
 import os
 import sys
 
-import ifcfg
+try:
+    import ifcfg
+except ImportError:
+    sys.stderr.write('WARNING: No ifcfg module found. '
```
sloretz: Recommend using the `warnings` module https://docs.python.org/3/library/warnings.html

ros2doctor/ros2doctor/api/format.py
claireyywang: In this case, if `ifcfg` is not imported successfully, the entire network module is skipped and report is None. I guess I should have added a note saying it is a temp fix. I will add the exception after `ifcfg` debian package becomes available.

claireyywang:
Linux [Build Status](https://ci.ros2.org/job/ci_linux/8065/)
Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/4073/)
OSX [Build Status](https://ci.ros2.org/job/ci_osx/6571/)
Windows [Build Status](https://ci.ros2.org/job/ci_windows/7937/)

ros2doctor/ros2doctor/api/__init__.py
```diff
@@ -15,19 +15,75 @@
...
 
+def generate_reports(*, cats=None):
+    """
+    Print all reports or reports of failed checks to terminal.
 
-def generate_report():
-    """Print report to terminal when `-r/--report` is attached."""
-    for report in iter_entry_points('ros2doctor.report'):
-        report.load()()  # load() returns method
+    Return: list of Report objects
```
sloretz: nit, sphinx style `:return:`

ros2doctor/ros2doctor/api/__init__.py
```diff
@@ -15,19 +15,75 @@
...
         if result is False:
-            failed_names.append(check.name)
-    return all_results, failed_names
+            fail += 1
+            failed_cats.add(cat)
+        total += 1
+    return failed_cats, fail, total
+
 
+def generate_reports(*, cats=None):
```
sloretz: nit, recommend `categories` instead of `cats`.

ros2doctor/ros2doctor/api/__init__.py
```diff
@@ -15,19 +15,75 @@
...
 
-def generate_report():
-    """Print report to terminal when `-r/--report` is attached."""
-    for report in iter_entry_points('ros2doctor.report'):
-        report.load()()  # load() returns method
+    Return: list of Report objects
+    """
+    reports = []
+    for report_entry_pt in iter_entry_points('ros2doctor.report'):
+        report_class = report_entry_pt.load()()
```
sloretz: Same comment about wrapping uses of `report_class` in a `try/except` block.

ros2doctor/ros2doctor/api/format.py
```diff
@@ -12,9 +12,31 @@
...
-    """Print term only if it exists."""
-    if v:
-        # TODO(clairewang): 20 padding needs to be dynamically set
-        print('{:20}: {}'.format(k, v))
+
+def format_print(report):
+    """Format print report content."""
+    # temp fix for missing ifcfg
+    if report is None:
+        sys.stderr.write('No report found. Skip print...')
```
sloretz: nit, add `\n`

```
No report found. Skip print...ros2-dev0|sloretz@c70527c77758:~$
```

ros2doctor/ros2doctor/api/format.py
```diff
@@ -12,9 +12,31 @@
...
+        return
+
+    print('\n', report.name)
```
sloretz: The comma adds a leading space. Is that intentional?

```
 ROS 2 INFORMATION
distribution name      : eloquent
```

ros2doctor/ros2doctor/api/platform.py
```diff
@@ -14,87 +14,102 @@
...
-    print('PYTHON INFORMATION')
-    print_term('version', platform.python_version())
-    print_term('compiler', platform.python_compiler())
-    print_term('build', platform.python_build())
-    print('\n')
-
-
-def check_platform_helper():
-    """Check ROS_DISTRO related environment variables and distribution name."""
+warnings.formatwarning = warning_format
```
sloretz: Recommend using a context manager to only override this while checks and reports are being run. This will make it more robust to something else being imported that also customizes `formatwarning`.

Something like

```python3

import warnings

def warning_format(msg, cat, filename, linenum, file=None, line=None):
    return 'testtesttest %s: %s: %s: %s\n' % (filename, linenum, cat.__name__, msg)


class CustomWarningFormat:

    def __enter__(self):
        self._old_format = warnings.formatwarning
        warnings.formatwarning = warning_format

    def __exit__(self, t, v, tb):
        warnings.formatwarning = self._old_format


# Run checks/reports inside `with` statement to limit who the warning is customized for

warnings.warn('old format')

with CustomWarningFormat():
    warnings.warn('custom format')

warnings.warn('old format again')
```

Results in 

```
$ python3 warn.py 
warn.py:18: UserWarning: old format
  warnings.warn('old format')
testtesttest warn.py: 21: UserWarning: custom format
warn.py:23: UserWarning: old format again
  warnings.warn('old format again')
```

ros2doctor/ros2doctor/command/doctor.py
```diff
@@ -13,31 +13,47 @@
...
 class DoctorCommand(CommandExtension):
     """Check ROS setup and other potential issues."""
 
     def add_arguments(self, parser, cli_name):
-        parser.add_argument(
+        group = parser.add_mutually_exclusive_group(required=False)
+        group.add_argument(
             '--report', '-r', action='store_true',
-            help='Print out doctor report.'
+            help='Print all report.'
```
sloretz: nit `report` -> `reports`

ros2doctor/ros2doctor/command/doctor.py
```diff
@@ -13,31 +13,47 @@
...
         if args.report:
-            generate_report()
+            all_reports = generate_reports()
+            for report_obj in all_reports:
+                format_print(report_obj)
+            return
+
+        # `ros2 doctor`
+        failed_cats, fail, total = run_checks()
+        if fail != 0:
```
sloretz: nit, same as `if fail:`

ros2doctor/ros2doctor/api/__init__.py
```diff
@@ -15,19 +15,75 @@
...
-    """Run all checks when `ros2 doctor/wtf` is called."""
-    all_results = []
-    failed_names = []
-    for check in iter_entry_points('ros2doctor.checks'):
-        result = check.load()()  # load() returns method
-        all_results.append(result)
+    """
+    Run all checks and return check results.
+
+    Return: list of String, Int, Int
```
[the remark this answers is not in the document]
dirk-thomas: How about type annotations instead?

ros2doctor/ros2doctor/api/format.py
claireyywang: It makes the header stands out more so I left it in. I can change it if it seems weird.

sloretz: No need to change, just checking it wasn't accidental

claireyywang:
Linux [Build Status](https://ci.ros2.org/job/ci_linux/8078/)
Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/4083/)
OSX [Build Status](https://ci.ros2.org/job/ci_osx/6582/)
Windows [Build Status](https://ci.ros2.org/job/ci_windows/7953/)

claireyywang:
[Build Status](https://ci.ros2.org/job/ci_linux/8096/)

claireyywang:
Linux [Build Status](https://ci.ros2.org/job/ci_linux/8117/)

ros2doctor/ros2doctor/api/__init__.py
```diff
@@ -12,22 +12,109 @@
...
+    fail = 0
+    total = 0
+    for check_entry_pt in iter_entry_points('ros2doctor.checks'):
+        try:
+            check_class = check_entry_pt.load()
+        except ValueError:
+            doctor_warn('Check entry point %s fails to load.' % check_entry_pt.name)
+        try:
+            check_instance = check_class()
+        except ValueError:
```
sloretz: `ValueError` -> [`Exception`](https://docs.python.org/3/library/exceptions.html#Exception).  The goal is to catch unexpected exceptions, not just `ValueError`.

ros2doctor/ros2doctor/api/__init__.py
```diff
@@ -12,22 +12,109 @@
...
+            doctor_warn('Unable to instantiate check object from %s.' % check_entry_pt.name)
+        try:
+            check_category = check_instance.category()
+            result = check_instance.check()
+        except ValueError:
```
sloretz: Same comment `ValueError` -> `Exception`

ros2doctor/ros2doctor/api/__init__.py
```diff
@@ -12,22 +12,109 @@
...
+    """
+    reports = []
+    for report_entry_pt in iter_entry_points('ros2doctor.report'):
+        try:
+            report_class = report_entry_pt.load()
+        except ValueError:
+            doctor_warn('Report entry point %s fails to load.' % report_entry_pt.name)
+        try:
+            report_instance = report_class()
+        except ValueError:
```
sloretz: `Exception` here too

ros2doctor/ros2doctor/api/__init__.py
```diff
@@ -12,22 +12,109 @@
...
+            doctor_warn('Unable to instantiate report object from %s.' % report_entry_pt.name)
+        try:
+            report_category = report_instance.category()
+            report = report_instance.report()
+        except ValueError:
```
sloretz: `Exception`

ros2doctor/ros2doctor/api/__init__.py
```diff
@@ -12,22 +12,109 @@
...
+    :return: 3-tuple (categories of failed checks, number of failed checks,
+             total number of checks)
+    """
+    failed_cats = set()  # remove repeating elements
+    fail = 0
+    total = 0
+    for check_entry_pt in iter_entry_points('ros2doctor.checks'):
+        try:
+            check_class = check_entry_pt.load()
+        except ValueError:
```
sloretz: It looks like `load()` can raise `UnknownExtra` (`from pkg_resources import UnknownExtra`) or `ImportError`. I don't see `ValueError` in the docs.

https://setuptools.readthedocs.io/en/latest/pkg_resources.html#entrypoint-objects

claireyywang:
Linux [Build Status](https://ci.ros2.org/job/ci_linux/8137/)
Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/4125/)
OSX [Build Status](https://ci.ros2.org/job/ci_osx/6627/)
Windows [Build Status](https://ci.ros2.org/job/ci_windows/8021/)

PR #330, opened 2019-09-13
install package manifest

dirk-thomas:
For osrf/docker_images#317.

### DIFF
--- a/ros2multicast/setup.cfg
+++ b/ros2multicast/setup.cfg
@@ -26,6 +26,10 @@ tests_require =
   pytest-cov
 zip_safe = true
 
+[options.data_files]
+share/ros2multicast =
+    package.xml
+
 [options.entry_points]
 ros2cli.command =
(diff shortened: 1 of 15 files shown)


ros2multicast/setup.cfg
```diff
@@ -26,6 +26,10 @@ tests_require =
   pytest-cov
 zip_safe = true
 
+[options.data_files]
```
Karsten1987: why is `ros2multicast` different from all the other changes here where `data_files` are installed via `setup.py`?

dirk-thomas: Because the package specifies all its arguments in a `setup.cfg` file. Afaik that is the more modern approach.

Karsten1987: sorry for the noise, this change requires setuptools v42 and does not work with setuptools distributed via debians (v39).

PR #332, opened 2019-09-16
Add warning and error handling for `ifcfg` import on Windows and OSX

claireyywang:
This PR tells windows and osx users to install `ifcfg` with `pip`, handles the error cases when package fails to be imported and when network check/report is unable to run. It adds additional type check for check result and report, to make sure when custom plug-in is implemented incorrectly, the process does not break.

### DIFF
--- a/ros2doctor/ros2doctor/api/__init__.py
+++ b/ros2doctor/ros2doctor/api/__init__.py
@@ -72,6 +72,7 @@ def run_checks() -> Tuple[Set[str], int, int]:
     fail = 0
     total = 0
     for check_entry_pt in iter_entry_points('ros2doctor.checks'):
+        result = ''  # default to invalid
         try:
             check_class = check_entry_pt.load()
         except (ImportError, UnknownExtra):
@@ -103,6 +104,8 @@ def generate_reports(*, categories=None) -> List[Report]:
     """
     reports = []
     for report_entry_pt in iter_entry_points('ros2doctor.report'):
+        report = ''  # default to invalid
+        report_category = ''  # init val
         try:
             report_class = report_entry_pt.load()
         except (ImportError, UnknownExtra):
--- a/ros2doctor/ros2doctor/api/__init__.py
+++ b/ros2doctor/ros2doctor/api/__init__.py
@@ -114,11 +114,11 @@ def generate_reports(*, categories=None) -> List[Report]:
         try:
             report_category = report_instance.category()
             report = report_instance.report()
+            if categories:
+                if report_category in categories:
+                    reports.append(report)
+            else:
+                reports.append(report)
         except Exception:
             doctor_warn('Fail to call %s class functions.' % report_entry_pt.name)
-        if categories:
-            if report_category in categories:
-                reports.append(report)
-        else:
-            reports.append(report)
     return reports
--- a/ros2doctor/ros2doctor/api/__init__.py
+++ b/ros2doctor/ros2doctor/api/__init__.py
@@ -101,8 +101,6 @@ def generate_reports(*, categories=None) -> List[Report]:
     """
     reports = []
     for report_entry_pt in iter_entry_points('ros2doctor.report'):
-        report = Report('')  # default to invalid
-        report_category = ''
         try:
             report_class = report_entry_pt.load()
         except (ImportError, UnknownExtra):
--- a/ros2doctor/ros2doctor/api/__init__.py
+++ b/ros2doctor/ros2doctor/api/__init__.py
@@ -91,7 +91,7 @@ def run_checks() -> Tuple[Set[str], int, int]:
                 failed_cats.add(check_category)
             total += 1
         else:
-            doctor_warn('Check result is invalid in %s.' % check_entry_pt.name)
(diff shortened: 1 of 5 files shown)


ros2doctor/ros2doctor/api/__init__.py
```diff
@@ -85,10 +85,13 @@ def run_checks() -> Tuple[Set[str], int, int]:
             result = check_instance.check()
         except Exception:
             doctor_warn('Fail to call %s class functions.' % check_entry_pt.name)
-        if result is False:
-            fail += 1
-            failed_cats.add(check_category)
-        total += 1
+        if isinstance(result, bool):
```
clalancette: Doesn't this lead to a `UnboundLocalError` if `check_instance.check()` throws an exception?  Actually, worse than that, if *this* loop iteration throws an exception, it will actually re-check the value of the *last* iteration `result`.  The simple way to fix this may be to set `result` to `False` at the top of the loop, and then the rest of this code will work.

ros2doctor/ros2doctor/api/__init__.py
```diff
@@ -85,10 +85,13 @@ def run_checks() -> Tuple[Set[str], int, int]:
...
+            if result is False:
+                fail += 1
+                failed_cats.add(check_category)
+            total += 1
+        else:
+            doctor_warn('Check result from %s is invalid.' % check_entry_pt.name)
```
clalancette: If the plugin is mis-implemented, should we consider that a failure and thus increment `fail`, `total`, and add it to `failed_cats`?  Or add a new return tuple, `broken_cats`?

ros2doctor/ros2doctor/api/__init__.py
```diff
@@ -113,9 +116,12 @@ def generate_reports(*, categories=None) -> List[Report]:
             report = report_instance.report()
         except Exception:
             doctor_warn('Fail to call %s class functions.' % report_entry_pt.name)
-        if categories:
-            if report_category in categories:
+        if isinstance(report, Report):
```
clalancette: Same comments as above.

ros2doctor/ros2doctor/api/__init__.py
claireyywang: `failed_cats` shows report for failed checks only and makes it easier for users to look into ROS related issues. Implementation mistakes are unrelated to ROS and I'd prefer to keep it separated from check failures. `broken_cats` is a feasible solution, but I think current warning and entry point name give enough info for developers to debug. Having a specific structure would seem like we expect lots of implementation errors, and would lead to a feature request like "debug-code-for-me" rabbit hole.

claireyywang: I'd like to keep implementation errors separated from check failures for now as they are unrelated to ROS system issues. Entry name and warning msg give enough info to help debug. I don't expect a lot of plugin errors yet to justify a specific return structure. Good suggestion on `broken_cats` though, I'll keep it in mind for future needs.

clalancette: That's OK as it goes.  To be honest, I wouldn't try to detect broken plugins at all; that's up to the plugin authors to get correct.  If they get it wrong, things won't work, but that's the same thing that happens across the entire codebase.

To be more concrete: I would just remove the whole `if isinstance(result, bool):` block, I just don't think it adds much value.

ros2doctor/ros2doctor/api/__init__.py
```diff
@@ -100,6 +101,8 @@ def generate_reports(*, categories=None) -> List[Report]:
     """
     reports = []
     for report_entry_pt in iter_entry_points('ros2doctor.report'):
+        report = Report('')  # default to invalid
+        report_category = ''
```
clalancette: I'm slightly concerned that making `report_category` an empty string can lead to a difficult-to-find bug.  In particular, if `categories` happens to contain an empty string, and the try/except blocks below throw an exception, we could insert a bogus Report.  Instead, I'll suggest that you move the if/else block at [119](https://github.com/ros2/ros2cli/pull/332/files#diff-26940033a37b6de29cd2efa292e66629R119) into the try block directly above it.  That way, this will never be an issue.

claireyywang:
Linux 
Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/4176/)
OSX [Build Status](https://ci.ros2.org/job/ci_osx/6679/)
Windows [Build Status](https://ci.ros2.org/job/ci_windows/8075/)

ros2doctor/ros2doctor/api/__init__.py
clalancette: You moved the block below (great!) but you forgot to remove these initializations, which should no longer be necessary.

claireyywang: oops 😳

claireyywang:
Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/4177/)
OSX [Build Status](https://ci.ros2.org/job/ci_osx/6680/)
Windows [Build Status](https://ci.ros2.org/job/ci_windows/8077/)

PR #333, opened 2019-09-16
add install instruction for ifcfg package on windows and osx

claireyywang:
Signed-off-by: claireyywang <clairewang@openrobotics.org>

### DIFF
--- a/ros2doctor/README.md
+++ b/ros2doctor/README.md
@@ -7,6 +7,8 @@ It will examine your ROS 2 setup, such as distribution, platform, network interf
 
 ## Usage
 
+Windows and OSX users: run `python3 -m pip install ifcfg` to enable network check.
+
 Run `ros2 doctor` or `ros2 wtf`(alias) to conduct checks.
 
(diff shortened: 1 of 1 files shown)


ros2doctor/README.md
```diff
@@ -7,6 +7,8 @@ It will examine your ROS 2 setup, such as distribution, platform, network interf
 
 ## Usage
 
+Windows and OSX users: run `python3 -m pip install ifcfg` to enable network check.
```
jacobperron: I don't believe `python3` will generically work on Windows. The executable is called `python` in my experience, but probably depends on how the user installed python. Our [install documentation](https://index.ros.org/doc/ros2/Installation/Dashing/Windows-Install-Binary/#install-dependencies) uses `python` for Windows. Speaking of which, I think we could list this install instruction on the ROS 2 installation documentation instead of here. What do you think?

claireyywang: Good point.

claireyywang:
Closing PR: moving installation instruction to ROS2 doc.

PR #338, opened 2019-09-24
ros2doctor: add `--include-warning` arg

claireyywang:
Adds arg `--include-warning/-iw` to ros2doctor CLI. It reports both warnings and errors as failed checks, while the default only includes errors. It can be used in combination with `--report-failed/-rf` or by itself when running `ros2 doctor` checks. It lets users choose how extensive the checks and reports cover.

### DIFF
--- a/ros2doctor/ros2doctor/api/network.py
+++ b/ros2doctor/ros2doctor/api/network.py
@@ -18,6 +18,7 @@ from typing import Tuple
 from ros2doctor.api import DoctorCheck
 from ros2doctor.api import DoctorReport
 from ros2doctor.api import Report
+from ros2doctor.api import Result
 from ros2doctor.api.format import doctor_warn
 
 try:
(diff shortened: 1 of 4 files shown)


ros2doctor/ros2doctor/api/network.py
```diff
@@ -54,21 +55,26 @@ def category(self):
...
         except NameError:
             doctor_warn('ifcfg is not imported. Unable to run network check.')
-            return False
+            result.increment_error()
+            return result
 
         has_loopback, has_non_loopback, has_multicast = _check_network_config_helper(ifcfg_ifaces)
         if not has_loopback:
             doctor_warn('ERROR: No loopback IP address is found.')
+            result.increment_error()
```
sloretz: It seems like this could be made a little simpler for Check classes. The two methods make sense from the perspective of `ros2doctor` where `doctor_warn()` is telling a human there was a problem and `increment_[error|warning]()` is telling code, but `NetworkCheck` doesn't need to know who is using it. What do you think of making the `Result` class responsible for that?

```python3
result.add_error('No loopback IP address is found')
...
result.add_warning('Only loopback IP address is found')
...
result.add_warning('No multicast IP address is found')
```

ros2doctor/ros2doctor/command/doctor.py
```diff
@@ -41,8 +45,11 @@ def main(self, *, parser, args):
                 format_print(report_obj)
             return
 
-        # `ros2 doctor`
-        failed_cats, fail, total = run_checks()
+        # `ros2 doctor
+        if args.include_warnings:
+            failed_cats, fail, total = run_checks(include_warnings=True)
+        else:
+            failed_cats, fail, total = run_checks()
```
sloretz: nit, can simplify to

```python3
failed_cats, fail, total = run_checks(include_warnings=args.include_warnings)
```

PR #346, opened 2019-10-03
[ros2multicast] Use setup.py fields rather than setup.cfg

mjcarroll:
Addresses failures in building ros2multicast for Eloquent debs.

Support for `data_files` in `setup.cfg` was introduced in [setuptools v40.5.0](https://setuptools.readthedocs.io/en/latest/history.html#v40-5-0), while Ubuntu Bionic has version `39.0.1`.

This follows the pattern of the rest of the `ros2cli` packages and puts the contents in `setup.py`


dirk-thomas:
Why not only move the `data_files` into the `setup.py` file?

mjcarroll:
> Why not only move the data_files into the setup.py file?

None of the other packages in this repository use the `setup.cfg` file, this seemed like the most consistent choice.

PR #352, opened 2019-10-07
Fix multicast entrypoint

jacobperron:
Otherwise, we see an error every time we run a ros2 command:

    Failed to load entry point 'multicast': No module named 'ros2node.command.multicast'

Bug introduced in #346

### DIFF
--- a/ros2multicast/setup.py
+++ b/ros2multicast/setup.py
@@ -34,7 +34,7 @@ The package provides the multicast command for the ROS 2 command line tools.""",
     tests_require=['pytest'],
     entry_points={
         'ros2cli.command': [
-            'multicast = ros2node.command.multicast:MulticastCommand',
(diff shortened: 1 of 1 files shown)


ros2multicast/setup.py
```diff
@@ -34,7 +34,7 @@
     tests_require=['pytest'],
     entry_points={
         'ros2cli.command': [
-            'multicast = ros2node.command.multicast:MulticastCommand',
+            'multicast = ros2multicast.command.multicast:MulticastCommand',
         ],
         'ros2cli.extension_point': [
             'ros2node.verb = ros2node.verb:VerbExtension',
```
dirk-thomas: This line looks wrong too.

jacobperron: Good catch. d07bb14

PR #355, opened 2019-10-08
Start version at 0.1.0

sloretz:
This changes package templates to start at version `0.1.0` instead of `0.0.0`. [0.1.0 is recommended by Semantic Versioning](https://semver.org/#how-should-i-deal-with-revisions-in-the-0yz-initial-development-phase). I think it makes sense to change it for the benefit of those that do use semantic versioning, assuming those who don't also don't care what the start version of the template is.

### DIFF
--- a/ros2pkg/ros2pkg/resource/ament_python/setup.py.em
+++ b/ros2pkg/ros2pkg/resource/ament_python/setup.py.em
@@ -4,7 +4,7 @@ package_name = '@project_name'
 
 setup(
     name=package_name,
-    version='0.0.0',
+    version='0.1.0',
     packages=[package_name],
     data_files=[
         ('share/ament_index/resource_index/packages',
--- a/ros2pkg/ros2pkg/resource/cmake/CMakeLists.txt.em
+++ b/ros2pkg/ros2pkg/resource/cmake/CMakeLists.txt.em
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.5)
 project(@(project_name))
 
 set(@(project_name)_MAJOR_VERSION 0)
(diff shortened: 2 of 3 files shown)


dirk-thomas:
Note this is the version number used *before* a first release is being made. A version 0.1.0 sounds like a good first tagged version but that is not the number which should be generated for the stub package since it is being bumped when actually doing the release using e.g. `catkin_prepare_release`.

sloretz:
> A version 0.1.0 sounds like a good first tagged version but that is not the number which should be generated for the stub package since it is being bumped when actually doing the release using e.g. catkin_prepare_release

That explains [all the `0.0.1` releases](https://github.com/ros/rosdistro/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aclosed+0.0.1+in%3Atitle+) . I opened ros-infrastructure/catkin_pkg#267 with a proposal for fixing that.

dirk-thomas:
> 0.1.0 is recommended by Semantic Versioning.

To quote the semver docs: "The simplest thing to do is start your initial development release at 0.1.0 ...". I wouldn't call that a recommendation but merely an example so people don't start with 1.0.0.

> I think it makes sense to change it for the benefit of those that do use semantic versioning

Even with 0.0.1 packages can use sem ver - there is no difference between those two starting versions.

So I don't see the reason why 0.0.1 a bad starting point? I would even argue since the package hasn't been tagged yet it is even better since it is the lowest version number.

sloretz:
> So I don't see the reason why 0.0.1 a bad starting point?

Looks like `0.0.z` is valid after all. semver.org has a regular expression example that includes `0.0.4` as a valid version. Closing.

https://regex101.com/r/Ly7O1x/3/

PR #356, opened 2019-10-10
Check/Report package versions

### LINKED ISSUE #343: ros2doctor: add support for warning on out of date packages
A common error that people have is that they may have old our outdated packages on their system. This is most common with build pipeline tools, but can also have an effect due to a partial install or update of their debian packages. 

ros2doctor could collect the list of all local packages in the workspace with their version number and then check against a declared rosdistro for the released version number. And if the latest version number and the version in the rosdistro do not match it should add it to the warnings. 

We don't have the rosdistro cache for our latest bootstrap packages, but there may be an approach that would be reasonable to query pypi for the latest version or the apt servers, or possibly a github API. Bloom has logic like this to warn if it's not the latest version that could potentially be replicated.

In the report mode it would be valuable to be able to export the list of packages present on the system as well as their version number. 

In report mode another extension that would be valuable would be the `vcs export --exact` of the workspace for easier reproduction/debugging. As well as potentially listing the exact debian packages (with build numbers) that have been installed for easier reproduction.

claireyywang:
close #343 
Check locally installed package versions against ones on rosdistro. Local versions are obtained by parsing `package.xml` found in install space.


ros2doctor/package.xml
```diff
@@ -9,6 +9,7 @@
 
   <depend>ros2cli</depend>
 
+  <exec_depend>ament_index</exec_depend>
```
dirk-thomas: There is no package with that name. I guess you need `ament_index_python`.

ros2doctor/ros2doctor/api/package.py
```diff
@@ -0,0 +1,127 @@
...
+    """
+    try:
+        prefix = get_package_prefix(package_name)
+    except PackageNotFoundError:
+        if check:  # suppress warning when printing report
+            doctor_warn('Unable to find prefix for `%s` using index system.' % package_name)
+        return
+    xml_file = os.path.join(prefix, 'share', package_name, 'package.xml')
+    try:
+        root = ET.parse(xml_file).getroot()
```
dirk-thomas: Instead of parsing the manifest file yourself please use the API provided by the Python package `catkin_pkg` (rosdep key `python3-catkin-pkg-modules`) instead.

ros2doctor/ros2doctor/api/package.py
```diff
@@ -0,0 +1,127 @@
...
+                distro_release = info.get('release')
+                distro_ver = distro_release.get('version')
+            except AttributeError:
+                result.add_warning('Unable to obtain the rosdistro version \
+                    of `%s`.' % package_name)
+            if local_ver and distro_ver:
+                if distro_ver[:3] != local_ver[:3]:  # 0.8.0 vs. 0.8.0-1
+                    result.add_warning('`%s` has different local version %s from \
+                        required version %s' % (package_name, distro_ver, local_ver))
+        return result
```
dirk-thomas: Instead of getting all packages and versions from rosdistro and then checking if they exist on the system I would suggest the opposite: collect all packages found and their versions and then compare them to the information from rosdistro.

ros2doctor/ros2doctor/api/package.py
```diff
@@ -0,0 +1,112 @@
...
+import rosdistro
+
+
+def get_ros_packages_info() -> dict:
+    """
+    Return all current distro's packages info using rosdistro API.
+
+    :return: a dictionary contains package name, release, source and status
+    """
+    distro_name = os.environ.get('ROS_DISTRO')
```
sloretz: It looks like this will raise `RuntimeError` if `ROS_DISTRO` is unset. What should `PackageCheck` and `PackageReport` do in that case?

```python3
>>> import rosdistro
>>> url = rosdistro.get_index_url()
>>> i = rosdistro.get_index(url)
>>> data = rosdistro.get_distribution(i, None).get_data()
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3/dist-packages/rosdistro/__init__.py", line 114, in get_distribution
    dist_file = get_distribution_file(index, dist_name)
  File "/usr/lib/python3/dist-packages/rosdistro/__init__.py", line 119, in get_distribution_file
    data = _get_dist_file_data(index, dist_name, 'distribution')
  File "/usr/lib/python3/dist-packages/rosdistro/__init__.py", line 186, in _get_dist_file_data
    raise RuntimeError("Unknown release: '{0}'. Valid release names are: {1}".format(dist_name, ', '.join(sorted(index.distributions.keys()))))
RuntimeError: Unknown release: 'None'. Valid release names are: ardent, bouncy, crystal, dashing, eloquent, groovy, hydro, indigo, jade, kinetic, lunar, melodic, noetic
```

ros2doctor/ros2doctor/api/package.py
```diff
@@ -0,0 +1,112 @@
...
+    """Check local package versions against release versions on rosdistro."""
+
+    def category(self):
+        return 'package'
+
+    def check(self):
+        """Check packages within the directory where command is called."""
+        result = Result()
+        try:
+            distro_packages_info = get_ros_packages_info()
```
sloretz: Recommend `distro_packages_info -> distro_repositories_info` and `get_ros_packages_info -> get_ros_repositories_info` since it's a dictionary of repositories, which may have 1 or more packages in them.

ros2doctor/ros2doctor/api/package.py
```diff
@@ -0,0 +1,112 @@
...
+            result.add_error('ERROR: Unable to obtain current distro package information \
+                from rosdistro.')
+            return result
+        local_packages_prefixes = get_packages_with_prefixes()
+        if not local_packages_prefixes or not distro_packages_info:
+            result.add_error('ERROR: local packages info or distro packages info is empty.')
+            return result
+        for package_name, package_prefix in local_packages_prefixes.items():
+            file_path = os.path.join(package_prefix, 'share', package_name)
+            if package_name in distro_packages_info:
```
sloretz: It looks like this won't work for repositories that have more than one package. For example, `rcl_action` is in the repository `rcl`.

```python3
>>> distro_packages_info['rcl']
{'doc': {'type': 'git', 'url': 'https://github.com/ros2/rcl.git', 'version': 'dashing'}, 'release': {'url': 'https://github.com/ros2-gbp/rcl-release.git', 'version': '0.7.7-1', 'tags': {'release': 'release/dashing/{package}/{version}'}, 'packages': ['rcl', 'rcl_action', 'rcl_lifecycle', 'rcl_yaml_param_parser']}, 'source': {'type': 'git', 'url': 'https://github.com/ros2/rcl.git', 'version': 'dashing', 'test_pull_requests': True}, 'status': 'developed'}
```
Note the `rcl` repo has a key `packages` which is a list of packages in the repo, which includes `rcl_action`.

There is no repo `rcl_action`.
```python3
>>> distro_packages_info['rcl_action']
  File "<stdin>", line 1, in <module>
KeyError: 'rcl_action'
```

ros2doctor/ros2doctor/api/package.py
```diff
@@ -0,0 +1,112 @@
...
+                                       % package_name)
+                    required_ver = ''
+            try:
+                package_obj = parse_package(file_path)
+                local_ver = package_obj.version
+            except (AttributeError, IOError, InvalidPackage):
+                result.add_warning('Unable to parse `%s` package.xml file.' % package_name)
+                local_ver = ''
+            if required_ver and local_ver:
+                if required_ver[:3] != package_obj.version[:3]:
```
sloretz: Recommend warning only if the released version is greater than the installed version, meaning the user needs to upgrade their packages. If the installed version is greater, the person calling `ros2 doctor` may be intentionally building a newer version from source.

ros2doctor/ros2doctor/api/package.py
```diff
@@ -0,0 +1,112 @@
...
+        report = Report('PACKAGE VERSIONS')
+        try:
+            distro_packages_info = get_ros_packages_info()
+        except AttributeError:
+            return report
+        local_package_prefixes = get_packages_with_prefixes()
+        if distro_packages_info and local_package_prefixes:
+            for package_name, package_prefix in local_package_prefixes.items():
+                try:
+                    package_info = distro_packages_info.get(package_name)
```
sloretz: Same comment about `distro_packages_info` being repository info, and this not working for repos with multiple packages.

ros2doctor/ros2doctor/api/package.py
claireyywang: I'll add `RuntimeError` to the `except` block in PackageCheck and PackageReport. In check, it will pass error message to result; in report, it will output an empty report with names only.

claireyywang:
messed up local, gonna make a new branch and PR

PR #357, opened 2019-10-10
[ros2node] Update info headings for actions

jacobperron:


jacobperron:
It's a bit of a nitpick :grin:

PR #364, opened 2019-10-17
add completion for parameter name arguments

dirk-thomas:
For the `get` and `delete` verbs. Also naming the argument `parameter_name` (rather than `name`) in the usage.

### DIFF
--- a/ros2param/ros2param/api/__init__.py
+++ b/ros2param/ros2param/api/__init__.py
@@ -15,7 +15,9 @@
 from rcl_interfaces.msg import ParameterType
 from rcl_interfaces.msg import ParameterValue
 from rcl_interfaces.srv import GetParameters
+from rcl_interfaces.srv import ListParameters
 from rcl_interfaces.srv import SetParameters
+from ros2cli.node.direct import DirectNode
 import rclpy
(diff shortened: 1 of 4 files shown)


dirk-thomas:
> My sole concern is response latency. How long does it take to provide suggestions?

It takes as long as it takes to spin up the node and wait for the service.

> I ask because I've seen ros2 topic almost blocking while retrieving topics.

In this case the delay happens when you use completion. If you don't it doesn't affect the call.

> `DirectNode` brings up a node on construction and tears it down on context leave -- maybe keeping it around may speed up subsequent queries, not the first though.

It would be great if instead this information would be queried over the node but that kind of API is not available atm.

> why not adding completion for the `set` verb?

Added in 96bee44e7fd097a5bc3829deb217aecf77ea0231.

hidmic:
> It would be great if instead this information would be queried over the node daemon but that kind of API is not available atm.

Yeah, that's what I was thinking of. Alright, food for thought.

dirk-thomas:
Sanity linter check:
* Linux: [Build Status](https://ci.ros2.org/job/ci_linux/8420/)

PR #365, opened 2019-10-17
Backport creating ament python packages to dashing

sloretz:
Backports `ros2 pkg create --build-type ament_python` from #296 to assist with ros2/ros2_documentation#362.

This PR does change two command line arguments `--cpp-node-name` and `--cpp-library-name` to `--node-name` and `--library-name`. I think that breakage is acceptable since it seems unlikely someone has automated calls to `ros2 pkg create` CLI, and backporting means the ROS 2 tutorial for package creation will work for Dashing.

### DIFF
--- a/ros2pkg/ros2pkg/api/create.py
+++ b/ros2pkg/ros2pkg/api/create.py
@@ -101,8 +101,77 @@ def create_package_environment(package, destination_directory):
         print('creating source and include folder')
         source_directory = _create_folder('src', package_directory)
         include_directory = _create_folder(package.name, package_directory + os.sep + 'include')
+    if package.get_build_type() == 'ament_python':
+        print('creating source folder')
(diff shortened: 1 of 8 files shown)


dirk-thomas:
> I think that breakage is acceptable since it seems unlikely someone has automated calls

Why is that unnecessary breakage acceptable? Even if there are no automated calls maybe people have written down instructions how to use it?

It is trivial to avoid those breakage by keeping the previous option names as an alternative beside the new name.

sloretz:
> Why is that unnecessary breakage acceptable? Even if there are no automated calls maybe people have written down instructions how to use it?

True someone might have written instructions. I didn't think it was that big of a deal. Maybe someone would have tried `--help`, found the renamed options, and updated their notes. I added the options back in 154e92c

PR #366, opened 2019-10-17
Handle bad or missing package on ros2 interface show

hidmic:
Precisely what the title says. A #304 spin-off, comprehensive test coverage can be found there.


hidmic:
Sanity linter check:
* Linux: [Build Status](https://ci.ros2.org/job/ci_linux/8422/)

PR #367, opened 2019-10-17
add ros2 param describe

### LINKED ISSUE #362: [ros2param] Add verb to describe parameters
#### Feature description

Every parameter has a [descriptor](https://github.com/ros2/rcl_interfaces/blob/93cedce2dacd25fa4f2969d022ccbe3f2903e3fe/rcl_interfaces/msg/ParameterDescriptor.msg). It would be nice if the `ros2 param` CLI tool had a verb for getting the description of a parameter.

The verb could be called `describe` or `info`. I think I prefer `info` since it's shorty and more consistent with our other CLI tools.

For example,

```sh
$ ros2 param info foo_node bar_param
Name: bar_param
Description: This is a text description of the parameter
Type: double
Constraints:
    Read only: false
    Min: 0.0
    Max: 1.0
    Step: 0.1
    Additional constraints:  A text description of constraints (if there is one).
```

There might be some better ideas for formatting, but I think the `Min`, `Max`, and `Step` constraints should only appear if there is a range constraint.
And the `Additional constraints` should only appear if the string is not empty.

Each node provides a [service](https://github.com/ros2/rcl_interfaces/blob/93cedce2dacd25fa4f2969d022ccbe3f2903e3fe/rcl_interfaces/srv/DescribeParameters.srv) (enabled by default) for requesting one or more parameter descriptors, so the tool can make a single service call when `ros2 param info` is called.
Considering that the input to the service request is a list of names, it might also make sense to support querying multiple parameters at a time in the CLI tool.

For example,

```sh
$ ros2 param info foo_node bar_param baz_param another_param
Name: bar_param
Description: This is a text description of the bar_param parameter
Type: string
Constraints:
    Read only: false
    Additional constraints:  A text description of constraints (if there is one).

Name: baz_param
Description: This is a text description of the baz parameter
Type: double
Constraints:
    Read only: true
    Min: 0.0
    Max: 1.0
    Step: 0.1

Name: another_param
Description: This is a text description of the another_param parameter
Type: integer array
Constraints:
    Read only: false
```

dirk-thomas:
Fixes #362.

I chose to use the parameter name as the item without indentation and all other keys are then indented one level. That works nicely for multiple parameters and doesn't use extra empty lines. See example output:

```
$ ros2 param describe turtlesim background_r background_g
Parameter name: background_r
  Type: integer
  Description: Red channel of the background color
  Constraints:
    Min value: 0
    Max value: 255
Parameter name: background_g
  Type: integer
  Description: Green channel of the background color
  Constraints:
    Min value: 0
    Max value: 255
    Additional constraints: foo
```

### DIFF
--- a/ros2param/ros2param/verb/describe.py
+++ b/ros2param/ros2param/verb/describe.py
@@ -93,4 +93,3 @@ def get_parameter_type_string(parameter_type):
         ParameterType.PARAMETER_NOT_SET: 'not set',
     }
     return mapping[parameter_type]
-
--- a/ros2param/ros2param/verb/describe.py
+++ b/ros2param/ros2param/verb/describe.py
@@ -1,0 +1,96 @@
+# Copyright 2019 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from rcl_interfaces.msg import ParameterType
+from ros2cli.node.direct import DirectNode
+from ros2cli.node.strategy import add_arguments
+from ros2cli.node.strategy import NodeStrategy
+from ros2node.api import get_absolute_node_name
+from ros2node.api import get_node_names
+from ros2node.api import NodeNameCompleter
+from ros2param.api import call_describe_parameters
+from ros2param.api import ParameterNameCompleter
+from ros2param.verb import VerbExtension
+
+
+class DescribeVerb(VerbExtension):
+    """Show descriptive information about declared parameters."""
+
+    def add_arguments(self, parser, cli_name):  # noqa: D102
+        add_arguments(parser)
+        arg = parser.add_argument(
+            'node_name', help='Name of the ROS node')
+        arg.completer = NodeNameCompleter(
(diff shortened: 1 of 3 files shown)


ros2param/ros2param/verb/describe.py
```diff
@@ -0,0 +1,96 @@
...
+        ParameterType.PARAMETER_STRING: 'string',
+        ParameterType.PARAMETER_BYTE_ARRAY: 'byte array',
+        ParameterType.PARAMETER_BOOL_ARRAY: 'boolean array',
+        ParameterType.PARAMETER_INTEGER_ARRAY: 'integer array',
+        ParameterType.PARAMETER_DOUBLE_ARRAY: 'double array',
+        ParameterType.PARAMETER_STRING_ARRAY: 'string array',
+        ParameterType.PARAMETER_NOT_SET: 'not set',
+    }
+    return mapping[parameter_type]
+
```
mjcarroll: Blank line makes flake8 unhappy.

dirk-thomas: Fixed: 346ac17eeaa652433c264284162769d6a7a3b9e5.

dirk-thomas:
> Since we can know the list of parameters, I think it would make for a nice feature if no parameter names are provided, then we describe all of them.

We could query the list with an additional service call. Unfortunately the `DescribeParameter` service doesn't offer a way to query all parameters without specifying them.

dirk-thomas:
Sanity linter check:
* Linux [Build Status](https://ci.ros2.org/job/ci_linux/8425/)

jacobperron:
> We could query the list with an additional service call. Unfortunately the DescribeParameter service doesn't offer a way to query all parameters without specifying them.

But I guess the tool but know about them from an additional service call since auto-completion is working?

dirk-thomas:
> But I guess the tool knows about them from an additional service call since auto-completion is working?

Not sure I understand but currently we don't call the list service. While it is being invoked during completion it is not run during the actual command.

As I said: it can certainly be added (with the extra call time). Feel free to do so in a new PR.

PR #371, opened 2019-10-19
Move rosidl implementation details to rosidl_runtime_py

jacobperron:
Implementation has been moved and refactored in https://github.com/ros2/rosidl_runtime_py/pull/3
This resolves several TODOs (5b54d70).

Also, using the abstracted functions it's easier to remove dependencies on packages deprecated in #363 (b01c63f).

Depends on  https://github.com/ros2/rosidl_runtime_py/pull/3

PR #372, opened 2019-10-21
Make ros2 interface show fail gracefully (no traceback)

hidmic:
Follow-up after #371. This pull request ensures `ros2 interface show` fails nicely, with no traceback.


hidmic:
CI (linter sanity check)

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8502/)

PR #384, opened 2019-10-24
Fix ros2topic test_echo_pub.py test suite

hidmic:
This pull request ensures we:
1. Stop re-using the same `rclpy` context, node and executor across tests. This causes failures on Linux-aarch64 when using Opensplice.
2. Force tests using `rmw_fastrtps*_cpp` to run before those using `rmw_opensplice_cpp`. Otherwise, we get failures for the former on Windows 10.  

My guess for (1.) is that publisher, subscriber, timer, etc. creation/destruction or even executor spinning on one test is negatively affecting others, whereas (2.) is simply mind blowing. Someway, somehow, `rmw_opensplice_cpp` is affecting `rmw_fastrtps_cpp` if run consecutively, not even simultaneously, on the same process. I've left TODOs and NOTES, we need to come back to this at some point, when we have more time.

### DIFF
--- a/ros2topic/test/test_echo_pub.py
+++ b/ros2topic/test/test_echo_pub.py
@@ -35,8 +35,6 @@ from rclpy.qos import DurabilityPolicy
 from rclpy.qos import QoSProfile
 from rclpy.qos import ReliabilityPolicy
 
-from rmw_implementation import get_available_rmw_implementations
-
 from std_msgs.msg import String
 
 
@@ -45,14 +43,8 @@ TEST_NAMESPACE = 'cli_echo_pub'
 
 
 @pytest.mark.rostest
-# NOTE(hidmic): this forces an execution order of RMW implementation parameterized tests
-#               i.e. a test instance using 'rmw_fastrtps_cpp' is executed before a test
-#               instance using 'rmw_opensplice_cpp'.
-# TODO(hidmic): investigate why running these tests against 'rmw_opensplice_cpp' first and
-#               against 'rmw_fastrtps*_cpp' after makes the latter fail on Windows 10.
-@launch_testing.parametrize('rmw_implementation', sorted(get_available_rmw_implementations()))
 @launch_testing.markers.keep_alive
-def generate_test_description(rmw_implementation, ready_fn):
+def generate_test_description(ready_fn):
     return LaunchDescription([
         # Always restart daemon to isolate tests.
         ExecuteProcess(
@@ -65,7 +57,6 @@ def generate_test_description(rmw_implementation, ready_fn):
                     on_exit=[
                         OpaqueFunction(function=lambda context: ready_fn())
                     ],
-                    additional_env={'RMW_IMPLEMENTATION': rmw_implementation}
                 )
             ]
         )
@@ -79,8 +70,7 @@ class TestROS2TopicEchoPub(unittest.TestCase):
     #               makes them fail on Linux-aarch64 when using 'rmw_opensplice_cpp'.
     #               Presumably, interfaces creation/destruction and/or executor spinning
     #               on one test is affecting the other.
-    def setUp(self, rmw_implementation):
-        os.environ['RMW_IMPLEMENTATION'] = rmw_implementation
+    def setUp(self):
         self.context = rclpy.context.Context()
         rclpy.init(context=self.context)
         self.node = rclpy.create_node(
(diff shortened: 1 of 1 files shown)


hidmic:
CI (up to `ros2topic`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8554/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4427/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6974/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8468)

hidmic:
Ok, I had to remove all `RMW_IMPLEMENTATION` parameterization. Because dynamic linkage of rmw symbols occurs once (and quite early too, see https://github.com/ros2/rcl/blob/513a185d1b8ad182f74519194682f10f132e5dc9/rcl/src/rcl/rmw_implementation_identifier_check.c#L40-L75), it's not possible to affect said setting from within the test itself (which in retrospection is pretty obvious).

hidmic:
@mjcarroll PTAL

nuclearsandwich:
@ros-pull-request-builder retest this please.

nuclearsandwich:
> Ok, I had to remove all `RMW_IMPLEMENTATION` parameterization

With this removed I don't see how are you handling issue (2) ? I expect it's still there I just don't understand it.

nuclearsandwich:
The only test failure on the Epr job is also failing on the Edev job currently.

hidmic:
> With this removed I don't see how are you handling issue (2) ? I expect it's still there I just don't understand it.

It's very likely still there, though now it's way more obscure.

See, affecting the launch test environment like this is brittle. Subprocesses' do abide to the setting, but any ROS node instances within the launch test process space will do so once, the first time an `rmw` API symbol gets resolved (see [`rmw_implementation` package sources](https://github.com/ros2/rmw_implementation/blob/master/rmw_implementation/src/functions.cpp#L122)). For `rclpy`, because C extension loading is delayed for some reasons that are not relevant to this, we might be able to have an impact if we tweak the environment _early enough_, though in launch tests that's impossible because launch itself is spinning up its own node way before the tests are run. 

If my logic is right, we weren't really testing pub/sub on a single RMW implementation but across them (the one that was there when `rclpy` was first initialized and the one being used for subprocesses).

hidmic:
@wjwwood @dirk-thomas I'm not sure if we should be louder about this, though I don't know how often someone may want to run the same test against multiple RMW implementations.

hidmic:
Argh, the default implementation changed.

wjwwood:
I don't know exactly what you're asking. Should we be louder about...? I mean the rmw implementation can only be picked before the process starts and only by use of the environment variable. I think it follows logically that it cannot be changed once the test has started.

This is why we have separate tests for each rmw in the cmake packages that do this. In rclpy we used to use `multiprocessing` to have separate rmw implementations tested, but we no longer do that.

For python packages I think the only way is to run tests as subprocesses and change the `RMW_IMPLEMENTATION` env var before you start the test. Since how pytest is run is out of our hands (from the perspective of the package) we cannot choose, much less change, which rmw is being used.

We may be able to assert the same rmw is being used in a few places to avoid unintentional mixing, but that's all I think.

hidmic:
> I mean the rmw implementation can only be picked before the process starts

Louder about precisely this. It follows logically if you have a clear understanding of how the system works under the hood, but do we have a bit of documentation stating that the chosen rmw implementation is fixed throughout the lifetime of a process? Couldn't find any in the corresponding tutorial. 

It follows logically if

wjwwood:
Well this section only gives one way to change it, and that's with an environment variable:

https://index.ros.org/doc/ros2/Tutorials/Working-with-multiple-RMW-implementations/#specifying-rmw-implementations

We can add a note there that it cannot change after starting the process, but I don't know why you would assume it could. Environment variables are usually only considered when you start a process, it's very uncommon that changing it during runtime affects the process. At least in my experience.

hidmic:
> This is why we have separate tests for each rmw in the cmake packages that do this.

Yeap, it's clear now. I wonder if we should have an easy/standard way of writing these kind of tests. As opposed to ad hoc solutions for each package, which I think we don't do often either.

hidmic:
Alright, tests are passing now except on MacOS for both Connext and Opensplice, but the same missing shared object issues are also happening on `master`. @mjcarroll let me know when and if we have green light to merge.

hidmic:
Alright, going in!

Issue #349, opened 2019-10-04
[ros2interface] does not allow input from stdin for `show`


wjwwood:
#### Feature description

In this ROS 1 tutorial (https://wiki.ros.org/ROS/Tutorials/UnderstandingTopics#Using_rostopic_hz) there is a shell one-liner:

```
$ rostopic type /turtle1/cmd_vel | rosmsg show
```

Which then shows the structure of the topic type for the `/turtle1/cmd_vel`.

It would be nice to have this ability in the ROS 2 `ros2 interface show` command line tool too.

sharminramli:
I'd like to work on this, but it looks like ros2msg was deprecated in #363. Can I go ahead and look into implementing this feature in ros2interface if it's still relevant?

jacobperron:
@sharminramli Thanks for volunteering to look into this!
Correct, ros2msg is deprecated, but I believe the feature is still relevant to ros2interface. I'll change the title and description of this ticket accordingly.

sharminramli:
this can probably be closed?

clalancette:
Yeah, good call.  Closing this out.

PR #387, opened 2019-10-25
[ros2interface] Allow stdin input for 'ros2 interface show'

sharminramli:
Implements #349.

Interface types can now be piped directly into show, like this:
`$ ros2 topic type /chatter | ros2 interface show`

### DIFF
--- a/ros2interface/ros2interface/verb/show.py
+++ b/ros2interface/ros2interface/verb/show.py
@@ -40,7 +40,7 @@ class ShowVerb(VerbExtension):
         arg = parser.add_argument(
             'type',
             action=ReadStdinPipe,
-            help="Show an interface definition (e.g. 'std_msgs/msg/String'). "
+            help="Show an interface definition (e.g. 'example_interfaces/msg/String'). "
                  "Passing '-' reads the argument from stdin (e.g. "
                  "'ros2 topic type /chatter | ros2 interface show -').")
         arg.completer = type_completer
--- a/ros2interface/ros2interface/verb/show.py
+++ b/ros2interface/ros2interface/verb/show.py
@@ -28,6 +28,8 @@ class ReadStdinPipe(argparse.Action):
             if sys.stdin.isatty():
                 parser.error('expected stdin pipe')
             values = sys.stdin.readline().strip()
+        if not values:
+            parser.error('the passed value is empty')
         setattr(namespace, self.dest, values)
 
 
--- a/ros2interface/ros2interface/verb/show.py
+++ b/ros2interface/ros2interface/verb/show.py
@@ -25,12 +25,10 @@ class ReadStdinPipe(argparse.Action):
 
     def __call__(self, parser, namespace, values, option_string=None):
         if values == '-':
-            if not sys.stdin.isatty():
-                setattr(namespace, self.dest, sys.stdin.readline().strip())
-            else:
+            if sys.stdin.isatty():
                 parser.error('expected stdin pipe')
-        else:
-            setattr(namespace, self.dest, values)
+            values = sys.stdin.readline().strip()
+        setattr(namespace, self.dest, values)
 
 
 class ShowVerb(VerbExtension):
--- a/ros2interface/ros2interface/verb/show.py
+++ b/ros2interface/ros2interface/verb/show.py
@@ -14,7 +14,6 @@
 
 import sys
 import argparse
-import textwrap
 
 from ros2interface.api import type_completer
 from ros2interface.verb import VerbExtension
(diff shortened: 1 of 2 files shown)


sharminramli:
I've attached an action to `add_argument` that reads from stdin if `-` was passed. Requiring `-` preserves the parser's `required` argument check on `type` since we don't have to allow `type` to be missing. This requires explicitly running

`$ ros2 topic type /chatter | ros2 interface show -`

rather than

`$ ros2 topic type /chatter | ros2 interface show`

This also replaces reading `stdin` directly for the default value and fixes the resulting exceptions.

I've also added more information in the help but I'm not sure if it's written clearly enough, or if there's a convention that I'm breaking.

dirk-thomas:
Thanks for the updates.

I applies a few changes. Please have a look:
* 0134dd07ccb5d0c5486b7e1c4e1a58715371ffe6 Move all the help text into the `help` attribute. That avoids potentially overwriting the epilog if set externally. Also the `textwrap.dedent` seems unnecessary.
* a085e270b6ec5f9e8e031e0844a6ba39f46e355b Fix the import order to pass the linters.
* f14c359c4456e2b141fbfb9eaad55dbcf2608dd7 Simplify the logic in the custom action to avoid duplicate parts.
* 3e7241a0818c79fe5f8795980b4092cb7acadfea Handle the case where the piped command fails / doesn't output anything more gracefully.

dirk-thomas:
It would be great if you could add a unit test to cover this new feature to avoid it regressing in the future.

dirk-thomas:
> It would be great if you could add a unit test to cover this new feature to avoid it regressing in the future.

@sharminramli Friendly ping.

sharminramli:
@dirk-thomas Sorry about the delay. Thanks for the changes, they look great.

I started working on the unit tests, but I ran into some issues piping stdin to the test process. I am patching a mock stdin to the test function with
```
    @patch('sys.stdin', new_callable=StringIO)
    def test_show_message_from_stdin(self, mocked_stdin):
        mocked_stdin.write('std_msgs/msg/String')
        mocked_stdin.seek(0)
        with self.launch_interface_command(
            arguments=['show', '-']
        ) as interface_command:
            assert interface_command.wait_for_shutdown(timeout=2)
        assert interface_command.exit_code == launch_testing.asserts.EXIT_OK
        assert launch_testing.tools.expect_output(
            expected_lines=[
                'string data'
            ],
            text=interface_command.output,
            strict=True
        )
```
but it appears to hang at
`values = sys.stdin.readline().strip()` of `show.py`. Is there something I'm missing?

hidmic:
@sharminramli you won't be able to pipe data into the launched process like that. `launch` does not expose the `stdin` pipe either, so your best bet right now is to use `shell=True` and pipe. For that, you'll have to change the test fixture a bit to be able to launch an action like:

```python
command_action = ExecuteProcess(
  cmd=[sys.executable, '-c', "print('std_msgs/msg/String')", '|', 'ros2', 'interface', 'show', '-'],
  shell=True,
  output='screen'
)
``` 

Alternatively, you can propose a reasonable to expose the `stdin` pipe for launched processes in general.

dirk-thomas:
@sharminramli Friendly ping.

sharminramli:
@dirk-thomas Sorry for the wait. I've added the test as `test_show_stdin` in `test_cli.py`.
@hidmic Thanks for the help. I'm not familiar with the `launch` repo as of now, so I've gone with the first option and modified `launch_interface_command` to accept the `shell` option and a list of arguments to be prepended.

sharminramli:
Rebased and changed the test to use `test_msgs/msg/Nested` rather than `std_msgs/msg/String` (https://github.com/ros2/common_interfaces/issues/89). The help message should also be updated but it's being modified in https://github.com/ros2/ros2cli/pull/516 as well.

hidmic:
> however, `test_show_message` fails because of the comments added to the message file in Foxy.

That was fixed in #516. As a result, this now needs a rebase.

sharminramli:
Rebased and changed the help message to refer to `example_interfaces` rather than `std_msgs`. I'm not sure why the Jenkins build is failing though.

audrow:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10895/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6251/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8873/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10787/)

hidmic:
Thanks for re-running CI @audrow !

hidmic:
Alright, thank you for contribution and your patience @sharminramli ! Merging.

PR #389, opened 2019-10-30
Provide tests for ros2param verbs

BMarchi:
As the title says, this PR provides a set of tests covering the usage of all the verbs for ros2param

### DIFF
--- a/ros2param/test/fixtures/param_node.py
+++ b/ros2param/test/fixtures/param_node.py
@@ -1,0 +1,56 @@
+# Copyright 2019 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import sys
+
+from rcl_interfaces.msg import ParameterDescriptor
+import rclpy
+from rclpy.node import Node
+
+
+class ParamNode(Node):
+
+    def __init__(self):
+        super().__init__('param_node')
+        self.declare_parameter('bool_param', True, ParameterDescriptor())
+        self.declare_parameter('int_param', 42, ParameterDescriptor())
+        self.declare_parameter('double_param', 1.23, ParameterDescriptor())
+        self.declare_parameter('str_param', 'Hello World', ParameterDescriptor())
+        self.declare_parameter('int_array', [1, 2, 3], ParameterDescriptor())
+        self.declare_parameter('double_array', [1.0, 2.0, 3.0], ParameterDescriptor())
+        self.declare_parameter('bool_array', [True, False, True], ParameterDescriptor())
+        self.declare_parameter('str_array', ['Hello', 'World'], ParameterDescriptor())
+        self.declare_parameter('byte_array', [b'p', b'v'], ParameterDescriptor())
+        self.declare_parameter('parameter_with_no_value', None, ParameterDescriptor())
+
+
+def main(args=None):
+    rclpy.init(args=args)
+
+    node = ParamNode()
+
+    try:
+        rclpy.spin(node)
+    except KeyboardInterrupt:
+        print('node stopped cleanly')
+    except BaseException:
+        print('exception in node:', file=sys.stderr)
+        raise
+    finally:
+        node.destroy_node()
+        rclpy.shutdown()
+
+
+if __name__ == '__main__':
+    main()
--- a/ros2param/test/test_cli.py
+++ b/ros2param/test/test_cli.py
@@ -1,0 +1,796 @@
+# Copyright 2019 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import contextlib
+import os
+import sys
+import unittest
+
+from launch import LaunchDescription
+from launch.actions import ExecuteProcess
+from launch.actions import OpaqueFunction
+
+from launch_ros.actions import Node
+
+import launch_testing
+import launch_testing.asserts
+import launch_testing.markers
+import launch_testing.tools
+import launch_testing_ros.tools
+
+import pytest
+
+from rmw_implementation import get_available_rmw_implementations
+
+PARAM_NODE_PARAMETER_LIST = """\
+  bool_array
+  bool_param
+  byte_array
+  double_array
+  double_param
+  int_array
+  int_param
+  parameter_with_no_value
+  str_array
+  str_param
+  use_sim_time
+"""
+
+PARAM_NODE_BOOL_PARAMETER_LIST = """\
+  bool_array
+  bool_param
+"""
+
+ALL_NODES_PARAMETER_LIST = """\
+/_hidden_param_node:
+  bool_array
+  bool_param
+  byte_array
+  double_array
+  double_param
+  int_array
+  int_param
+  parameter_with_no_value
+  str_array
+  str_param
+  use_sim_time
+/launch_ros:
+  use_sim_time
+/param_node:
+  bool_array
+  bool_param
+  byte_array
+  double_array
+  double_param
+  int_array
+  int_param
+  parameter_with_no_value
+  str_array
+  str_param
+  use_sim_time
+"""
+
+DESCRIBE_PARAMETER_TYPE = """\
+Parameter name: {}
+  Type: {}
+  Constraints:{}
+"""
+
+
+@pytest.mark.rostest
+@launch_testing.parametrize('rmw_implementation', get_available_rmw_implementations())
+def generate_test_description(rmw_implementation, ready_fn):
+    path_to_param_node_script = os.path.join(
+        os.path.dirname(__file__), 'fixtures', 'param_node.py'
+    )
+    additional_env = {'RMW_IMPLEMENTATION': rmw_implementation}
+    return LaunchDescription([
+        # Always restart daemon to isolate tests.
+        ExecuteProcess(
+            cmd=['ros2', 'daemon', 'stop'],
+            name='daemon-stop',
+            on_exit=[
+                ExecuteProcess(
+                    cmd=['ros2', 'daemon', 'start'],
+                    name='daemon-start',
+                    on_exit=[
+                        # Add test fixture actions.
+                        Node(
+                            node_executable=sys.executable,
+                            arguments=[path_to_param_node_script],
+                            node_name='param_node',
+                            additional_env=additional_env
+                        ),
+                        Node(
+                            node_executable=sys.executable,
+                            arguments=[path_to_param_node_script],
+                            node_name='_hidden_param_node',
+                            additional_env=additional_env
+                        ),
+                        OpaqueFunction(function=lambda context: ready_fn())
+                    ],
+                    additional_env=additional_env
+                )
+            ]
+        ),
+    ])
+
+
+class TestROS2ParamCLI(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(
+        cls,
+        launch_service,
+        proc_info,
+        proc_output,
+        rmw_implementation
+    ):
+        @contextlib.contextmanager
+        def launch_node_command(self, arguments):
+            node_command_action = ExecuteProcess(
+                cmd=['ros2', 'param', *arguments],
+                additional_env={
+                    'RMW_IMPLEMENTATION': rmw_implementation,
+                    'PYTHONUNBUFFERED': '1'
+                },
+                name='ros2param-cli',
+                output='screen'
+            )
+            with launch_testing.tools.launch_process(
+                launch_service, node_command_action, proc_info, proc_output,
+                output_filter=launch_testing_ros.tools.basic_output_filter(
+                    # ignore launch_ros and ros2cli daemon nodes
+                    filtered_patterns=['.*ros2cli.*'],
+                    filtered_rmw_implementation=rmw_implementation
+                )
+            ) as node_command:
+                yield node_command
+        cls.launch_node_command = launch_node_command
+
+    # Set verb tests
+    @launch_testing.markers.retry_on_failure(times=3)
+    def test_set_invalid_param_existing_node(self):
+        with self.launch_node_command(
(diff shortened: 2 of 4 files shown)


ros2param/test/test_cli.py
```diff
@@ -0,0 +1,718 @@
...
+                    'PYTHONUNBUFFERED': '1'
+                },
+                name='ros2param-cli',
+                output='screen'
+            )
+            with launch_testing.tools.launch_process(
+                launch_service, node_command_action, proc_info, proc_output,
+                output_filter=launch_testing_ros.tools.basic_output_filter(
+                    # ignore launch_ros and ros2cli daemon nodes
+                    filtered_patterns=['.*ros2cli.*'],
```
hidmic: @BMarchi are you ignoring `launch_ros`?

ros2param/test/test_cli.py
```diff
@@ -0,0 +1,718 @@
...
+                )
+            ) as node_command:
+                yield node_command
+        cls.launch_node_command = launch_node_command
+
+    # Set verb tests
+    @launch_testing.markers.retry_on_failure(times=3)
+    def test_set_invalid_param_existing_node(self):
+        with self.launch_node_command(
+                arguments=['set', '/param_node', 'unexistent_parameter', '1']) as node_command:
```
hidmic: @BMarchi nit: `nonexistent_parameter` or `invalid_parameter`?

ros2param/test/test_cli.py
```diff
@@ -0,0 +1,718 @@
...
+        assert node_command.exit_code == launch_testing.asserts.EXIT_OK
+        assert launch_testing.tools.expect_output(
+            expected_lines=[
+                "Setting parameter failed: ('Invalid access to undeclared parameter(s)', [])"],
+            text=node_command.output,
+            strict=True
+        )
+
+    @launch_testing.markers.retry_on_failure(times=3)
+    def test_set_valid_param_existing_node(self):
```
hidmic: @BMarchi I think we should probably test `ros2 param set` on each parameter type.

ros2param/test/test_cli.py
```diff
@@ -0,0 +1,718 @@
...
+        assert node_command.exit_code == launch_testing.asserts.EXIT_OK
+        assert launch_testing.tools.expect_output(
+            expected_lines=[
+                "Setting parameter failed: ('Invalid access to undeclared parameter(s)', [])"],
+            text=node_command.output,
+            strict=True
+        )
+
+    @launch_testing.markers.retry_on_failure(times=3)
+    def test_set_inexisting_node(self):
```
hidmic: @BMarchi nit: I think I'd normalize to `nonexistent`. `unexistent` seems to be valid but rare, and I can't find a convincing example of `inexisting` existence (no pun intended).

ros2param/test/test_cli.py
```diff
@@ -0,0 +1,718 @@
...
+    # Get verb tests without --hide-type flag
+    @launch_testing.markers.retry_on_failure(times=3)
+    def test_get_invalid_param_existing_node(self):
+        with self.launch_node_command(
+                arguments=['get', '/param_node', 'unexistent_parameter']) as node_command:
+            assert node_command.wait_for_shutdown(timeout=5)
+        assert node_command.exit_code == launch_testing.asserts.EXIT_OK
+        assert launch_testing.tools.expect_output(
+            expected_lines=[
+                'Parameter not set'],
```
hidmic: @BMarchi nit: odd formatting

ros2param/test/test_cli.py
```diff
@@ -0,0 +1,718 @@
...
+            expected_lines=[
+                "Setting parameter failed: ('Invalid access to undeclared parameter(s)', [])"],
+            text=node_command.output,
+            strict=True
+        )
+
+    @launch_testing.markers.retry_on_failure(times=3)
+    def test_set_valid_param_existing_node(self):
+        with self.launch_node_command(
+                arguments=['set', '/param_node', 'double_param', '3.14']) as node_command:
```
hidmic: @BMarchi nit: here and everywhere, odd formatting. Consider pushing the closing parenthesis one newline below.

ros2param/test/test_cli.py
```diff
@@ -0,0 +1,718 @@
...
+        )
+
+    @launch_testing.markers.retry_on_failure(times=3)
+    def test_get_valid_double_param_existing_node(self):
+        with self.launch_node_command(
+                arguments=['get', '/param_node', 'double_param']) as node_command:
+            assert node_command.wait_for_shutdown(timeout=5)
+        assert node_command.exit_code == launch_testing.asserts.EXIT_OK
+        assert launch_testing.tools.expect_output(
+            expected_lines=['Double value is: 1.23'],
```
hidmic: @BMarchi all of these tests use the same fixture node, and thus we have to be careful when setting and getting parameters. If `test_set_valid_param_existing_node` was run before `test_get_valid_double_param_existing_node`, this assertion would not succeed. It makes the test brittle. Consider avoiding that overlap or restoring fixture state after each test.

ros2param/test/test_cli.py
```diff
@@ -0,0 +1,718 @@
...
+            assert node_command.wait_for_shutdown(timeout=5)
+        assert node_command.exit_code == launch_testing.asserts.EXIT_OK
+        assert launch_testing.tools.expect_output(
+            expected_lines=['String value is: Hello World'],
+            text=node_command.output,
+            strict=True
+        )
+
+    @launch_testing.markers.retry_on_failure(times=3)
+    def test_get_valid_string_array_param_existing_node(self):
```
hidmic: @BMarchi meta: I wonder if we can parameterize these `ros2 param get` tests somehow, it's pretty much the same code except for `arguments` and `expected_lines`.

ros2param/test/test_cli.py
```diff
@@ -0,0 +1,718 @@
...
+    # def test_list_all_nodes_all_parameters(self):
+    #    with self.launch_node_command(
+    #            arguments=['list', '--include-hidden-nodes']) as node_command:
+    #        assert node_command.wait_for_shutdown(timeout=5)
+    #    assert node_command.exit_code == launch_testing.asserts.EXIT_OK
+    #    assert launch_testing.tools.expect_output(
+    #        expected_text=ALL_NODES_PARAMETER_LIST,
+    #        text=node_command.output,
+    #        strict=True
+    #    )
```
hidmic: @BMarchi if it only happens with Opensplice, consider checking `rmw_implementation` value and using [`skipTest` ](https://docs.python.org/3.6/library/unittest.html#unittest.TestCase.skipTest) appropriately instead of commenting out the entire test.

ros2param/test/fixtures/param_node.py
```diff
@@ -0,0 +1,56 @@
...
+        self.declare_parameter('bool_param', True, ParameterDescriptor())
+        self.declare_parameter('int_param', 42, ParameterDescriptor())
+        self.declare_parameter('double_param', 1.23, ParameterDescriptor())
+        self.declare_parameter('str_param', 'Hello World', ParameterDescriptor())
+        self.declare_parameter('int_array', [1, 2, 3], ParameterDescriptor())
+        self.declare_parameter('double_array', [1.0, 2.0, 3.0], ParameterDescriptor())
+        self.declare_parameter('bool_array', [True, False, True], ParameterDescriptor())
+        self.declare_parameter('str_array', ['Hello', 'World'], ParameterDescriptor())
+        self.declare_parameter('byte_array', [b'p', b'v'], ParameterDescriptor())
+        self.declare_parameter('parameter_with_no_value', None, ParameterDescriptor())
```
hidmic: @BMarchi hmm, I'm slightly inclined towards not keeping parameter values here and in the test. You could provide parameter value via overrides when you launch the node in the test itself.

ros2param/test/test_cli_delete.py
```diff
@@ -0,0 +1,188 @@
+# Copyright 2019 Open Source Robotics Foundation, Inc.
```
hidmic: @BMarchi I presume you needed a separate separate file because otherwise tests would affect each other. Consider doing so for the `set` command as well, but instead of duplicating the entire test infrastructure you could split it out in a separate module and reuse.

ros2param/test/test_cli.py
BMarchi: Typo

ros2param/test/test_cli.py
BMarchi: I was before, forgot to remove it

ros2param/test/fixtures/param_node.py
BMarchi: > and in the test

What do you mean? Since the node is for test only I don't see the problem of having the parameter declaration here

hidmic: IMHO it's better to have all value definitions in a single place instead of duplicating them across files and keeping them in sync. `launch` provides means to override parameters we could use.

ros2param/test/test_cli.py
BMarchi: Turns out it happens in every implementation, I can't remember if it worked before.

BMarchi: No, only launching the process

ros2param/test/test_cli_set.py
```diff
@@ -0,0 +1,190 @@
...
+                strict=True
+            )
+        for param_key, param_value in new_parameters.items():
+            with self.launch_node_command(
+                    arguments=['get', '/param_node', param_key]) as node_command:
+                assert node_command.wait_for_shutdown(timeout=5)
+            assert node_command.exit_code == launch_testing.asserts.EXIT_OK
+            if param_value is None:
+                param_node.GET_VERB_PARAM_VALUES[param_key] = \
+                    param_node.GET_VERB_PARAM_VALUES['str_param']
```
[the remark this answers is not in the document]
BMarchi: Because I don't want to duplicate the same dictionary, I'll change the name.

ros2param/test/param_node.py
```diff
@@ -0,0 +1,92 @@
...
+    'int_param': 'Integer value is: {}\n',
+    'double_param': 'Double value is: {}\n',
+    'str_param': 'String value is: {}\n',
+    'byte_array': 'Byte values are: {}\n',
+    'bool_array': 'Boolean values are: {}\n',
+    'int_array': "Integer values are: array('q', {})\n",
+    'double_array': "Double values are: array('d', {})\n",
+    'str_array': 'String values are: {}\n',
+    'parameter_with_no_value': 'Parameter not set.{}\n'
+}
```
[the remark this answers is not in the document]
BMarchi: To avoid duplications since I need it for the get and set verb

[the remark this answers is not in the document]
hidmic: Right, and that connects to https://github.com/ros2/ros2cli/pull/389#discussion_r342614281. Why not having specific instances, `GET_VERB_*` and `SET_VERB_*`, on each test? It doesn't look like we're re-using much and it obscures a bit the intent of the test.

ros2param/test/test_cli.py
hidmic: Hmm, interesting.

ros2param/test/test_cli.py
```diff
@@ -0,0 +1,365 @@
...
+                        arguments=['get', '/param_node',
+                                   param_key, '--hide-type']) as param_command:
+                    assert param_command.wait_for_shutdown(timeout=5)
+                assert param_command.exit_code == launch_testing.asserts.EXIT_OK
+                if param_value is None:
+                    param_value = ':None\n'
+                assert launch_testing.tools.expect_output(
+                    expected_lines=[
+                        GET_VERB_PARAM_VALUES[param_key].format(
+                            str(param_value)).split(':')[1].strip()],
```
hidmic: @BMarchi do we really need this piece of logic, as opposed to having yet another collection? IMHO it'd help readability (specially if we can drop line 217).

ros2param/test/test_cli_delete.py
```diff
@@ -0,0 +1,188 @@
...
+        )
+
+    @launch_testing.markers.retry_on_failure(times=3)
+    def test_delete_any_param_nonexistent_node(self):
+        with self.launch_param_command(
+                arguments=['delete', '/foo/nonexistent_node', 'int_param']) as param_command:
+            assert param_command.wait_for_shutdown(timeout=5)
+        assert param_command.exit_code == 1
+        assert launch_testing.tools.expect_output(
+            expected_text='Node not found\n',
```
hidmic: @BMarchi consider using `expected_lines=['Node not found']` instead (better not use newline literals).

ros2param/test/test_cli.py
```diff
@@ -0,0 +1,365 @@
...
+                )
+
+    @launch_testing.markers.retry_on_failure(times=3)
+    def test_get_verb_hide_type(self):
+        parameters = param_node.ParamNode.get_node_parameters()
+        for param_key, param_value in parameters.items():
+            with self.subTest(param_key=param_key, param_value=param_value):
+                with self.launch_param_command(
+                        arguments=['get', '/param_node',
+                                   param_key, '--hide-type']) as param_command:
```
hidmic: @BMarchi nit: consider normalizing format. On other places we use something like:

```python
with self.launch_some_command(
    arguments=[
        'get',
        # ...
    ]
) as some_command:
    # ...
```

I don't mind the format as long as we comply with the [style guide](https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#python) and we're consistent about it.

ros2param/test/test_cli.py
BMarchi: It seems I completely forgot about this one!

ros2param/test/test_cli.py
hidmic: This one still applies (throughout all tests I mean).

BMarchi:
Yep, lets hope it doesn't hang

BMarchi:
- Linux [Build Status](https://ci.ros2.org/job/ci_linux/8630/)
- Arch [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/4479/)
- Windows [Build Status](https://ci.ros2.org/job/ci_windows/8552/)
- OSX [Build Status](https://ci.ros2.org/job/ci_osx/7044/)

BMarchi:
Playing around with a Windows VM I was able to get all test running just fine **if and only if** I run them individually for each `rmw_implementation`, when I combine them (at least `fastrtps, cyclone and connext`), some tests hang and return exit code 1 after a sigterm is sent. I wasn't able yet to determine the root cause of this. Any thoughts?

claireyywang:
> Playing around with a Windows VM I was able to get all test running just fine **if and only if** I run them individually for each `rmw_implementation`, when I combine them (at least `fastrtps, cyclone and connext`), some tests hang and return exit code 1 after a sigterm is sent. I wasn't able yet to determine the root cause of this. Any thoughts?

I've encountered similar situation when running CI on another `ros2cli` PR on windows. It seems like there's no definite fix for it yet according to @nuclearsandwich 's investigation https://github.com/ros2/rclpy/pull/456 and https://github.com/ros2/build_cop/issues/248

nuclearsandwich:
> I've encountered similar situation when running CI on another `ros2cli` PR on windows. It seems like there's no definite fix for it yet according to @nuclearsandwich 's investigation [ros2/rclpy#456](https://github.com/ros2/rclpy/pull/456) and [ros2/build_cop#248](https://github.com/ros2/build_cop/issues/248)

The hangs in ros2/build_cop#248 were resolved by https://github.com/eProsima/Fast-RTPS/pull/871 and the hung processes were deadlocked requiring a SIGKILL so it is likely something else going on here.

ivanpauno:
@hidmic I assigned you, so it doesn't appear more in our triage search, I also understand you're planning to follow-up on this.

PR #390, opened 2019-11-01
Add tests for ros2component

BMarchi:
As the title says, this PR adds tests for the ros2component cli. It's lacking tests for the standalone verb and opensplice is returning a different code error that needs further investigation.
I wasn't able to separate all the verbs because the container is kept alive for every test case and I'm not sure if we can rerun the container each time a test is run.

### DIFF
--- a/ros2component/test/test_cli.py
+++ b/ros2component/test/test_cli.py
@@ -1,0 +1,268 @@
+# Copyright 2019 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import contextlib
+import unittest
+
+from launch import LaunchDescription
+from launch.actions import ExecuteProcess
+from launch.actions import OpaqueFunction
+
+from launch_ros.actions import Node
+
+import launch_testing
+import launch_testing.asserts
+import launch_testing.markers
+import launch_testing.tools
+import launch_testing_ros.tools
+
+import pytest
+
+from rmw_implementation import get_available_rmw_implementations
+
+DEMO_NODES_TYPES = """\
+demo_nodes_cpp
+  demo_nodes_cpp::OneOffTimerNode
+  demo_nodes_cpp::ReuseTimerNode
+  demo_nodes_cpp::ServerNode
+  demo_nodes_cpp::ClientNode
+  demo_nodes_cpp::ListParameters
+  demo_nodes_cpp::ParameterBlackboard
+  demo_nodes_cpp::SetAndGetParameters
+  demo_nodes_cpp::ParameterEventsAsyncNode
+  demo_nodes_cpp::EvenParameterNode
+  demo_nodes_cpp::Talker
+  demo_nodes_cpp::LoanedMessageTalker
+  demo_nodes_cpp::SerializedMessageTalker
+  demo_nodes_cpp::Listener
+  demo_nodes_cpp::SerializedMessageListener
+  demo_nodes_cpp::ListenerBestEffort
+"""
+
+LIST_THREE_NODES = """\
+/ComponentManager
+  2  /listener
+  3  /talker
+  4  /talker
+"""
+
+LIST_FOUR_NODES = """\
+/ComponentManager
+  2  /listener
+  3  /talker
+  4  /talker
+  5  /talker
+"""
+
+
+@pytest.mark.rostest
+@launch_testing.parametrize('rmw_implementation', get_available_rmw_implementations())
+def generate_test_description(rmw_implementation, ready_fn):
+    additional_env = {'RMW_IMPLEMENTATION': rmw_implementation}
+    component_node = Node(
+        package='rclcpp_components', node_executable='component_container', output='screen')
+    return LaunchDescription([
+        # Always restart daemon to isolate tests.
+        ExecuteProcess(
+            cmd=['ros2', 'daemon', 'stop'],
+            name='daemon-stop',
+            on_exit=[
+                ExecuteProcess(
+                    cmd=['ros2', 'daemon', 'start'],
+                    name='daemon-start',
+                    on_exit=[
+                        component_node,
+                        OpaqueFunction(function=lambda context: ready_fn())
+                    ],
+                    additional_env=additional_env
+                )
+            ]
+        ),
+    ]), locals()
+
+# TODO(BMarchi): Open splice is returning a different code (2) when
+# the launch of a component verb ends. It should return EXIT_OK unless
+# the command returns a error string, which should be code 1.
+class TestROS2ComponentCLI(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(
+        cls,
+        launch_service,
+        proc_info,
+        proc_output,
+        rmw_implementation
+    ):
+        @contextlib.contextmanager
+        def launch_node_command(self, arguments):
+            node_command_action = ExecuteProcess(
+                cmd=['ros2', 'component', *arguments],
+                additional_env={
+                    'RMW_IMPLEMENTATION': rmw_implementation,
+                    'PYTHONUNBUFFERED': '1'
+                },
+                name='ros2component-cli',
+                output='screen'
+            )
+            with launch_testing.tools.launch_process(
+                launch_service, node_command_action, proc_info, proc_output,
+                output_filter=launch_testing_ros.tools.basic_output_filter(
+                    # ignore ros2cli daemon nodes
+                    filtered_patterns=['.*ros2cli.*'],
+                    filtered_rmw_implementation=rmw_implementation
+                )
+            ) as node_command:
+                yield node_command
+        cls.launch_node_command = launch_node_command
+
+    # Set verb tests
+    @launch_testing.markers.retry_on_failure(times=3)
+    def test_types_component(self):
+        with self.launch_node_command(
+                arguments=['types']) as node_command:
+            assert node_command.wait_for_shutdown(timeout=20)
+        assert node_command.exit_code == launch_testing.asserts.EXIT_OK
+        assert launch_testing.tools.expect_output(
+            expected_text=DEMO_NODES_TYPES,
+            text=node_command.output,
+            strict=True
+        )
+
+    @launch_testing.markers.retry_on_failure(times=2)
+    def test_load_unload_verb(self):
+        with self.launch_node_command(
+                arguments=[
+                    'load', '/ComponentManager',
+                    'demo_nodes_cpp', 'demo_nodes_cpp::Talker']) as talker_node:
+            assert talker_node.wait_for_shutdown(timeout=20)
+        assert talker_node.exit_code == launch_testing.asserts.EXIT_OK
+        assert launch_testing.tools.expect_output(
+            expected_lines=[
+                "Loaded component 1 into '/ComponentManager' "
+                "container node as '/talker'"],
+            text=talker_node.output,
+            strict=True
+        )
+        with self.launch_node_command(
+                arguments=[
+                    'load', '/ComponentManager',
+                    'demo_nodes_cpp', 'demo_nodes_cpp::Listener']) as listener_node:
+            assert listener_node.wait_for_shutdown(timeout=20)
+        assert listener_node.exit_code == launch_testing.asserts.EXIT_OK
+        assert launch_testing.tools.expect_output(
+            expected_lines=[
+                "Loaded component 2 into '/ComponentManager' "
+                "container node as '/listener'"],
+            text=listener_node.output,
+            strict=True
+        )
+        with self.launch_node_command(
+                arguments=[
+                    'unload', '/ComponentManager', '1']) as unload_command:
+            assert unload_command.wait_for_shutdown(timeout=20)
+        assert unload_command.exit_code == launch_testing.asserts.EXIT_OK
+        assert launch_testing.tools.expect_output(
+            expected_lines=["Unloaded component 1 from '/ComponentManager' container"],
+            text=unload_command.output,
+            strict=True
+        )
+        # Test the unique id for loaded nodes.
+        with self.launch_node_command(
+                arguments=[
+                    'load', '/ComponentManager',
+                    'demo_nodes_cpp', 'demo_nodes_cpp::Talker']) as talker_node:
+            assert talker_node.wait_for_shutdown(timeout=20)
+        assert talker_node.exit_code == launch_testing.asserts.EXIT_OK
+        assert launch_testing.tools.expect_output(
+            expected_lines=[
+                "Loaded component 3 into '/ComponentManager' "
+                "container node as '/talker'"],
+            text=talker_node.output,
+            strict=True
+        )
+        # Test we can load the same node more than once.
+        with self.launch_node_command(
+                arguments=[
+                    'load', '/ComponentManager',
+                    'demo_nodes_cpp', 'demo_nodes_cpp::Talker']) as talker_node:
+            assert talker_node.wait_for_shutdown(timeout=20)
+        assert talker_node.exit_code == launch_testing.asserts.EXIT_OK
+        assert launch_testing.tools.expect_output(
+            expected_lines=[
+                "Loaded component 4 into '/ComponentManager' "
+                "container node as '/talker'"],
+            text=talker_node.output,
+            strict=True
+        )
+        with self.launch_node_command(
+                arguments=['list']) as list_command:
+            assert list_command.wait_for_shutdown(timeout=20)
+        assert list_command.exit_code == launch_testing.asserts.EXIT_OK
+        assert launch_testing.tools.expect_output(
+            expected_text=LIST_THREE_NODES,
+            text=list_command.output,
+            strict=True
+        )
+
+        # Unexisting class in existing plugin
+        with self.launch_node_command(
+                arguments=[
+                    'load', '/ComponentManager',
+                    'demo_nodes_cpp', 'demo_nodes_cpp::NonExistingPlugin']) as command:
+            assert command.wait_for_shutdown(timeout=20)
+        assert command.exit_code == 1
+        assert launch_testing.tools.expect_output(
+            expected_lines=[
+                'Failed to load component: '
+                'Failed to find class with the requested plugin name.'],
+            text=command.output,
+            strict=True
+        )
+        # Test unexisting plugin
+        with self.launch_node_command(
+                arguments=[
+                    'load', '/ComponentManager',
+                    'non_existing_plugin', 'non_existing_plugin::Test']) as command:
+            assert command.wait_for_shutdown(timeout=20)
+        assert command.exit_code == 1
+        assert launch_testing.tools.expect_output(
+            expected_lines=[
+                'Failed to load component: '
+                'Could not find requested resource in ament index'],
+            text=command.output,
+            strict=True
+        )
+        with self.launch_node_command(
+                arguments=[
+                    'load', '/ComponentManager',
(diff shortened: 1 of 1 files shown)


BMarchi:
Something to notice is that these tests don't run with opensplice. It always hangs for almost all verb conditions which ends up failing every test. I suggest to create an issue for opensplice only.

ros2component/package.xml
```diff
@@ -9,12 +9,13 @@
   <maintainer email="michel@ekumenlabs.com">Michel Hidalgo</maintainer>
   <license>Apache License 2.0</license>
 
+  <depend>ros2cli</depend>
```
hidmic: @BMarchi why this change?

ros2component/test/test_cli_list.py
```diff
@@ -0,0 +1,150 @@
...
+                cmd=['ros2', 'component', 'load', '/ComponentManager',
+                     'ros2component_fixtures', 'ros2component_fixtures::Talker'],
+                additional_env={
+                    'RMW_IMPLEMENTATION': rmw_implementation,
+                    'PYTHONUNBUFFERED': '1'
+                },
+                name='ros2component-cli',
+                output='screen',
+                on_exit=[
+                    ExecuteProcess(
```
hidmic: @BMarchi why nesting actions instead of putting one after the other?

ros2component/test/test_cli_list.py
```diff
@@ -0,0 +1,150 @@
...
+        ExecuteProcess(
+            cmd=['ros2', 'daemon', 'stop'],
+            name='daemon-stop',
+            on_exit=[
+                ExecuteProcess(
+                    cmd=['ros2', 'daemon', 'start'],
+                    name='daemon-start',
+                    on_exit=[
+                        component_node,
+                        OpaqueFunction(function=lambda context: ready_fn()),
```
hidmic: @BMarchi why isn't this the last action being called? Also, consider using `launch_testing.actions.ReadyToTest` instead (use of `ready_fn` is due for deprecation, eventually).

ros2component/test/test_cli_list.py
```diff
@@ -0,0 +1,150 @@
...
+            with launch_testing.tools.launch_process(
+                launch_service, node_command_action, proc_info, proc_output,
+                output_filter=launch_testing_ros.tools.basic_output_filter(
+                    # ignore ros2cli daemon nodes
+                    filtered_patterns=['.*ros2cli.*'],
+                    filtered_rmw_implementation=rmw_implementation
+                )
+            ) as node_command:
+                yield node_command
+        cls.launch_node_command = launch_node_command
```
hidmic: @BMarchi here and elsewhere, `*node_*` -> `*component_*`?

ros2component/test/test_cli_list.py
```diff
@@ -0,0 +1,150 @@
...
+        cls.rmw_implementation = rmw_implementation
```
hidmic: @BMarchi unused?

ros2component/test/test_cli_load.py
```diff
@@ -0,0 +1,196 @@
...
+            ) as node_command:
+                yield node_command
+        cls.launch_node_command = launch_node_command
+
+    @launch_testing.markers.retry_on_failure(times=2)
+    def test_load_verb(self):
+        with self.launch_node_command(
+                arguments=[
+                    'load', '/ComponentManager',
+                    'ros2component_fixtures', 'ros2component_fixtures::Talker']) as talker_node:
```
hidmic: @BMarchi nit: here and elsewhere, this isn't a `talker_node` itself, but a `load_command` at most.

ros2component/test/test_cli_load.py
```diff
@@ -0,0 +1,196 @@
...
+        assert launch_testing.tools.expect_output(
+            expected_lines=[
+                "Loaded component 4 into '/ComponentManager' "
+                "container node as '/talker'"],
+            text=talker_node.output,
+            strict=True
+        )
+
+    @launch_testing.markers.retry_on_failure(times=2)
+    def test_load_verb_nonexistent_class(self):
```
hidmic: @BMarchi nit: `test_load_nonexistent_component`

ros2component/test/test_cli_load.py
```diff
@@ -0,0 +1,196 @@
...
+        with self.launch_node_command(
+                arguments=[
+                    'load', '/ComponentManager',
+                    'ros2component_fixtures',
+                    'ros2component_fixtures::NonExistingPlugin']) as command:
```
hidmic: @BMarchi nit: `'ros2component_fixtures::NonExistentComponent'`

ros2component/test/test_cli_load.py
```diff
@@ -0,0 +1,196 @@
...
+        assert launch_testing.tools.expect_output(
+            expected_lines=[
+                'Failed to load component: '
+                'Failed to find class with the requested plugin name.'],
+            text=command.output,
+            strict=True
+        )
+
+    @launch_testing.markers.retry_on_failure(times=2)
+    def test_load_verb_nonexistent_plugin(self):
```
hidmic: @BMarchi same as above.

ros2component/test/test_cli_types.py
```diff
@@ -0,0 +1,110 @@
...
+
+import pytest
+
+from rmw_implementation import get_available_rmw_implementations
+
+DEMO_NODES_TYPES = """\
+ros2component_fixtures
+  ros2component_fixtures::Talker
+  ros2component_fixtures::Listener
+"""
```
hidmic: @BMarchi this "constant" is used once, consider instantiating it in place (i.e. where it is used) as a list of lines.

ros2component/test/test_cli_list.py
```diff
@@ -0,0 +1,150 @@
...
+import pytest
+
+from rmw_implementation import get_available_rmw_implementations
+
+LIST_THREE_NODES = """\
+/ComponentManager
+  1  /listener
+  2  /talker
+  3  /talker
+"""
```
hidmic: @BMarchi this is used in a single place, consider just instantiating it there as a list of lines.

ros2component/test/test_cli_load.py
```diff
@@ -0,0 +1,196 @@
...
+            assert listener_node.wait_for_shutdown(timeout=20)
+        assert listener_node.exit_code == launch_testing.asserts.EXIT_OK
+        assert launch_testing.tools.expect_output(
+            expected_lines=[
+                "Loaded component 2 into '/ComponentManager' "
+                "container node as '/listener'"],
+            text=listener_node.output,
+            strict=True
+        )
+        with self.launch_node_command(
```
hidmic: @BMarchi I think it'd be good to assert on container output to make sure components were indeed loaded.

ros2component_fixtures/CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
...
+find_package(rclcpp_components REQUIRED)
+find_package(std_msgs REQUIRED)
+
+include_directories(include)
+
+# create ament index resource which references the libraries in the binary dir
+set(node_plugins "")
+
+add_library(talker_component SHARED
+  src/talker_component.cpp)
```
hidmic: @BMarchi do we need a full talker? Would it be the same if we had a dummy component that simply logs something on construction and something else on destruction?

ros2component_fixtures/CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
...
+  PRIVATE "ROS2COMPONENT_FIXTURES_BUILDING_DLL")
+ament_target_dependencies(talker_component
+  "rclcpp"
+  "rclcpp_components"
+  "std_msgs")
+rclcpp_components_register_nodes(talker_component "ros2component_fixtures::Talker")
+set(node_plugins "${node_plugins}ros2component_fixtures::Talker;$<TARGET_FILE:talker_component>\n")
+
+add_library(listener_component SHARED
+  src/listener_component.cpp)
```
hidmic: @BMarchi do we need a second component at all to test the CLI?

ros2component/test/test_cli_list.py
BMarchi: For some reason the opaque function is working as expected meanwhile ReadyToTest throws me a time out for process launching.

ros2component/test/test_cli_load.py
BMarchi: Same what? I can't call things the same

ros2component/test/test_cli_load.py
BMarchi: That implies that I'll have to deal with discovery delay and related stuff

BMarchi: And to check that I'll need to get the output from the container that I'm loading in `generate_test_description` process.

ros2component_fixtures/CMakeLists.txt
BMarchi: I prefer to have a second one with a different name rather than some weird stuff that could happen using just one component.

ros2component/test/test_cli_list.py
BMarchi: It seems that not always the listener gets instantiated after the talkers even if I put them one after the other so I need to ensure this order for proper printing.

ros2component_fixtures/CMakeLists.txt
hidmic: What weird stuff?

ros2component/test/test_cli_load.py
hidmic: > This should only test the cli, not if the node is working as it should, that should be tested on its own place.

I fully agree, but that applies to both talker and/or listener components (and thus my comment, https://github.com/ros2/ros2cli/pull/390#discussion_r343117111). Checking the container output, assuming we can trust it (and that's being tested elsewhere), is the simplest way to make sure that `load` and `unload` actually worked. Otherwise you're not checking if the CLI has the expected side effect.

ros2component/test/test_cli_list.py
hidmic: That's unexpected. `ReadyToTest` should be a drop-in replacement. Can you confirm it isn't? If so, open an issue in `launch` with a small repro.

ros2component_fixtures/CMakeLists.txt
BMarchi: I don't know, for some reason the cli could clone nodes or it's ignoring new ones, for me is better to be prepared for the worst and I don't see anything bad of having just one extra node.

ros2component/test/test_cli_list.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+@pytest.mark.rostest
+@launch_testing.parametrize('rmw_implementation', get_available_rmw_implementations())
+def generate_test_description(rmw_implementation, ready_fn):
+    additional_env = {'RMW_IMPLEMENTATION': rmw_implementation}
+    component_node = Node(
+        package='rclcpp_components', node_executable='component_container', output='screen',
+        additional_env=additional_env)
+    listener_command_action = ExecuteProcess(
```
hidmic: @BMarchi nit^2: could you flatten action definitions? Like 

```python
some_action = ExecuteProcess(...)
another_action =  ExecuteProcess(..., on_exit=[some_action])
...
```

ros2component/test/test_cli_list.py
```diff
@@ -0,0 +1,145 @@
...
+import pytest
+
+from rmw_implementation import get_available_rmw_implementations
+
+
+@pytest.mark.rostest
+@launch_testing.parametrize('rmw_implementation', get_available_rmw_implementations())
+def generate_test_description(rmw_implementation, ready_fn):
+    additional_env = {'RMW_IMPLEMENTATION': rmw_implementation}
+    component_node = Node(
```
hidmic: @BMarchi nit: consider `container_node_action` instead

ros2component/test/test_cli_list.py
```diff
@@ -0,0 +1,145 @@
...
+                )
+            ) as component_command:
+                yield component_command
+        cls.launch_component_command = launch_component_command
+
+    @launch_testing.markers.retry_on_failure(times=10)
+    def test_list_verb(self):
+        with self.launch_component_command(
+                arguments=['list']) as list_command:
+            assert list_command.wait_for_shutdown(timeout=30)
```
hidmic: @BMarchi wow, is this timeout arbitrarily long or did you really need 30 sec?

ros2component/test/test_cli_load.py
```diff
@@ -0,0 +1,217 @@
...
+import pytest
+
+from rmw_implementation import get_available_rmw_implementations
+
+
+@pytest.mark.rostest
+@launch_testing.parametrize('rmw_implementation', get_available_rmw_implementations())
+def generate_test_description(rmw_implementation, ready_fn):
+    additional_env = {'RMW_IMPLEMENTATION': rmw_implementation}
+    component_node = Node(
```
hidmic: @BMarchi same about naming here

ros2component/test/test_cli_load.py
```diff
@@ -0,0 +1,217 @@
...
+                    'ros2component_test_fixtures',
+                    'ros2component_test_fixtures::Talker']) as load_command:
+            assert load_command.wait_for_shutdown(timeout=20)
+            assert self.component_node.wait_for_output(functools.partial(
+                launch_testing.tools.expect_output, expected_lines=[
+                    "[INFO] [talker]: Publishing: 'Hello World'"
+                ], strict=False
+            ), timeout=10)
+        assert load_command.exit_code == launch_testing.asserts.EXIT_OK
+        assert launch_testing.tools.expect_output(
```
hidmic: @BMarchi why not testing this right after `wait_for_shutdown()`?

ros2component/test/test_cli_load.py
```diff
@@ -0,0 +1,217 @@
...
+    def test_load_verb(self):
+        with self.launch_component_command(
+                arguments=[
+                    'load', '/ComponentManager',
+                    'ros2component_test_fixtures',
+                    'ros2component_test_fixtures::Talker']) as load_command:
+            assert load_command.wait_for_shutdown(timeout=20)
+            assert self.component_node.wait_for_output(functools.partial(
+                launch_testing.tools.expect_output, expected_lines=[
+                    "[INFO] [talker]: Publishing: 'Hello World'"
```
hidmic: @BMarchi consider asserting container output instead of talker output.

ros2component/test/test_cli_load.py
```diff
@@ -0,0 +1,217 @@
...
+        # Test we can load the same node more than once.
+        with self.launch_component_command(
+                arguments=[
+                    'load', '/ComponentManager',
+                    'ros2component_test_fixtures',
+                    'ros2component_test_fixtures::Talker']) as load_command:
+            assert load_command.wait_for_shutdown(timeout=20)
+            assert self.component_node.wait_for_output(functools.partial(
+                launch_testing.tools.expect_output, expected_lines=[
+                    "[INFO] [talker]: Publishing: 'Hello World'"
```
hidmic: @BMarchi same about checking container output, not the component's

ros2component/test/test_cli_types.py
```diff
@@ -0,0 +1,110 @@
...
+from rmw_implementation import get_available_rmw_implementations
+
+
+@pytest.mark.rostest
+@launch_testing.parametrize('rmw_implementation', get_available_rmw_implementations())
+def generate_test_description(rmw_implementation, ready_fn):
+    additional_env = {'RMW_IMPLEMENTATION': rmw_implementation}
+    component_node = Node(
+        package='rclcpp_components', node_executable='component_container', output='screen',
+        additional_env=additional_env)
```
hidmic: @BMarchi I think we don't need a container up to test `ros2 component types`. Use `@launch_testing.markers.keep_alive` instead.

ros2component/test/test_cli_types.py
```diff
@@ -0,0 +1,110 @@
...
+                name='ros2component-cli',
+                output='screen'
+            )
+            with launch_testing.tools.launch_process(
+                launch_service, component_command_action, proc_info, proc_output,
+                output_filter=launch_testing_ros.tools.basic_output_filter(
+                    # ignore ros2cli daemon nodes
+                    filtered_patterns=['.*ros2cli.*'],
+                    filtered_rmw_implementation=rmw_implementation
+                )
```
hidmic: @BMarchi I think this output filter is not necessary

ros2component/test/test_cli_types.py
```diff
@@ -0,0 +1,110 @@
...
+    def test_types_verb(self):
+        with self.launch_component_command(
+                arguments=['types']) as component_command:
+            assert component_command.wait_for_shutdown(timeout=20)
+        assert component_command.exit_code == launch_testing.asserts.EXIT_OK
+        DEMO_NODES_TYPES = [
+            'ros2component_test_fixtures',
+            '  ros2component_test_fixtures::Talker',
+            '  ros2component_test_fixtures::Listener'
+        ]
```
hidmic: @BMarchi why defining a separate variable? Isn't it single use? You could simply set the list literal as an argument

ros2component/test/test_cli_unload.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+@pytest.mark.rostest
+@launch_testing.parametrize('rmw_implementation', get_available_rmw_implementations())
+def generate_test_description(rmw_implementation, ready_fn):
+    additional_env = {'RMW_IMPLEMENTATION': rmw_implementation}
+    component_node = Node(
+        package='rclcpp_components', node_executable='component_container', output='screen',
+        additional_env=additional_env)
+    listener_command_action = ExecuteProcess(
```
hidmic: @BMarchi same about naming and nesting here.

ros2component_test_fixtures/src/talker_component.cpp
```diff
@@ -0,0 +1,56 @@
...
+  pub_ = create_publisher<std_msgs::msg::String>("chatter", 10);
+
+  auto msg = std::make_unique<std_msgs::msg::String>();
+  msg->data = "Hello World";
+  RCLCPP_INFO(this->get_logger(), "Publishing: '%s'", msg->data.c_str());
+  std::flush(std::cout);
+
+  // Put the message into a queue to be processed by the middleware.
+  // This call is non-blocking.
+  pub_->publish(std::move(msg));
```
hidmic: @BMarchi if you check container output, you don't even have to publish.

ros2component/test/test_cli_list.py
BMarchi: I think I needed more retries rather than long timeout

ros2component/test/test_cli_load.py
BMarchi: The component node is the container

ros2component/test/test_cli_load.py
BMarchi: This comes from the container instead of the talker, everything prints out in the process where the container is run

ros2component_test_fixtures/src/talker_component.cpp
BMarchi: I can't see why I don't have to publish, can you explain? The container process is the one who gets the output

hidmic: I mean you can simply check the output the container generates on component load, no need for the component to generate extra output or even have any logic beyond these tests' purposes.

ros2component/test/test_cli_load.py
hidmic: I mean, check the component container's node output, not the talker node's output.

BMarchi:
Currently I'm dealing with the `standalone` verb extension. There's an issue with `launch` regarding the process that runs the standalone verb in `opensplice` ignores the `sigint` signal, which escalates to sigterm and leaves the process alive. For fastrtps I had to resend a sigint signal in the except portion [here](https://github.com/ros2/ros2cli/blob/f067c791d679a1feec6d5a04231952e34663eb22/ros2component/ros2component/verb/standalone.py#L60-L67) so it can kill the process properly. I'm still investigating this behavior.

BMarchi:
I wasn't able to find out the reason of why `standalone` tests are finishing without  killing the process launched. For `FastRTPS` I was able to wipe out the processes by re sending the `SIGINT` signal [here](https://github.com/ros2/ros2cli/blob/6cd4868c36df2ade9465b09ec82383a30a19ec82/ros2component/ros2component/verb/standalone.py#L63-L66) when a keyboard interrupt ocurrs, making sure that the process is still alive before sending it again. But for opensplice, the process is always alive (it's ignoring the `SIGINT` signal completely).

PR #398, opened 2019-11-13
Fix ros2 topic pub --node-name

sloretz:
`ros2 topic pub` has a `--node-name` option, but it appears to not do anything. This PR fixes that by making the direct_node take a keyword argument `node_name`.

To reproduce:

This should publish a string message on the topic `/foo/asdf`

```
ros2 topic pub -n "foo" "~/asdf" std_msgs/msg/String "data: be"
```

Without this PR, it appears the node name was not changed.

```
$ ros2 topic list --include-hidden-topics
/_ros2cli_982/asdf
```

### DIFF
--- a/ros2cli/ros2cli/node/direct.py
+++ b/ros2cli/ros2cli/node/direct.py
@@ -35,8 +35,13 @@ class DirectNode:
             args, 'node_name_suffix', '_%d' % os.getpid())
         start_parameter_services = getattr(
             args, 'start_parameter_services', False)
+
+        node_name = NODE_NAME_PREFIX + node_name_suffix
+        if hasattr(args, 'node_name') and args.node_name:
+            node_name = args.node_name
+
         self.node = rclpy.create_node(
-            NODE_NAME_PREFIX + node_name_suffix,
+            node_name,
             start_parameter_services=start_parameter_services)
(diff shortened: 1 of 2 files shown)


ros2cli/ros2cli/node/direct.py
```diff
@@ -35,8 +35,12 @@ def timer_callback():
             args, 'node_name_suffix', '_%d' % os.getpid())
         start_parameter_services = getattr(
             args, 'start_parameter_services', False)
+
+        if not node_name:
```
dirk-thomas: How about checking explicitly for `node_name is None`?

sloretz: Sounds good to me, `node_name is None` in 048791c

sloretz:
CI (testing ros2cli and ros2topic)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8642/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4485/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7049/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8561/)

PR #399, opened 2019-11-15
fix new linter warnings as of flake8-comprehensions 3.1.0

dirk-thomas:
Address new linter warnings.

It didn't make sense that the functions were already sorting their return value since in several cases the caller doesn't care about the order. Therefore I moved the sorting to the callers where applicable.


dirk-thomas:
* Linux CI: [Build Status](https://ci.ros2.org/job/ci_linux/8650/)

PR #407, opened 2019-11-25
Remove ros2msg and ros2srv

jacobperron:
Both packages were deprecated in Eloquent, superceded by ros2interface.


jacobperron:
I'm pretty sure there's no places left in our core packages referencing these packages, so I'm inclined to skip CI.

jacobperron:
@hidmic Correct me if I'm mistaken, but it looks like the CI job failures are all present in the latest nightly.

jacobperron:
I'm fairly certain none of the test failures are due to the removal of ros2msg and ros2srv. If I overlooked something, I'll fix it in post.

PR #392, opened 2019-11-07
add service clients to ros2node info

mikaelarguedas:
Currently `ros2 node info` shows only service servers:
```
$ ros2 run demo_nodes_py add_two_ints_client &
$ ros2 node info /add_two_ints_client
/add_two_ints_client
  Subscribers:

  Publishers:
    /parameter_events: rcl_interfaces/msg/ParameterEvent
    /rosout: rcl_interfaces/msg/Log
  Services:
    /add_two_ints_client/describe_parameters: rcl_interfaces/srv/DescribeParameters
    /add_two_ints_client/get_parameter_types: rcl_interfaces/srv/GetParameterTypes
    /add_two_ints_client/get_parameters: rcl_interfaces/srv/GetParameters
    /add_two_ints_client/list_parameters: rcl_interfaces/srv/ListParameters
    /add_two_ints_client/set_parameters: rcl_interfaces/srv/SetParameters
    /add_two_ints_client/set_parameters_atomically: rcl_interfaces/srv/SetParametersAtomically
  Action Servers:

  Action Clients:


```

After this PR:
```
$ ros2 run demo_nodes_py add_two_ints_client &
$ ros2 node info /add_two_ints_client
$ ros2 node info /add_two_ints_client
/add_two_ints_client
  Subscribers:

  Publishers:
    /parameter_events: rcl_interfaces/msg/ParameterEvent
    /rosout: rcl_interfaces/msg/Log
  Service Servers:
    /add_two_ints_client/describe_parameters: rcl_interfaces/srv/DescribeParameters
    /add_two_ints_client/get_parameter_types: rcl_interfaces/srv/GetParameterTypes
    /add_two_ints_client/get_parameters: rcl_interfaces/srv/GetParameters
    /add_two_ints_client/list_parameters: rcl_interfaces/srv/ListParameters
    /add_two_ints_client/set_parameters: rcl_interfaces/srv/SetParameters
    /add_two_ints_client/set_parameters_atomically: rcl_interfaces/srv/SetParametersAtomically
  Service Clients:
    /add_two_ints: example_interfaces/srv/AddTwoInts
  Action Servers:

  Action Clients:

```

Signed-off-by: Mikael Arguedas <mikael.arguedas@gmail.com>

PR #408, opened 2019-11-26
[dashing] add get_service_server_info to provide same API as Eloquent

dirk-thomas:
Since #392 broke the API on Eloquent this patch adds the new API to Dashing to allow packages to move forward without breaking Dashing.


dirk-thomas:
> It would be reasonable to backport #392 completely, without deleting `get_service_info` (i.e. left it as an alias of `get_service_server_info`).

Sounds good to me. Closing this in favor of #409.

PR #409, opened 2019-11-27
[dashing backport] add service clients to ros2node info

dirk-thomas:
Backport of #392.

### DIFF
--- a/ros2node/ros2node/verb/info.py
+++ b/ros2node/ros2node/verb/info.py
@@ -17,7 +17,8 @@ from ros2cli.node.strategy import add_arguments
 from ros2cli.node.strategy import NodeStrategy
 from ros2node.api import get_node_names
 from ros2node.api import get_publisher_info
-from ros2node.api import get_service_info
+from ros2node.api import get_service_client_info
+from ros2node.api import get_service_server_info
 from ros2node.api import get_subscriber_info
 from ros2node.api import NodeNameCompleter
 from ros2node.verb import VerbExtension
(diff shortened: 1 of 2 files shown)


dirk-thomas:
> (with green CI)

The referenced PR jobs is already green?

ivanpauno:
> The referenced PR job is already green?

Don't we run again CI against `dashing` branch?
IMO, it's not the same thing.
It's pretty unlikely that this will introduce a failure, as there is not much code in the Dashing branch testing the CLI (except the api, which this is modifying trivially).
So, I don't mind if the re-run is ignored.

dirk-thomas:
> Don't we run again CI against `dashing` branch?

I would say it always depends on the kind of patch. Since this one doesn't have any additional tests and no platform specific logic and passes the PR build I don't see a need to run the CI build on all platforms.

> there is not much code in the Dashing branch testing the CLI (except the api, which this is modifying trivially).

Yeah, the API update is trivial and there is no test covering it on the Dashing branch anyway.

ros2node/ros2node/verb/info.py
```diff
@@ -49,8 +50,13 @@ def main(self, *, args):
                 publishers = get_publisher_info(node=node, remote_node_name=args.node_name)
                 print('  Publishers:')
                 print_names_and_types(publishers)
-                services = get_service_info(node=node, remote_node_name=args.node_name)
-                print('  Services:')
-                print_names_and_types(services)
+                service_servers = get_service_server_info(
+                    node=node, remote_node_name=args.node_name, include_hidden=args.include_hidden)
```
dirk-thomas: On the `dashing` branch the `info` verb doesn't have the `--include-hidden` option. See #417.

PR #410, opened 2019-11-27
[WIP] Added incompatible event support to ros2 topic echo and ros2 topic pub

jaisontj:
**_Depends on [ros2/rclpy:459](https://github.com/ros2/rclpy/pull/459)_**

Related to the feature discussed [here](https://github.com/ros2/ros2/issues/822). This PR handles the `ON_REQUESTED_INCOMPATIBLE_QOS` and `ON_OFFERED_INCOMPATIBLE_QOS` callback events on `ros2 topic echo` and `ros2 topic pub` commands respectively. 

This way, users now will get a warning when they `echo/pub` to a topic with incompatible QoS Policies.

Signed-off-by: Jaison Titus <jaisontj92@gmail.com>

### DIFF
--- a/ros2topic/ros2topic/verb/info.py
+++ b/ros2topic/ros2topic/verb/info.py
@@ -61,7 +61,7 @@ class InfoVerb(VerbExtension):
                 except NotImplementedError as e:
                     return str(e)
 
-            print('Subscription count: %d' % node.count_subscribers(topic_name), end=line_end)
+            print('Subscription count: %d' % node.count_subscribers(topic_name))
             if args.verbose:
                 try:
                     for info in node.get_subscriptions_info_by_topic(topic_name):
--- a/ros2topic/ros2topic/verb/info.py
+++ b/ros2topic/ros2topic/verb/info.py
@@ -61,7 +61,7 @@ class InfoVerb(VerbExtension):
                 except NotImplementedError as e:
                     return str(e)
 
-            print('Subscription count: %d' % node.count_subscribers(topic_name))
+            print('Subscription count: %d' % node.count_subscribers(topic_name), end=line_end)
             if args.verbose:
                 try:
                     for info in node.get_subscriptions_info_by_topic(topic_name):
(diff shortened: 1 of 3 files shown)


mm318:
@jaisontj, @ivanpauno, @wjwwood, can any of you bring this pull request out of draft mode?

ros2topic/ros2topic/verb/pub.py
```diff
@@ -90,6 +92,9 @@ def main(args):
             args.once,
             qos_profile)
 
+def handle_incompatible_qos_event(event):
+    incompatible_qos_name = qos_policy_name_from_kind(event.last_policy_kind)
+    print('Incompatible QoS Policy detected: {name}'.format(name=incompatible_qos_name))
```
ivanpauno: `f-strings`

ros2topic/ros2topic/verb/pub.py
```diff
@@ -107,7 +112,13 @@ def publisher(
     if not isinstance(values_dictionary, dict):
         return 'The passed value needs to be a dictionary in YAML format'
 
-    pub = node.create_publisher(msg_module, topic_name, qos_profile)
+    publisher_callbacks = PublisherEventCallbacks(
+        incompatible_qos=lambda event: handle_incompatible_qos_event(event))
```
ivanpauno: same about `lambda` usage

ros2topic/ros2topic/verb/info.py
```diff
@@ -61,7 +61,7 @@ def main(self, *, args):
                 except NotImplementedError as e:
                     return str(e)
 
-            print('Subscription count: %d' % node.count_subscribers(topic_name))
+            print('Subscription count: %d' % node.count_subscribers(topic_name), end=line_end)
```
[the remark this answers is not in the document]
mm318: Oops. I'll open a separate pull request for this change.

Karsten1987:
This change calls `rmw_create_publisher` twice per topic to publish on. In the case of DDS (tested on FastRTPS only) this might work well, but other RMW implementations, i.e. iceoryx might not support this.  

Is there a way that the publisher is called only once? Meaning, can the QoS event checked somewhat before creating a publisher?

mm318:
We could add an option to `rclcpp`'s and `rclpy`'s `create_publisher()`/`create_subscription()` to ignore registering callbacks of unsupported event types, instead of throwing and retrying to `create_publisher()`/`create_subscription()` at the client level.

PR #363, opened 2019-10-16
Deprecate ros2msg and ros2srv

jacobperron:
Both packages have been superseded by ros2interface, which provides the same functionality
and more.

PR #405, opened 2019-11-25
[ros2action] Remove show verb

jacobperron:
The verb is redundant with 'ros2 interface show'.
Equivalent tests for the ones removed already exist in ros2interface.

Connects to https://github.com/ros2/ros2cli/issues/218

PR #413, opened 2019-12-02
[eloquent] [ros2action] Deprecate show verb 

jacobperron:
This is a follow-up to #363, as we've consolidated interface CLI logic to ros2interface.

### DIFF
--- a/ros2action/ros2action/verb/show.py
+++ b/ros2action/ros2action/verb/show.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import warnings
+
 from ros2action.api import action_type_completer
 from ros2action.api import get_action_path
 from ros2action.verb import VerbExtension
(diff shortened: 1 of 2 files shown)


jacobperron:
Related PR removing this verb for Foxy: https://github.com/ros2/ros2cli/pull/405

dirk-thomas:
Why remove all the tests?

jacobperron:
> Why remove all the tests?

they were failing because of the deprecation warning.

dirk-thomas:
Updating the tests to still pass would be better to avoid that the command regresses in Eloquent.

jacobperron:
> Updating the tests to still pass would be better to avoid that the command regresses in Eloquent.

Good point. I've updated the tests instead (c5b5f8a).

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8780/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4602/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7182/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8699/)

PR #414, opened 2019-12-03
[ros2doctor] Simplify helper function in platform check

claireyywang:
Consolidate errors and warnings


ros2doctor/ros2doctor/api/platform.py
```diff
@@ -59,12 +48,12 @@ def category(self):
     def check(self):
         """Check system platform against ROS 2 Distro."""
         result = Result()
-        distros = _check_platform_helper()
-        if not distros:
-            result.add_error('ERROR: Missing rosdistro info. Unable to check platform.')
+        try:
+            distros = _check_platform_helper()
+        except (AttributeError, URLError, KeyError):
+            result.add_error('Error: Unable to fetch rosdistro information online.')
```
hidmic: @claireyywang I'm not sure if giving back _less_ information to the user is a good path forward. I don't know if this was agreed elsewhere though.

claireyywang: Thanks for reviewing! In `package.py`, the same logic is used and merged. I simplified the logic here to keep it consistent.  The errors are most likely to occur when either user's machine has no internet connection, or there is something wrong with `rosdistro` package and it's not returning the right urls. The first case is covered by the network check, while the latter they can't do much about and should probably file a ticket. I think it is reasonable to remove the lengthy warnings and replace with a higher level one.

claireyywang:
Linux [Build Status](https://ci.ros2.org/job/ci_linux/8862/)
Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/4655/)
OSX [Build Status](https://ci.ros2.org/job/ci_osx/7256/)
Windows [Build Status](https://ci.ros2.org/job/ci_windows/8774/)

ros2doctor/ros2doctor/api/platform.py
hidmic: Oops, sorry for the delay @claireyywang ! 

I see. I will say though that not giving enough information doesn't make things any easier. Regular ROS users will file a ticket, but will be otherwise unable to provide any insight as to what happened, which ultimately puts the burden on us. Power ROS users likely won't wait for a maintainer reply round-trip and will instead be forced to remove that (quietly) 
 catch-all clause to see the actual error. I understand your point about consistency, but I can't help wondering if the code that should be changed is the one in `package.py`, and not the other way around. 

Anyways, if folks things this is the way, I won't stand in the way.

claireyywang: No worries. I see what you mean and it makes sense to expose more info to users. I don't feel strongly about how it should be, so I'll close this and make changes to `package.py` instead.

claireyywang: No worries. I see what you mean and it makes sense to expose more info to users. I don't feel strongly about how it should be, so I'll close this and make changes to package.py instead.

claireyywang:
closing to keep original logic

Issue #218, opened 2019-04-13
Move rosidl specific logic to rosidl_runtime_py


jacobperron:
#### Feature description

There are several TODOs for moving rosidl-related logic to a rosidl package. Now that the [rosidl_runtime_py](https://github.com/ros2/rosidl_python/tree/master/rosidl_runtime_py) package exists, it seems like a good place.

TODOs:

https://github.com/ros2/ros2cli/blob/07640db78d1963eee526d377aea6eb59e23edc6c/ros2msg/ros2msg/api/__init__.py#L39-L44

https://github.com/ros2/ros2cli/blob/07640db78d1963eee526d377aea6eb59e23edc6c/ros2msg/ros2msg/api/__init__.py#L52-L54

https://github.com/ros2/ros2cli/blob/07640db78d1963eee526d377aea6eb59e23edc6c/ros2msg/ros2msg/verb/show.py#L30

https://github.com/ros2/ros2cli/blob/07640db78d1963eee526d377aea6eb59e23edc6c/ros2srv/ros2srv/api/__init__.py#L39-L44

https://github.com/ros2/ros2cli/blob/07640db78d1963eee526d377aea6eb59e23edc6c/ros2srv/ros2srv/api/__init__.py#L52-L54

https://github.com/ros2/ros2cli/blob/07640db78d1963eee526d377aea6eb59e23edc6c/ros2srv/ros2srv/verb/show.py#L30-L31

https://github.com/ros2/ros2cli/blob/07640db78d1963eee526d377aea6eb59e23edc6c/ros2service/ros2service/verb/call.py#L60

Also, from ros2action being added in #214

jacobperron:
I believe all of the TODOs are address by https://github.com/ros2/ros2cli/pull/371, https://github.com/ros2/ros2cli/pull/407, https://github.com/ros2/ros2cli/pull/405, and https://github.com/ros2/ros2cli/pull/406. Once they are merged I'll close this issue.

jacobperron:
Found another TODO; addressing it in #415.

jacobperron:
All connected PRs have been merged.

PR #415, opened 2019-12-04
[ros2topic] Use import message logic provided by rosidl_runtime_py

jacobperron:
Connects to #218.

Note that the action feedback logic in the echo verb was incorrect, resulting in a ModuleImportError.
The new logic added in https://github.com/ros2/rosidl_runtime_py/pull/9 should fix the error.

Depends on https://github.com/ros2/rosidl_runtime_py/pull/9

### DIFF
--- a/ros2topic/ros2topic/verb/echo.py
+++ b/ros2topic/ros2topic/verb/echo.py
@@ -26,12 +26,12 @@ from rclpy.validate_full_topic_name import validate_full_topic_name
 from ros2cli.node.direct import DirectNode
 from ros2topic.api import add_qos_arguments_to_argument_parser
 from ros2topic.api import get_topic_names_and_types
-from ros2topic.api import import_message_type
 from ros2topic.api import qos_profile_from_short_keys
 from ros2topic.api import TopicNameCompleter
 from ros2topic.verb import VerbExtension
 from rosidl_runtime_py import message_to_csv
 from rosidl_runtime_py import message_to_yaml
+from rosidl_runtime_py.utilities import get_message
 
 DEFAULT_TRUNCATE_LENGTH = 128
 MsgType = TypeVar('MsgType')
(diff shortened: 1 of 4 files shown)


jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8823/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4636/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7237/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8739/)

ros2topic/test/fixtures/repeater_node.py
```diff
@@ -37,14 +36,10 @@ def callback(self):
...
 
-def message_type(message_typename):
-    return import_message_type('~/output', message_typename)
-
-
 def parse_arguments(args=None):
     parser = argparse.ArgumentParser()
     parser.add_argument(
-        'message_type', type=message_type,
+        'message_type',
```
hidmic: @jacobperron fyi: the initial intent here was to have the argument parser fail right away and react accordingly if a given `message_type` is invalid or the actual message type instance cannot be looked up.

jacobperron: If I change this line to have `type=get_message`, I run into an error as it seems the message object is passed (instead of a string).

Examples error: `AttributeError: type object 'BoundedSequences' has no attribute 'split'`

I'm not sure what's going on.

jacobperron: Ah, I see now. I'm accidentally calling `get_message` twice. Fixed in 83bce83

jacobperron:
Rebuild with latest change for sanity: [Build Status](https://ci.ros2.org/job/ci_linux/8850/)

ros2topic/ros2topic/verb/echo.py
```diff
@@ -128,7 +128,7 @@ def subscriber(
             raise RuntimeError(
                 'Could not determine the type for the passed topic')
 
-    msg_module = import_message_type(topic_name, message_type)
+    msg_module = get_message(message_type)
```
hidmic: @jacobperron meta^2: I wonder what'd happen if you used `nargs='?', type=get_message` above.

jacobperron: I guess it would be fine if we change this line to `msg_module = message_type`. But this seems out of scope of this PR. If you'd like to change it in a follow-up, go for it :)

PR #420, opened 2019-12-06
Extend CLI daemon based features

hidmic:
Connected to #419. This pull request:

- Augments daemon support for the ROS graph API. 
- Adds a `--no-daemon` option for `NodeStrategy` client code to use and allow users to NOT use nor spawn a daemon.

PR #421, opened 2019-12-09
Make CLI more robust to discovery latency.

### LINKED ISSUE #419: Deal with discovery latency consistently
#### Feature description

To decouple CLI commands from node discovery latencies, we currently spawn a daemon process that caches the ROS graph, speeding up queries. But we do so sparsely. Several verbs implementations -- [`ros2 security generate policy`](https://github.com/ros2/sros2/blob/083a850efa5cb8900b3fa8d8a9ec9460e6d7e034/sros2/sros2/verb/generate_policy.py#L119-L130), [`ros2 node info`](https://github.com/ros2/ros2cli/blob/19a491851efb043fac6b7e740c2838d6e1fb15e6/ros2node/ros2node/verb/info.py#L47-L52) and [`ros2 param list`](https://github.com/ros2/ros2cli/blob/19a491851efb043fac6b7e740c2838d6e1fb15e6/ros2param/ros2param/verb/list.py#L46-L59), just to name a few -- first use the daemon via `NodeStrategy` and then use `DirectNode`, effectively racing against discovery (e.g. the node in the daemon may have discovered a remote node already, but the node created by the CLI may have not _yet_). 

Thus, we should use the daemon correctly, potentially adding support for a larger subset of the ROS graph API in the process, and/or spin long enough so that discovery can take place.

hidmic:
Closes #419. Follow up after #420. This pull request extends daemon API support a bit more and adapts all CLI verbs to make use of it.

### DIFF
--- a/ros2cli/ros2cli/daemon/__init__.py
+++ b/ros2cli/ros2cli/daemon/__init__.py
@@ -23,6 +23,7 @@ from xmlrpc.server import SimpleXMLRPCServer
 import netifaces
 
 import rclpy
+import rclpy.action
 
 from ros2cli.node.direct import DirectNode
 
@@ -63,12 +64,18 @@ def main(*, script_name='_ros2_daemon', argv=None):
 
             # expose getter functions of node
             server.register_function(
-                _print_invoked_function_name(
-                    node.get_node_names_and_namespaces))
+                _print_invoked_function_name(node.get_name))
+            server.register_function(
+                _print_invoked_function_name(node.get_namespace))
+            server.register_function(
+                _print_invoked_function_name(node.get_node_names_and_namespaces))
             server.register_function(
                 _print_invoked_function_name(node.get_topic_names_and_types))
             server.register_function(
                 _print_invoked_function_name(node.get_service_names_and_types))
+            server.register_function(
+                _print_invoked_function_name(_bind_function(
+                    rclpy.action.get_action_names_and_types, node)))
             server.register_function(
                 _print_invoked_function_name(node.get_publisher_names_and_types_by_node))
             server.register_function(
@@ -77,6 +84,12 @@ def main(*, script_name='_ros2_daemon', argv=None):
                 _print_invoked_function_name(node.get_service_names_and_types_by_node))
             server.register_function(
                 _print_invoked_function_name(node.get_client_names_and_types_by_node))
+            server.register_function(
+                _print_invoked_function_name(_bind_function(
+                    rclpy.action.get_action_server_names_and_types_by_node, node)))
+            server.register_function(
+                _print_invoked_function_name(_bind_function(
+                    rclpy.action.get_action_client_names_and_types_by_node, node)))
             server.register_function(
                 _print_invoked_function_name(node.count_publishers))
             server.register_function(
@@ -182,6 +195,20 @@ def get_daemon_port():
 
 class RequestHandler(SimpleXMLRPCRequestHandler):
     rpc_paths = ('/ros2cli/',)
+
+
+def _bind_function(func, *args, **kwargs):
+    """
+    Bind a function with a set of arguments.
+
+    A functools.partial equivalent that is actually a function.
+    """
+    partial = functools.partial(func, *args, **kwargs)
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
+        return partial(*args, **kwargs)
+    wrapper.__signature__ = inspect.signature(func)
+    return wrapper
 
 
 def _print_invoked_function_name(func):
--- a/ros2cli/ros2cli/daemon/__init__.py
+++ b/ros2cli/ros2cli/daemon/__init__.py
@@ -17,8 +17,6 @@ from collections import namedtuple
 import functools
 import inspect
 import os
-from xmlrpc.server import SimpleXMLRPCRequestHandler
-from xmlrpc.server import SimpleXMLRPCServer
 
 import netifaces
 
@@ -26,6 +24,8 @@ import rclpy
 import rclpy.action
 
 from ros2cli.node.direct import DirectNode
+
+import ros2cli.xmlrpc
 
 
 def main(*, script_name='_ros2_daemon', argv=None):
@@ -55,8 +55,9 @@ def main(*, script_name='_ros2_daemon', argv=None):
         node_name_suffix='_daemon_%d' % args.ros_domain_id,
         start_parameter_services=False)
     with NetworkAwareNode(node_args) as node:
-        server = LocalXMLRPCServer(
-            addr, logRequests=False, requestHandler=RequestHandler,
+        server = ros2cli.xmlrpc.local_server.LocalXMLRPCServer(
+            addr, logRequests=False,
+            requestHandler=ros2cli.xmlrpc.local_server.RequestHandler,
             allow_none=True)
 
         try:
@@ -79,7 +80,11 @@ def main(*, script_name='_ros2_daemon', argv=None):
             server.register_function(
                 _print_invoked_function_name(node.get_publisher_names_and_types_by_node))
             server.register_function(
+                _print_invoked_function_name(node.get_publishers_info_by_topic))
+            server.register_function(
                 _print_invoked_function_name(node.get_subscriber_names_and_types_by_node))
+            server.register_function(
+                _print_invoked_function_name(node.get_subscriptions_info_by_topic))
             server.register_function(
                 _print_invoked_function_name(node.get_service_names_and_types_by_node))
             server.register_function(
@@ -179,22 +184,10 @@ class NetworkAwareNode:
             print('Daemon node was reset')
 
 
-class LocalXMLRPCServer(SimpleXMLRPCServer):
-
-    def verify_request(self, request, client_address):
-        if client_address[0] != '127.0.0.1':
-            return False
-        return super(LocalXMLRPCServer, self).verify_request(request, client_address)
-
-
 def get_daemon_port():
     base_port = 11511
     base_port += int(os.environ.get('ROS_DOMAIN_ID', 0))
     return base_port
-
-
-class RequestHandler(SimpleXMLRPCRequestHandler):
-    rpc_paths = ('/ros2cli/',)
 
 
 def _bind_function(func, *args, **kwargs):
--- a/ros2cli/ros2cli/node/direct.py
+++ b/ros2cli/ros2cli/node/direct.py
@@ -15,6 +15,7 @@
 import os
 
 import rclpy
+import rclpy.action
 
 from ros2cli.node import NODE_NAME_PREFIX
 DEFAULT_TIMEOUT = 0.5
@@ -54,6 +55,20 @@ class DirectNode:
 
     def __enter__(self):
         return self
+
+    # TODO(hidmic): generalize/standardize rclpy graph API
+    #               to not have to make a special case for
+    #               rclpy.action
+    def get_action_names_and_types(self):
+        return rclpy.action.get_action_names_and_types(self.node)
+
+    def get_action_client_names_and_types_by_node(self, remote_node_name, remote_node_namespace):
+        return rclpy.action.get_action_client_names_and_types_by_node(
+            self.node, remote_node_name, remote_node_namespace)
+
+    def get_action_server_names_and_types_by_node(self, remote_node_name, remote_node_namespace):
+        return rclpy.action.get_action_server_names_and_types_by_node(
+            self.node, remote_node_name, remote_node_namespace)
 
     def __getattr__(self, name):
         if not rclpy.ok():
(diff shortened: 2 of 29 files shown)


ros2cli/ros2cli/node/direct.py
```diff
@@ -56,7 +59,15 @@ def __enter__(self):
...
-
+        # TODO(hidmic): generalize/standardize rclpy graph API
+        #               to not have to make a special case for
+        #               rclpy.action.graph.
+        if hasattr(rclpy.action.graph, name):
+            function = getattr(rclpy.action.graph, name)
+            if inspect.isfunction(function):
+                function_signature = inspect.signature(function)
+                if 'node' == tuple(function_signature.parameters)[0]:
+                    return functools.partial(function, self)
```
ivanpauno: This is not really nice.
It would be better if `DaemonNode` could be used as a "normal node", instead of patching the `DaemonNode` and modifying the `DirectNode` to work in the same way.
Probably, this is not possible using XML-RPC. Maybe we could use a different IPC mechanism?

ros2component/ros2component/api/__init__.py
```diff
@@ -82,7 +81,10 @@ def get_container_components_info(*, node, remote_container_node_name):
         '{}/_container/list_nodes'.format(remote_container_node_name)
     )
     try:
-        list_nodes_client.wait_for_service()
+        if not list_nodes_client.wait_for_service(timeout_sec=5.0):
+            raise RuntimeError(
+                f'No list_nodes service found for {remote_container_node_name} container'
+            )
```
ivanpauno: I think this fix should be part of a different PR, or at least, don't squash the commit.

ros2component/ros2component/api/__init__.py
```diff
@@ -128,7 +130,10 @@ def load_component_into_container(
         '{}/_container/load_node'.format(remote_container_node_name)
     )
     try:
-        load_node_client.wait_for_service()
+        if not load_node_client.wait_for_service(timeout_sec=5.0):
+            raise RuntimeError(
+                f'No load_node service found for {remote_container_node_name} container'
+            )
```
ivanpauno: Same here.

ros2component/ros2component/api/__init__.py
```diff
@@ -177,7 +182,10 @@ def unload_component_from_container(*, node, remote_container_node_name, compone
         '{}/_container/unload_node'.format(remote_container_node_name)
     )
     try:
-        unload_node_client.wait_for_service()
+        if not unload_node_client.wait_for_service(timeout_sec=5.0):
```
ivanpauno: and here

ros2cli/ros2cli/daemon/__init__.py
```diff
@@ -63,12 +64,18 @@ def main(*, script_name='_ros2_daemon', argv=None):
 
             # expose getter functions of node
             server.register_function(
-                _print_invoked_function_name(
-                    node.get_node_names_and_namespaces))
+                _print_invoked_function_name(node.get_name))
```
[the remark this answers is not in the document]
hidmic: For `NodeStrategy` to be a drop-in replacement for `DirectNode`, yes it is. First use introduced [here](https://github.com/ros2/ros2cli/blob/a46034bc4e33db6b5d3268e446f78ec09c99a0b2/ros2action/ros2action/api/__init__.py#L32).

ros2cli/ros2cli/daemon/__init__.py
```diff
@@ -63,12 +64,18 @@ def main(*, script_name='_ros2_daemon', argv=None):
...
+            server.register_function(
+                _print_invoked_function_name(node.get_namespace))
```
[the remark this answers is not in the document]
hidmic: Same as above.

ros2component/ros2component/api/__init__.py
hidmic: Not planning to squash, that's why I pushed separate commits.

ros2component/ros2component/api/__init__.py
hidmic: Same as above.

ros2component/ros2component/api/__init__.py
hidmic: Same as above.

ros2cli/ros2cli/node/direct.py
hidmic: Yeap, I agree. But the problem isn't the IPC mechanism, really. Graph API for actions in `rclpy` departs from the pattern that the rest follow and on which `NodeStrategy` implementation rests upon. On a regular node, you'd simply call `rclpy.action.get_action_*(node, *args)`, but those simply forward down to the C layer (see [here](https://github.com/ros2/rclpy/blob/2882040a6f3e78e84fc6d08edf169a9a14fa6ef8/rclpy/rclpy/action/graph.py#L37-L39)). The pattern doesn't lend itself well to polymorphism. That's why I made `DirectNode` look like `NodeStrategy`, and not the other way around.

I have yet to come up with a way to do this nicely and transparent for those who write API like that of `rclpy.action`. We could make our C calls through the handle, like:

```python
def get_action_client_names_and_types_by_node(
    node: Node,
    remote_node_name: str,
    remote_node_namespace: str
) -> List[Tuple[str, List[str]]]:
    get_action_client_names_and_types_by_node = node.handle.bind(
        _rclpy_action.rclpy_action_get_client_names_and_types_by_node)
    return get_action_client_names_and_types_by_node(
        remote_node_name, remote_node_namespace)
``` 
It isn't pretty, but it gives us a chance to proxy calls. 

Tagging @jacobperron @wjwwood, both probably have thoughts on this.

ivanpauno: I like that proposal.
It's much more work though, but the result is cleaner.

jacobperron: I agree that we could be more consistent with the API in rclpy. I think any refactoring in rclpy can be left as part of a more thorough design review. For now, what if we just add a couple action specific functions to the DirectNode API?
I think it's easier to understand than inspecting function signatures, e.g.

```python
class DirectNode:

    # ...

    def get_action_names_and_types():
        return rclpy.action.get_action_names_and_types(self.node)

    def get_action_client_names_and_types_by_node(node_name, node_ns):
        return rclpy.action.get_action_client_names_and_types_by_node(self.node, node_name, node_ns)

     # ... etc
```

hidmic: I was trying to not do that :). But I'm fine with being explicit. See https://github.com/ros2/ros2cli/pull/421/commits/c0a78feac7d0e4f78f5d35d598541739e37b639e.

hidmic:
@jacobperron @ivanpauno once you are both OK with this patch, I'll rebase and run CI.

ros2topic/test/test_cli.py
```diff
@@ -504,6 +504,7 @@ def test_truncate_length_topic_echo(self):
             ), timeout=10)
         assert topic_command.wait_for_shutdown(timeout=10)
 
+    @launch_testing.markers.retry_on_failure(times=5)
```
dirk-thomas: Why are these changes needed and how are they related to this PR?

ros2cli/ros2cli/node/direct.py
jacobperron: I understand, but I think the diff is easier to parse with override functions. The TODO still holds in any case.

ros2topic/test/test_cli.py
hidmic: Those changes slipped in, my bad. These tests flaked me once locally, and later I forgot to take decorators out. We can split into their own PR if you'd like. Truth is I cannot consistently reproduce the flake.

dirk-thomas: Yeah, splitting sounds reasonable since nobody will expect this change in the squashed commit after the fact.

hidmic:
Running CI up to `ros2cli`, `ros2topic`, `ros2action`, `ros2component`, `ros2service`, `ros2node`:
- Linux [Build Status](https://ci.ros2.org/job/ci_linux/8874/)
- Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/4665/)
- OSX [Build Status](https://ci.ros2.org/job/ci_osx/7268/)
- Windows [Build Status](https://ci.ros2.org/job/ci_windows/8782/)

hidmic:
@jacobperron @dirk-thomas @ivanpauno I've rebased to leave a clean history. I plan to keep each commit. PTAL while CI runs.

ros2cli/ros2cli/node/direct.py
```diff
@@ -53,10 +54,31 @@ def timer_callback():
...
+    # TODO(hidmic): generalize/standardize rclpy graph API
+    #               to not have to make a special case for
+    #               rclpy.action
+    def get_action_names_and_types(self):
+        return rclpy.action.get_action_names_and_types(self.node)
+
+    def get_action_client_names_and_types_by_node(
+        self,
+        remote_node_name,
+        remote_node_namespace
```
dirk-thomas: Nitpick: how about keeping all these arguments on one line?

Same a few lines below.

ros2cli/ros2cli/node/direct.py
```diff
@@ -53,10 +54,31 @@ def timer_callback():
...
+    ):
+        return rclpy.action.get_action_server_names_and_types_by_node(
+            self.node, remote_node_name, remote_node_namespace)
+
     def __getattr__(self, name):
         if not rclpy.ok():
             raise RuntimeError('!rclpy.ok()')
-
```
dirk-thomas: Nitpick: unrelated whitespace change.

ros2component/ros2component/api/__init__.py
```diff
@@ -82,7 +81,10 @@ def get_container_components_info(*, node, remote_container_node_name):
         '{}/_container/list_nodes'.format(remote_container_node_name)
     )
     try:
-        list_nodes_client.wait_for_service()
+        if not list_nodes_client.wait_for_service(timeout_sec=5.0):
+            raise RuntimeError(
+                f'No list_nodes service found for {remote_container_node_name} container'
```
dirk-thomas: Please add single quotes around the placeholder (and therefore use double quotes for the string literal).

Same twice below.

hidmic: Done in https://github.com/ros2/ros2cli/pull/421/commits/7f5057a27aa0abf03a011c1f08da87f94fdd70c0.

ros2cli/ros2cli/node/direct.py
hidmic: Done in https://github.com/ros2/ros2cli/pull/421/commits/e3eee5c56a2faef4664f54289df965f94448bb13.

ros2cli/ros2cli/node/direct.py
hidmic: Huh, yeah, it gets close to a 100 characters long but not quite. Done in https://github.com/ros2/ros2cli/pull/421/commits/e3eee5c56a2faef4664f54289df965f94448bb13.

dirk-thomas: I wasn't asking to keep the whole function signature in one line - only the arguments:

```
def get_action_client_names_and_types_by_node(
        self, remote_node_name, remote_node_namespace
):
```

hidmic: Do you have a strong opinion on this regard? It does comply with `flake8` as-is now, I over-indented it initially.

dirk-thomas: If it passes it can stay as is.

hidmic:
Checking CI failures I came across a small pre-existing bug (https://github.com/ros2/ros2cli/pull/421/commits/645b3ee0262a9d527a0f3da86abd0421f8dce0de). Of course I cannot locally reproduce any of these failures. Let's see what comes out of a Windows CI now.

hidmic:
Running CI again up to `ros2cli`, `ros2topic`, `ros2action`, `ros2component`, `ros2service`, `ros2node`:
- Linux [Build Status](https://ci.ros2.org/job/ci_linux/9943/)

hidmic:
Ok, CI's green, which I attribute to many `ros2cli` tests being disabled under Windows. @jacobperron @dirk-thomas what would you like to do with this patch? Do we merge it now or do we defer?

dirk-thomas:
(I invalidated all previous reviews since there have been significant changes to the PR since then.)

Imo the current patch contains too many unrelated changes to land in a single PR. It should be split up into individually reviewed changes.

hidmic:
> Imo the current patch contains too many unrelated changes to land in a single PR. It should be split up into individually reviewed changes.

Fair enough, I can split it.

hidmic:
Alright, now that #493 and #494 are up we can close this one.

Issue #411, opened 2019-12-02
[ros2topic] info verb should support hidden topics


jacobperron:
#### Feature description

We can list hidden topics with the flag `--include-hidden-topics`

    ros2 topic list --include-hidden-topics

But we can't get info about any hidden topics.

For example, start publishing to a hidden topic:

    ros2 topic pub /_chatter std_msgs/msg/String "data: hello" 

And in another terminal, attempt to get info for the hidden topic:

    ros2 topic info /_chatter

We get the error:

    Unknown topic '/_chatter'

But I would expect the output to be:

    Type: std_msgs/msg/String
    Publisher count: 1
    Subscriber count: 0

evshary:
This feature can be supported by adding argument `include_hidden_topics=True`.
However I'm not sure whether it's fine to show hidden topic directly or it's necessary to add extra option, like `--include-hidden-topics`.

jacobperron:
Closed by #423

PR #423, opened 2019-12-15
Able to show info of hidden topic.

evshary:
According #411, support show hidden topic info with `ros2 topic info <hidden topic>`
However, I'm not sure whether it's fine to show hidden topic directly or it's necessary to add extra option, like `--include-hidden-topics`.

### DIFF
--- a/ros2topic/ros2topic/verb/info.py
+++ b/ros2topic/ros2topic/verb/info.py
@@ -30,7 +30,8 @@ class InfoVerb(VerbExtension):
 
     def main(self, *, args):
         with DirectNode(args) as node:
-            topic_names_and_types = get_topic_names_and_types(node=node)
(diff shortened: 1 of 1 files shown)


jacobperron:
@evshary Thanks for the PR. I think it makes sense to **not** show the hidden topics by default since they are hidden for a reason. Most users shouldn't need to worry about them because they are an implementation detail. But in cases where we want to probe the hidden topics, it's nice to have the option. So, I think we should add the extra option as described (it is also consistent with other verbs like `ros2 topic list`).

jacobperron:
Disregard my previous comment (I still need a coffee). Since we provide the topic name explicitly, an extra option shouldn't be necessary :)

jacobperron:
Thanks for the contribution!

Issue #425, opened 2019-12-23
[ros2doctor] Python exception thrown when following tutorial. 


LoyVanBeek:
## Bug report
Running `ros2 doctor --report` in a Docker container fails with a Python stack trace

**Required Info:**

- Operating System:
  - Ubuntu 18.04.3 in `ros:eloquent` docker image
- Installation type:
  - Docker image, not sure how the docker image itself is built
- Version or commit hash:
  - 0.8.6 (as found by `dpkg -S ros2cli`, there is no `-v` flag yet)
- DDS implementation:
  - Fast-RTPS (nothing installed additional to what's in the docker image)

#### Steps to reproduce issue
I was following https://index.ros.org/doc/ros2/Tutorials/Getting-Started-With-Ros2doctor/#id6 in said docker image:
```
ros2 doctor
ros2 doctor --report
```

#### Expected behavior
An error information about network configuration

#### Actual behavior
```
root@5832e3497fea:~/ros2/dev_ws# ros2 doctor
/opt/ros/eloquent/lib/python3.6/site-packages/ros2doctor/api/format.py: 76: UserWarning: ERROR: No loopback IP address is found.
/opt/ros/eloquent/lib/python3.6/site-packages/ros2doctor/api/format.py: 76: UserWarning: Only loopback IP address is found.
/opt/ros/eloquent/lib/python3.6/site-packages/ros2doctor/api/format.py: 76: UserWarning: No multicast IP address is found.
/opt/ros/eloquent/lib/python3.6/site-packages/ros2doctor/api/format.py: 76: UserWarning: Publisher without subscriber detected on /topic.

1/3 checks failed

Failed modules: network
root@5832e3497fea:~/ros2/dev_ws# ros2 doctor --report

  NETWORK CONFIGURATION
Traceback (most recent call last):
  File "/opt/ros/eloquent/bin/ros2", line 11, in <module>
    load_entry_point('ros2cli==0.8.6', 'console_scripts', 'ros2')()
  File "/opt/ros/eloquent/lib/python3.6/site-packages/ros2cli/cli.py", line 69, in main
    rc = extension.main(parser=parser, args=args)
  File "/opt/ros/eloquent/lib/python3.6/site-packages/ros2doctor/command/doctor.py", line 45, in main
    format_print(report_obj)
  File "/opt/ros/eloquent/lib/python3.6/site-packages/ros2doctor/api/format.py", line 27, in format_print
    padding_num = compute_padding(report.items)
[...]

PR #426, opened 2019-12-30
Depend on `net-tools`

LoyVanBeek:
`net-tools` (or `iproute2`) is an implicit dependency of `python3-ifcfg`, which can work with either `ip` (from `iproute2`) or `ifconfig` (from `net-tools`)

`net-tools` is a smaller dependency in terms of bytes installed and did not install recursive dependencies on the `ros:eloquent` docker image, but `ip` is more modern. If that is preferred, it's an easy change here as both packages are already in rosdep. 

Fixes https://github.com/ros2/ros2cli/issues/425

### DIFF
--- a/ros2doctor/package.xml
+++ b/ros2doctor/package.xml
@@ -13,6 +13,7 @@
   <exec_depend>python3-catkin-pkg-modules</exec_depend>
   <exec_depend>python3-ifcfg</exec_depend>
   <exec_depend>python3-rosdistro-modules</exec_depend>
+  <exec_depend>net-tools</exec_depend>
   
   <test_depend>ament_copyright</test_depend>
   <test_depend>ament_flake8</test_depend>
--- a/ros2doctor/package.xml
+++ b/ros2doctor/package.xml
@@ -10,10 +10,10 @@
   <depend>ros2cli</depend>
 
   <exec_depend>ament_index_python</exec_depend>
+  <exec_depend>net-tools</exec_depend>
   <exec_depend>python3-catkin-pkg-modules</exec_depend>
   <exec_depend>python3-ifcfg</exec_depend>
   <exec_depend>python3-rosdistro-modules</exec_depend>
-  <exec_depend>net-tools</exec_depend>
   
   <test_depend>ament_copyright</test_depend>
   <test_depend>ament_flake8</test_depend>


claireyywang:
Thanks for submitting a PR. I think the [`net-tools`](https://sourceforge.net/projects/net-tools/) used by `python3-ifcfg` is different from the python package [`net-tools`](https://pypi.org/project/net-tools/). In this case the `net-tools` we want would be installed by `sudo apt-get install net-tools`. If this dependency is added, the installation would be `pip install net-tools`, which is a different package as far as I know.

LoyVanBeek:
From what I gather from the source of `python3-ifcfg`, it does use the `ifconfig` (the one from net-tools) executable under the hood: https://github.com/ftao/python-ifcfg/blob/2d6a87f106c19aadad1aecec823f757574eb7fe4/src/ifcfg/parser.py#L214. It makes no mention of `net-tools` in it's dependencies (it does not specify any requirements, only some dev-requirements).
`net-tools` is defined in rosdep on https://github.com/ros/rosdistro/blob/master/rosdep/base.yaml#L4298 which is an `apt` package AFAICT. There is no mention of `net-tools` in https://github.com/ros/rosdistro/blob/master/rosdep/python.yaml. 

You say
> If this dependency is added, the installation would be `pip install net-tools`

How so? I must admit I'm not super familiar with how rosdep works. The dependency does not mention `python` and it's not in the `python.yaml` in distro so I don't see why rosdep would use pip over apt for this?

But I can imagine this PR being a hassle for little benefit as any full system will have either `ip` or `ifconfig` installed already, but for completeness sake.

claireyywang:
I misunderstood, and yes you are right about rosdep and debian installation, but there is another problem. `net-tools` is linux specific, and adding it as a general dependency will cause conflict on osx and windows when trying to install a package not present on their platforms. 

An alternative approach to resolve this issue would be that, we can either add `net-tools` to [linux installation documentation](https://github.com/ros2/ros2_documentation/tree/master/source/Installation/Eloquent), or add a warning similar to https://github.com/ros2/ros2cli/blob/93e717c042a06d21e267a89a7de7780335acb323/ros2doctor/ros2doctor/api/network.py#L70-L73 in network check, so that package exit gracefully when `ifconfig` is not present. If you're interested in implementing either of those, feel free to submit another PR to the related repo. I'll go ahead and close this unless further discussion is desired. In which case feel free to reopen it.

LoyVanBeek:
Sorry for bothering once more, but `ifcfg` simply reports an empty list of interfaces when `ip` and `ifconfig` are not available. That's an easy check of course. I'm not sure if there ever can be a case where there even isn't a loopback interface. If there is always at least a loopback interface, than a empty-list check is enough to raise an error on. 

In the docker image, this is NOT correct: there are network interfaces but the underlying command fails with `ip: command not found`. 

I created https://github.com/ftao/python-ifcfg/issues/44 for this issue

Issue #427, opened 2020-01-02
add support for additional build-types, e.g., ros2 pkg create --build-type anewbuildtype


wayneparrott:
This enhancement request is to enable the creation of packages for build-types other than c/c++ and python, i.e.,`ros2 pkg create --build-type mynewbuildtype`

Specifically I'm proposing the addition of an extension point for additional build-types. Something like this:

```python
entry_points={
        'ros2cli.extension_point': [
            'ros2pkg.build_type = ros2pkg.build_type:BuildTypeExtension',
        ],
```

As an experiment I implemented this extension point and it was quite simple to work into  ros2pkg.verb.command.py with minimal changes, no refactoring of existing build-types required. I've been using my implementation for creating javascript/[rclnodejs](https://github.com/RobotWebTools/rclnodejs) projects. I would be glad to share a PR for your consideration. If included in ros2cli, we (the rclnodejs team) will release a new build type for creating nodejs projects that properly work with `ros2 run` and `ros2 launch` commands.  

Thoughts?

dirk-thomas:
That sounds great. Please create a PR with that proposal.

wayneparrott:
@ivanpauno I'm in the process of prep'ing a PR for this enhancement. I'm blocked with a dev environment issue running the ros2pkg/test_cli.py tests. Once I can run those tests successfully I'll submit the PR for review.

ivanpauno:
> @ivanpauno I'm in the process of prep'ing a PR for this enhancement. I'm blocked with a dev environment issue running the ros2pkg/test_cli.py tests. Once I can run those tests successfully I'll submit the PR for review.

Sounds great, let us know if you have any question.

ros-discourse:
This issue has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/proposal-ability-to-add-new-build-types-to-ros2cli-e-g-ros2-pkg-create-build-type-newbuildtype/13259/1

PR #429, opened 2020-01-03
add build-type extension-point for the ros2 pkg create verb #427

wayneparrott:
This PR introduces the new extension-point **ros2pkg.build_type** for contributing custom build-types to ros2, see #427. 

A developer would implement a custom build-type extension as follows:

In his python package's setup.py include:

```
entry_points={
        'ros2pkg.build_type': [
            'foobar = myproject.build_type.foobar:FooBarBuildType',
        ],
}
```

In his build-type package implement:
```
from ros2pkg.build_type import BuildTypeExtension
class FooBarBuildType(BuildTypeExtension):

    def __init__(self):
        super(FooBarBuildType, self).__init__()

    def populate(self, package, package_directory, node_name, library_name):
       #custom build stuff here
```

Note: I did not include tests (yet) as I will need guidance on how one should go about testing an extension point.

### DIFF
--- a/ros2pkg/ros2pkg/build_type/__init__.py
+++ b/ros2pkg/ros2pkg/build_type/__init__.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from ros2cli.entry_points import load_entry_points
+from ros2cli.entry_points import get_entry_points
 from ros2cli.plugin_system import instantiate_extensions
 from ros2cli.plugin_system import PLUGIN_SYSTEM_VERSION
 from ros2cli.plugin_system import satisfies_version
@@ -20,7 +20,7 @@ from ros2cli.plugin_system import satisfies_version
 
 class BuildTypeExtension:
     """
-    The interface for verb extensions.
+    The interface for build-type extensions.
 
     The following properties must be defined:
     * `NAME` (will be set to the entry point name)
@@ -30,8 +30,8 @@ class BuildTypeExtension:
     """
 
     NAME = None
-    EXTENSION_POINT_VERSION = '0.1'
     EXTENSION_POINT_NAME = 'ros2pkg.build_type'
+    EXTENSION_POINT_VERSION = '0.1'
     NATIVE_BUILD_TYPES = ['cmake', 'ament_cmake', 'ament_python']
 
     def __init__(self):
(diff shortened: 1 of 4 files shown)


ros2pkg/ros2pkg/build_type/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+    * `NAME` (will be set to the entry point name)
+
+    The following methods must be defined:
+    * `main`
+    """
+
+    NAME = None
+    EXTENSION_POINT_VERSION = '0.1'
+    EXTENSION_POINT_NAME = 'ros2pkg.build_type'
+    NATIVE_BUILD_TYPES = ['cmake', 'ament_cmake', 'ament_python']
```
dirk-thomas: I would suggest to not makes these build types special but implement them with an entry point too.

wayneparrott: Refactoring the existing build-types as extensions is a little bigger than I wanted to bite off in my first ros2 contribution. Let me propose that we move forward with this in 2 phases. Phase-1: intro the new build-type extension point - I already have a custom build-type that can demonstrate the value and serve as a guide for others.  Then in phase-2 we refactor the current 4 build-types to plugin via the build-type extension point? I will commit to helping impl phase-2 if this proposal is approved. Thoughts?

dirk-thomas:
> Then in phase-2 we refactor the current 4 build-types to plugin via the build-type extension point? 

Sounds good to me.

wayneparrott:
Next steps for phase-1 (i.e., the build-type extension point) approval?

dirk-thomas:
Someone from the @ros2/team needs to review / run CI / approve.

ros2pkg/ros2pkg/build_type/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+
+    The following properties must be defined:
+    * `NAME` (will be set to the entry point name)
+
+    The following methods must be defined:
+    * `main`
+    """
+
+    NAME = None
+    EXTENSION_POINT_VERSION = '0.1'
```
claireyywang: `EXTENSION_POINT_VERSION` doesn't seem to be used anywhere. Do we need it?

ros2pkg/ros2pkg/build_type/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+        satisfies_version(PLUGIN_SYSTEM_VERSION, '^0.1')
+
+    def populate(self, package, package_directory, node_name, library_name):
+        pass
+
+
+def get_build_types():
+    build_types = BuildTypeExtension.NATIVE_BUILD_TYPES.copy()
+    extension_pts = load_entry_points(BuildTypeExtension.EXTENSION_POINT_NAME)
+    for name, extension_pt in extension_pts.items():
```
claireyywang: Suggest using `keys()` instead of `items()` to get all names

ros2pkg/ros2pkg/build_type/__init__.py
wayneparrott: I followed the pattern of all the rospkg extension points which include EXTENSION_POINT_VERSION. I scanned the codebase and did not find functional references to it. My assumption is the original developer had planned to make use of it at future date. I've elected to retain it in the build-type extension point for consistency with other extension points.

claireyywang:
CI 
Linux [Build Status](https://ci.ros2.org/job/ci_linux/9066/)
Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/4814/)
OSX [Build Status](https://ci.ros2.org/job/ci_osx/7417/)
Windows [Build Status](https://ci.ros2.org/job/ci_windows/8996/)

wayneparrott:
Please **HOLD** on merging this PR. After a quick look into what will be involved to support phase-2 (i.e., refactor of the current build-types, e.g. ament_python) I believe a basic design change is needed now to avoid legacy api problems in the very near future. 

In the current phase-1 design, build-type extensions are not delegated to until after creation of the package directory and package.xml file. I believe this is will cause us problems in the future. Instead I propose we delegate 100% of the package creation process to the build-type extension-point, including creation of the package directory and the package.xml file. Thus to support this change the build-type extension api needs additional info and should be changed immediately.

I would like to refactor the design and update the PR. How to proceed?  Thoughts?

claireyywang:
If the proposed change differs significantly from this PR, I'd recommend you close this one and reference it on a new PR for record keeping; else, updating the PR description with info of the proposed change and how to use/test would be useful in review.

I will retract the approval if you choose to go with the latter. Feel free to request another round of review whenever you're ready.

wayneparrott:
Let's close this PR and I'll open a new one asap.

PR #432, opened 2020-01-09
Multimachine communication

claireyywang:
> How about a verb called “yell” (or something more clever), where the program just spins and sends “hello-world” and it’s hostname once per second, over both a hard-coded ROS2/DDS topic and a “custom” multicast UDP packet as well. At the same time, it is listening for other messages on that same ROS2 topic and/or UDP multicast port and printing a summary table periodically of which other hosts it has heard from, message counts, and so on. This would help debug the current #1 problem we are seeing, where people have messaging working on localhost but can’t receive those messages on another host. This spins into a debug cycle of trying to figure out if the problem lies with their software, their network configuration, DDS discovery or transport bugs, the network infrastructure, and so on. Currently we debug this using known-good ros2 nodes such as examples_rclpy_minimal_* , wireshark, ping, ros2multicast, ros2node, and so on, but having a nicer way to do it would be great. 

Run `ros2 doctor call` command to test pub/sub, multicast send/receive quality across multiple machines. The command outputs a summary table /1s. 
```bash
MULTIMACHINE COMMUNICATION SUMMARY
Topic: knockknock, Published Msg Count: 20
Subscribed from:
                Hostname             Msg Count /2s
Multicast Group/Port: 225.0.0.1/49150, Sent Msg Count: 20
Received from:
                Hostname             Msg Count /2s
                Claires-MBP          3         
------------------------------------------------------------

```
Test cases will be submitted in a separate PR.

### DIFF
--- a/ros2doctor/ros2doctor/verb/call.py
+++ b/ros2doctor/ros2doctor/verb/call.py
@@ -249,10 +249,10 @@ class SummaryTable():
         """Increment subscribed msg count from different host(s)."""
         self.lock.acquire()
         try:
-            if hostname not in self.receive:
-                self.receive[hostname] = 1
+            if hostname not in self.sub:
+                self.sub[hostname] = 1
             else:
-                self.receive[hostname] += 1
+                self.sub[hostname] += 1
         finally:
             self.lock.release()
 
@@ -268,10 +268,10 @@ class SummaryTable():
         """Increment multicast-received msg count from different host(s)."""
         self.lock.acquire()
         try:
-            if hostname not in self.sub:
-                self.sub[hostname] = 1
+            if hostname not in self.receive:
+                self.receive[hostname] = 1
             else:
-                self.sub[hostname] += 1
+                self.receive[hostname] += 1
         finally:
             self.lock.release()
 
--- a/ros2doctor/ros2doctor/verb/call.py
+++ b/ros2doctor/ros2doctor/verb/call.py
@@ -1,0 +1,178 @@
+# Copyright 2019 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import socket
+import struct
+import threading
+import time
+
+import rclpy
+from rclpy.executors import MultiThreadedExecutor
+from rclpy.node import Node
+from ros2doctor.verb import VerbExtension
+
+from std_msgs.msg import String
+
+DEFAULT_GROUP = '225.0.0.1'
+DEFAULT_PORT = 49150
+DEFAULT_TOPIC = 'knockknock'
+summary_table = {}
+
+
+class CallVerb(VerbExtension):
+    """Publish and subscribe, multicast send and receive, print table."""
+
+    def main(self, *, args):
+        rclpy.init()
+        pub_node = Talker()
+        sub_node = Listener()
+
+        executor = MultiThreadedExecutor()
+        executor.add_node(pub_node)
+        executor.add_node(sub_node)
+        try:
+            count = 0
+            _spawn_summary_table()
+            while True:
+                if (count % 20 == 0 and count != 0):
+                    format_print(summary_table)
+                    _spawn_summary_table()
+                # pub/sub threads
+                executor.spin_once()
+                executor.spin_once()
+                # multicast threads
+                send_thread = threading.Thread(target=send, args=())
+                send_thread.daemon = True
+                receive_thread = threading.Thread(target=receive, args=())
+                receive_thread.daemon = True
+                receive_thread.start()
+                send_thread.start()
+                count += 1
+                time.sleep(0.1)
+        except KeyboardInterrupt:
+            executor.shutdown()
+            pub_node.destroy_node()
+            sub_node.destroy_node()
+
+
+class Talker(Node):
+    """Initialize talker node."""
+
+    def __init__(self):
+        super().__init__('talker')
+        self.i = 0
+        self.pub = self.create_publisher(String, DEFAULT_TOPIC, 10)
+        time_period = 0.1
+        self.timer = self.create_timer(time_period, self.timer_callback)
+
+    def timer_callback(self):
+        msg = String()
+        hostname = socket.gethostname()
+        # publish
+        msg.data = f'Publish hello from {hostname}'
+        summary_table['pub'] += 1
+        self.pub.publish(msg)
+        self.i += 1
+
+
+class Listener(Node):
+    """Initialize listener node."""
+
+    def __init__(self):
+        super().__init__('listener')
+        self.sub = self.create_subscription(String,
+                                            DEFAULT_TOPIC,
+                                            self.sub_callback,
+                                            10)
+
+    def sub_callback(self, msg):
+        # subscribe
+        msg_data = msg.data.split()
+        caller_hostname = msg_data[-1]
+        if caller_hostname != socket.gethostname():
+            if caller_hostname not in summary_table['sub']:
+                summary_table['sub'][caller_hostname] = 1
+            else:
+                summary_table['sub'][caller_hostname] += 1
+
+
+def send():
+    """Multicast send."""
+    hostname = socket.gethostname()
+    s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM, socket.IPPROTO_UDP)
+    try:
+        summary_table['send'] += 1
+        s.sendto(f'Multicast hello from {hostname}'.encode('utf-8'), (DEFAULT_GROUP, DEFAULT_PORT))
+    finally:
+        s.close()
+
+
+def receive():
+    """Multicast receive."""
+    s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM, socket.IPPROTO_UDP)
+    try:
+        s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        try:
+            s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEPORT, 1)
+        except AttributeError:
+            # not available on Windows
+            pass
+        s.bind(('', DEFAULT_PORT))
+
+        s.settimeout(None)
+
+        mreq = struct.pack('4sl', socket.inet_aton(DEFAULT_GROUP), socket.INADDR_ANY)
+        s.setsockopt(socket.IPPROTO_IP, socket.IP_ADD_MEMBERSHIP, mreq)
+        try:
+            data, _ = s.recvfrom(4096)
+            data = data.decode('utf-8')
+            sender_hostname = data.split()[-1]
+            if sender_hostname != socket.gethostname():
+                if sender_hostname not in summary_table['receive']:
+                    summary_table['receive'][sender_hostname] = 1
+                else:
+                    summary_table['receive'][sender_hostname] += 1
+        finally:
+            s.setsockopt(socket.IPPROTO_IP, socket.IP_DROP_MEMBERSHIP, mreq)
+    finally:
+        s.close()
+
+
+def _spawn_summary_table():
+    """Spawn summary table with new content after each print."""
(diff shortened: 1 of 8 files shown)


claireyywang:
@jacobperron The `bypassing network` part explained why I wasn't able to get pub/sub working on two machines when testing. Thanks for pointing it out! I copied the feature request in the PR description for completeness. Based on my understanding, it seems like they want one verb to do it all. What do you think? If that's not the best way to do it, I'd be happy to refactor the code and split it into send/pub and receive/sub.

ros2doctor/ros2doctor/verb/call.py
```diff
@@ -0,0 +1,178 @@
...
+            executor.shutdown()
+            pub_node.destroy_node()
+            sub_node.destroy_node()
+
+
+class Talker(Node):
+    """Initialize talker node."""
+
+    def __init__(self):
+        super().__init__('talker')
```
jacobperron: Make this node name unique; same with listener.

jacobperron:
Thanks for updating the description, it's helpful to have the context! I think I missed the point originally; you can disregard my previous comment.

There shouldn't be any network bypassing happening if there are different hosts on the same ROS network. Double check that they are using the same ROS_DOMAIN_ID. I think we can make some improvements to the executor/send-receive loop too.

ros2doctor/ros2doctor/verb/call.py
```diff
@@ -0,0 +1,178 @@
...
+        sub_node = Listener()
+
+        executor = MultiThreadedExecutor()
+        executor.add_node(pub_node)
+        executor.add_node(sub_node)
+        try:
+            count = 0
+            _spawn_summary_table()
+            while True:
+                if (count % 20 == 0 and count != 0):
```
jacobperron: Rather than counting to a number, I think it is more intuitive to run the loop at a given period (e.g. in seconds). I would make this period configurable with a command line option.

ros2doctor/ros2doctor/verb/call.py
```diff
@@ -0,0 +1,178 @@
...
+                executor.spin_once()
+                # multicast threads
+                send_thread = threading.Thread(target=send, args=())
+                send_thread.daemon = True
+                receive_thread = threading.Thread(target=receive, args=())
+                receive_thread.daemon = True
+                receive_thread.start()
+                send_thread.start()
+                count += 1
+                time.sleep(0.1)
```
jacobperron: We can also make the time we sleep configurable, e.g. letting the user specify the frequency in Hz or something.

ros2doctor/ros2doctor/verb/call.py
```diff
@@ -0,0 +1,178 @@
...
+        try:
+            count = 0
+            _spawn_summary_table()
+            while True:
+                if (count % 20 == 0 and count != 0):
+                    format_print(summary_table)
+                    _spawn_summary_table()
+                # pub/sub threads
+                executor.spin_once()
+                executor.spin_once()
```
jacobperron: This is tricky, since it doesn't guarantee that one iteration of this loop will result exactly one timer callback and one subscription callback. It could be two timer callbacks or two subscription callbacks, depending on the queue of events.

It may make sense to run the executor in it's own thread and just call `spin()`.

ros2doctor/ros2doctor/verb/call.py
```diff
@@ -0,0 +1,178 @@
...
+from std_msgs.msg import String
+
+DEFAULT_GROUP = '225.0.0.1'
+DEFAULT_PORT = 49150
+DEFAULT_TOPIC = '/knockknock'
+summary_table = {}
+
+
+class CallVerb(VerbExtension):
+    """Publish and subscribe, multicast send and receive, print table."""
```
jacobperron: I suggest making this description a bit more verbose so it's clear what this tool is does. E.g.

```
"""
Check network connectivity between multiple hosts.

This command can be invoked on multiple hosts to confirm that they can talk to each other.

TODO <describe how pub/sub is used to test ROS comms>
TODO <describe how multicast send/receive is used to confirm multicast UDP comms>
```

ros2doctor/ros2doctor/verb/call.py
claireyywang: > It may make sense to run the executor in it's own thread and just call `spin()`.

Do you mean like `threading.Thread(target=executor.spin)`? I thought `executor` already has multiple threads running, wouldn't putting the executor in another thread be repetitive?

jacobperron: > Do you mean like threading.Thread(target=executor.spin)? 

Yes.

The MultiThreadedExecutor does use multiple threads to execute callbacks, but it still needs someone to call `spin_once` (or `spin`) to poll for callbacks to submit for execution. Versus the SingleThreadedExecutor, which executes the callbacks in the same thread that `spin` is called in.

I'm not sure if we really need to use a MultiThreadedExecutor in this case; I think either will executor will work.

ros2doctor/ros2doctor/verb/call.py
```diff
@@ -0,0 +1,260 @@
...
+
+
+class Talker(Node):
+    """Initialize talker node."""
+
+    def __init__(self, topic, time_period, *, qos=10):
+        super().__init__('ros2doctor_talker')
+        self.i = 0
+        self.pub = self.create_publisher(String, topic, qos)
+        self.timer = self.create_timer(time_period, self.timer_callback)
```
jacobperron: nitpick: I think all three of these variables can be prefixed with an underscore since they are implementation details.

ros2doctor/ros2doctor/verb/call.py
```diff
@@ -0,0 +1,260 @@
...
+        self.pub.publish(msg)
+        self.i += 1
+
+
+class Listener(Node):
+    """Initialize listener node."""
+
+    def __init__(self, topic, *, qos=10):
+        super().__init__('ros2doctor_listener')
+        self.sub = self.create_subscription(
```
jacobperron:
```suggestion
        self._sub = self.create_subscription(
```

ros2doctor/ros2doctor/verb/call.py
```diff
@@ -0,0 +1,260 @@
...
+class SummaryTable():
+    """Summarize number of msgs published/sent and subscribed/received."""
+
+    def __init__(self):
+        """Initialize empty summary table."""
+        self.lock = threading.Lock()
+        self.pub = 0
+        self.send = 0
+        self.sub = {}
+        self.receive = {}
```
jacobperron: nitpick: I think all these variables can be prefixed with an underscore since they are implementation details.

ros2doctor/ros2doctor/verb/call.py
```diff
@@ -0,0 +1,260 @@
...
+        sub_node = Listener(args.topic_name)
+        executor.add_node(pub_node)
+        executor.add_node(sub_node)
+        try:
+            prev_time = time.time()
+            timeout = time.time() + args.duration
+            # pub/sub thread
+            exec_thread = threading.Thread(target=executor.spin)
+            exec_thread.start()
+            while time.time() < timeout:
```
jacobperron: Sorry if I gave you the wrong impression before, but I think looping indefinitely until a keyboard interrupt is fine here. Unless you think there's a use-case for only running for a limited duration, in which case we should still have the option to run indefinitely. 
```suggestion
            while True:
```

ros2doctor/ros2doctor/verb/call.py
```diff
@@ -0,0 +1,260 @@
...
+            'topic_name', nargs='?', default='/canyouhearme',
+            help="Name of ROS topic to publish to (default: '/canyouhearme')")
+        parser.add_argument(
+            'time_period', nargs='?', default=0.1,
+            help='Time period to publish/send one message (default: 0.1s)')
+        parser.add_argument(
+            'duration', nargs='?', default=20, type=positive_int,
+            help='How long this process runs (default: 20s)')
+        parser.add_argument(
+            '-r', '--rate', metavar='N', type=float, default=1.0,
```
jacobperron: I think that "rate" and "period" (or "time_period") may be easily confused. I suggest making this option name more verbose and change it to a period for consistency, e.g. "--output-period".

ros2doctor/ros2doctor/verb/call.py
```diff
@@ -0,0 +1,260 @@
...
+        super().__init__('ros2doctor_talker')
+        self.i = 0
+        self.pub = self.create_publisher(String, topic, qos)
+        self.timer = self.create_timer(time_period, self.timer_callback)
+
+    def timer_callback(self):
+        msg = String()
+        hostname = socket.gethostname()
+        # publish
+        msg.data = f"hello, it's me {hostname}"
```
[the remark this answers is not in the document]
claireyywang: I think it's double quotes cuz of the single quote in `it's`. The linters seem to prefer double quotes in this case to using `it\'s`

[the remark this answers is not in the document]
jacobperron: You're right, I didn't see the single quote.

ros2doctor/ros2doctor/verb/hello.py
```diff
@@ -0,0 +1,249 @@
...
+
+
+class HelloVerb(VerbExtension):
+    """
+    Check network connectivity between multiple hosts.
+
+    This command can be invoked on multiple hosts to confirm that they can talk to each other
+    by using talker/listener, multicast send/receive to check topic discovering and
+    UDP communication.
+    This command outputs a summary table of msgs statistics at a custom rate(Hz).
```
jacobperron:
```suggestion
    This command outputs a summary table of msgs statistics at a custom period (s)
```

ros2doctor/ros2doctor/verb/hello.py
```diff
@@ -0,0 +1,249 @@
...
+    """
+
+    def add_arguments(self, parser, cli_name):
+        parser.add_argument(
+            '-t', '--topic', nargs='?', default='/canyouhearme',
+            help="Name of ROS topic to publish to (default: '/canyouhearme')")
+        parser.add_argument(
+            '-rp', '--emit-period', metavar='N', type=float, default=0.1,
```
jacobperron: Why `-rp` instead of `-ep`?

ros2doctor/test/test_hello.py
```diff
@@ -0,0 +1,48 @@
...
+def test_hello_single_host():
+    """Run HelloVerb for one emit period on a single host."""
+    args = Namespace()
+    args.topic = '/canyouhearme'
+    args.emit_period = 0.1
+    args.print_period = 1.0
+    args.ttl = None
+    args.once = True
+    s = StringIO()
+    with redirect_stderr(s):
```
jacobperron: Why do we need to redirect stderr? It might be worth adding a comment explaining why.

ros2doctor/ros2doctor/verb/hello.py
claireyywang: ugh typo. nice catch!

claireyywang:
CI
Linux [Build Status](https://ci.ros2.org/job/ci_linux/9265/)
Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/4950/)
osx [Build Status](https://ci.ros2.org/job/ci_osx/7575/)
windows [Build Status](https://ci.ros2.org/job/ci_windows/9185/)
windows-container [Build Status](https://ci.ros2.org/job/ci_windows-container/67/)

claireyywang:
Linux-aarch64: [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/4958/) 
windows-container: [Build Status](https://ci.ros2.org/job/ci_windows-container/76/)

ros2doctor/test/test_hello.py
```diff
@@ -14,10 +14,39 @@
...
+        # Always restart daemon to isolate tests.
+        ExecuteProcess(
+            cmd=['ros2', 'daemon', 'stop'],
+            name='daemon-stop',
+            on_exit=[
+                ExecuteProcess(
+                    cmd=['ros2', 'daemon', 'start'],
+                    name='daemon-start',
+                    on_exit=[
+                        OpaqueFunction(function=lambda context: ready_fn())
```
jacobperron: FYI, this is the old ([soon-to-be deprecated](https://github.com/ros2/launch/pull/346)) way of doing this. Use `launch_testing.actions.ReadyToTest()` instead.

claireyywang:
CI again
linux [Build Status](https://ci.ros2.org/job/ci_linux/9274/)
linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/4961/)
osx [Build Status](https://ci.ros2.org/job/ci_osx/7588/)
windows [Build Status](https://ci.ros2.org/job/ci_windows/9195/)
windows-container [Build Status](https://ci.ros2.org/job/ci_windows-container/82/)

Issue #424, opened 2019-12-18
ros2 commands are very slow


christianrauch:
## Bug report
The `ros2` command is very slow (more than 100 times slower than comparable ROS1 commands). E.g. on a Raspberry Pi 2, showing the help via `ros2 run -h` takes more than 7 seconds.

**Required Info:**
- Operating System: Ubuntu 18.04
- Installation type: binaries
- Version or commit hash: eloquent
- Client library (if applicable): rclpy

#### Steps to reproduce issue
1. install ROS2 on Ubuntu on a Raspberry Pi 2
2. source and run `time ros2 run -h`
3. example timing:
```
real    0m7,423s
user    0m6,863s
sys     0m0,559s
```

#### Expected behavior
The help text show be shown immediately (less than 0.1 second).

#### Actual behavior
Showing the help text will take more than 7 seconds!

#### Additional information
On a more powerful PC, the `ros2` command will take e.g. `0m0,808s`. The comparable command on ROS1 (`time rosrun -h`) will take `0m0,006s` (i.e. less than 10ms). The ROS2 version of these commands is more than 100 times slower than the ROS1 commands. Showing help or other generic information should not take more than 100ms.

wjwwood:
`rosrun` vs `ros2 run` is not a very fair comparison, because in that particular case we had to convert from bash to python. And I don't think there's much to be done about that technical point. However, it still may be slower than it needs to be, or was in ROS 1.

A better comparison would be between `rostopic -h` and `ros2 topic -h`. Maybe you care to redo your analysis on those tools?

wjwwood:
I'm going to close this because I believe the issue has been addressed. Feel free to continue to comment and I can reopen if necessary.

dirk-thomas:
@christianrauch Please see #436 which improves the performance a bit (something between 20-50% depending on the command). While that doesn't get us to the time of native shell functions it is at least a step in the right direction.

PR #436, opened 2020-01-18
only load required entry points which improves the performance

dirk-thomas:
Follow up of #424.

The changes are split over four commits to ease the review.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9081/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4828/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7432/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9019/)

### DIFF
--- a/ros2cli/ros2cli/command/__init__.py
+++ b/ros2cli/ros2cli/command/__init__.py
@@ -72,6 +72,12 @@ def add_subparsers(
     For each extension a subparser is created.
     If the extension has an ``add_arguments`` method it is being called.
 
+    This method is deprecated.
+    Use the function ``add_subparsers_on_demand`` instead.
+    Their signatures are almost identical.
+    Instead of passing the extensions the new function expects the group name
+    of these extensions.
+
     :param parser: the parent argument parser
     :type parser: :py:class:`argparse.ArgumentParser`
     :param str cli_name: name of the command line command to which the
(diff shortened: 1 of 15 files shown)


mjcarroll:
Sorry, I was only approving the first commit, didn't realize that it marks the whole thing as approved, still looking.

ros2cli/ros2cli/command/__init__.py
```diff
@@ -112,3 +125,125 @@ def add_subparsers(
...
+        self.value = ''
+
+    def __getattr__(self, name):
+        return getattr(self.value, name)
+
+    def __iter__(self):
+        return self.value.__iter__()
+
+
+def add_subparsers_on_demand(
```
mjcarroll: Since `add_subparsers` is deprecated, should there be documentation here?

dirk-thomas: Good point, added in 01616f394b849dbe1e61239da04574efeffc1106.

ivanpauno:
Cancelled hanging windows CI, here new one:
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9​02​1/)

dirk-thomas:
While the Windows build has a very high test failure count (75) that seems to match the latest nightly Windows Debug build: https://ci.ros2.org/view/nightly/job/nightly_win_deb/1504/

Merging...

ivanpauno:
I noticed that autocomplete isn't working at all in master in all the ros2 cli tools.
I reverted this together with https://github.com/ros2/ros2cli/pull/437, and it started working again.

@dirk-thomas can you take a look? If not, I can try to fix it (though I don't have much context).

PR #439, opened 2020-01-24
Request for backport of 428 to eloquent

mikaelarguedas:
Currently the generated packages fail to pass rosdep making them hard to reuse.
Ideally this would be backported and released in Dashing as well.

Note: The tutorials refering to this invalid builtool dependency explicitly would need to be updated. That'll make them valid when used with the nightly archives as well that don't list that dependency.

### DIFF
--- a/ros2pkg/ros2pkg/verb/create.py
+++ b/ros2pkg/ros2pkg/verb/create.py
@@ -114,9 +114,12 @@ class CreateVerb(VerbExtension):
                 print('[WARNING] node name can not be equal to the library name', file=sys.stderr)
                 print('[WARNING] renaming node to %s' % node_name, file=sys.stderr)
 
-        buildtool_depends = args.build_type
-        if args.build_type == 'ament_cmake' and args.library_name:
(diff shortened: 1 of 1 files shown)


mjcarroll:
Since binary packages for Eloquent Patch Release 1 have already been rolled and are being tested, this won't make it in there, but I don't see any problem with getting it in the next one.

CC: @nuclearsandwich for dashing.

nuclearsandwich:
>  Ideally this would be backported and released in Dashing as well.

If you can open the Dashing backport @mikaelarguedas we can do both in parallel.

mikaelarguedas:
> If you can open the Dashing backport mikaelarguedas we can do both in parallel.

Sure thing, https://github.com/ros2/ros2cli/pull/442

dirk-thomas:
@tfoote @mjcarroll What is the status of this PR?

mjcarroll:
Good to go, lost track of it in the patch release last week.

dirk-thomas:
The ticket also needs to be added to the project for the next Eloquent patch release to not be forgotten.

PR #442, opened 2020-01-25
[dashing] Backport of 428 'fix create_pkg dependencies for python'

mikaelarguedas:
More context at https://github.com/ros2/ros2cli/pull/439#issuecomment-578319122


ivanpauno:
@mjcarroll can you run CI?

mjcarroll:
I believe that the `Dpr` job should suitably cover it in this case.

PR #448, opened 2020-02-01
use f-string

dirk-thomas:
CI builds in ament/ament_package#112.

### DIFF
--- a/ros2param/ros2param/api/__init__.py
+++ b/ros2param/ros2param/api/__init__.py
@@ -95,8 +95,7 @@ def get_parameter_value(*, string_value):
 def call_describe_parameters(*, node, node_name, parameter_names=None):
     # create client
     client = node.create_client(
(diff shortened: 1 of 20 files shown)


ros2param/ros2param/api/__init__.py
```diff
@@ -140,16 +136,13 @@ def call_get_parameters(*, node, node_name, parameter_names):
     if response is None:
         e = future.exception()
         raise RuntimeError(
-            'Exception while calling service of node '
-            "'{args.node_name}': {e}".format_map(locals()))
+            f"Exception while calling service of node '{node_name}': {e}")
```
dirk-thomas: Great example why f-strings are better: this bug was catched by the linters (before the string contained `args.node_name` but `args` isn't a variable in the current scope.

Same below in old line 170 / new line 162.

PR #376, opened 2019-10-22
Remove ready_fn, self.proc_info

pbaughman:
The ready_fn will be deprecated in the future in favor of the ReadyToTest() action in launch_testing. See ros2/launch#346 (comment) for background information

Signed-off-by: Pete Baughman <pete.baughman@apex.ai>

PR #454, opened 2020-02-20
Fix race in ros2action test_cli

pbaughman:
This test was trying to talk to the fibonacci action server before it
was ready to be talked to.

Fix the test by having the fibonacci action server write to stdout when
it's ready and having the test wait for the stdout to start

I believe this was the cause of the test failures in https://github.com/ros2/ros2cli/pull/376 but I've found the ros2cli tests are very fragile, so I'd like to try to merge this separately so there aren't confounding factors if the tests continue to fail.

## How to reproduce the original race/failure:
If you delay the Fibonacci action server program's startup by adding
```
import time
time.sleep(5)
```
right before `main` is called [here](https://github.com/ros2/ros2cli/blob/master/ros2action/test/fixtures/fibonacci_action_server.py#L71)

. . . and also remove the retry_on_failure [here](https://github.com/ros2/ros2cli/blob/master/ros2action/test/test_cli.py#L138) you could make `test_fibonacci_info` fail every time.

Signed-off-by: Pete Baughman <peter.c.baughman@gmail.com>

### DIFF
--- a/ros2action/test/fixtures/fibonacci_action_server.py
+++ b/ros2action/test/fixtures/fibonacci_action_server.py
@@ -56,6 +56,12 @@ def main(args=None):
 
     node = FibonacciActionServer()
     try:
+        # Spend up to 2 seconds doing all the work we need to have the fibonacci
+        # action server actually talking over DDS before signaling to the test it's
+        # OK to start
+        for _ in range(20):
+            rclpy.spin_once(node, timeout_sec=0.1)
+        print('fibonacci_action_server has started')
         rclpy.spin(node)
     except KeyboardInterrupt:
         print('server stopped cleanly')
--- a/ros2action/test/test_cli.py
+++ b/ros2action/test/test_cli.py
@@ -21,9 +21,11 @@ import unittest
 from launch import LaunchDescription
 from launch.actions import ExecuteProcess
 from launch.actions import OpaqueFunction
+from launch.actions import RegisterEventHandler
 
 import launch_testing
 import launch_testing.asserts
+import launch_testing.event_handlers
 import launch_testing.markers
 import launch_testing.tools
 import launch_testing_ros.tools
@@ -41,6 +43,16 @@ def generate_test_description(rmw_implementation, ready_fn):
     path_to_action_server_executable = os.path.join(
         os.path.dirname(__file__), 'fixtures', 'fibonacci_action_server.py'
     )
+
+    fibonacci_action_server = ExecuteProcess(
+        cmd=[sys.executable, path_to_action_server_executable],
+        additional_env={
+            'RMW_IMPLEMENTATION': rmw_implementation,
+            'PYTHONUNBUFFERED': '1'
+        },
+        log_cmd=True
+    )
+
     return LaunchDescription([
         # Always restart daemon to isolate tests.
         ExecuteProcess(
@@ -51,15 +63,23 @@ def generate_test_description(rmw_implementation, ready_fn):
                     cmd=['ros2', 'daemon', 'start'],
                     name='daemon-start',
                     on_exit=[
-                        ExecuteProcess(
-                            cmd=[sys.executable, path_to_action_server_executable],
-                            additional_env={'RMW_IMPLEMENTATION': rmw_implementation}
-                        ),
-                        OpaqueFunction(function=lambda context: ready_fn())
+                        fibonacci_action_server,
                     ],
                     additional_env={'RMW_IMPLEMENTATION': rmw_implementation}
                 )
             ]
+        ),
+
+        # Wait until the fibonacci action server process has created its action server
+        # before starting the tests, otherwise they will race with the process start-up
+        RegisterEventHandler(
+            launch_testing.event_handlers.StdoutReadyListener(
+                target_action=fibonacci_action_server,
+                ready_txt='fibonacci_action_server has started',
+                actions=[
+                    OpaqueFunction(function=lambda context: ready_fn())
+                ]
+            )
         ),
     ])
 
@@ -135,7 +155,6 @@ class TestROS2ActionCLI(unittest.TestCase):
             strict=False
         )
 
-    @launch_testing.markers.retry_on_failure(times=5)
     def test_fibonacci_info(self):
         with self.launch_action_command(arguments=['info', '/fibonacci']) as action_command:
             assert action_command.wait_for_shutdown(timeout=10)


ros2action/test/test_cli.py
```diff
@@ -135,7 +155,6 @@ def test_info_on_nonexistent_action(self):
             strict=False
         )
 
-    @launch_testing.markers.retry_on_failure(times=5)
```
pbaughman: I believe this was a workaround for the race condition that I've fixed above and is no longer necessary

ros2action/test/fixtures/fibonacci_action_server.py
```diff
@@ -56,6 +56,12 @@ def main(args=None):
 
     node = FibonacciActionServer()
     try:
+        # Spend up to 2 seconds doing all the work we need to have the fibonacci
+        # action server actually talking over DDS before signaling to the test it's
+        # OK to start
+        for _ in range(20):
```
pbaughman: There may be a way to spin_until_future_complete here, but I just don't know.  Is there a future I can wait on that indicates the action server is up, ready, and can be communicated with by other processes before I print out 'fibonacci_action_server has started' ??

pbaughman:
. . . Now i'm seeing failures locally with rmw_cyclonedds_cpp on my branch and also when I build/test using 'master'. . . It feels totally impossible to make any progress 😭 

```
[213.948s] test_fibonacci_info[rmw_cyclonedds_cpp] (test_cli.TestROS2ActionCLI) ... ok                     
[213.948s] test_fibonacci_info_count[rmw_cyclonedds_cpp] (test_cli.TestROS2ActionCLI) ... ok               
[213.949s] test_fibonacci_info_with_types[rmw_cyclonedds_cpp] (test_cli.TestROS2ActionCLI) ... ok          
[213.949s] test_info_on_nonexistent_action[rmw_cyclonedds_cpp] (test_cli.TestROS2ActionCLI) ... ok         
[213.949s] test_list[rmw_cyclonedds_cpp] (test_cli.TestROS2ActionCLI) ... FAIL                             
[213.949s] test_list_count[rmw_cyclonedds_cpp] (test_cli.TestROS2ActionCLI) ... FAIL                       
[213.949s] test_list_with_types[rmw_cyclonedds_cpp] (test_cli.TestROS2ActionCLI) ... FAIL                  
[213.949s] test_send_fibonacci_goal[rmw_cyclonedds_cpp] (test_cli.TestROS2ActionCLI) ... FAIL              
[213.950s] test_send_fibonacci_goal_with_feedback[rmw_cyclonedds_cpp] (test_cli.TestROS2ActionCLI) ... FAIL
```

pbaughman:
@hidmic Side-note:  I'm not sure if this is being tracked anywhere, but the pytest plugin for launch tests makes it really hard to see what has failed here.  Instead of the results showing 9 test cases x however many DDS implementations there are, we just get one failing test called test_cli.py and you have to cat the logs to see what actually happened.  You can't figure it out programmatically.

pbaughman:
Does

```
PRESCstReaderCollator_storeSampleData:deserialize sample error in topic 'DISCParticipant' with type 'DISCParticipantP...storeSampleData:deserialize sample error in topic 'DISCParticipant' with type 'DISCParticipantParameter'\n/fibonacci\n",
```

mean anything to anybody?

pbaughman:
@sloretz Maybe we can run CI on this and see if we get lucky?  I'm so sorry about all of the bad CI jobs we're spinning up for this issue.  This package is extremely difficult for me to wrangle.

dirk-thomas:
> Fix the test by having the fibonacci action server write to stdout when
it's ready and having the test wait for the stdout to start

I don't think this is the right way to fix this test. We couldn't recommend the same approach to a user trying to use actions in their own code.

pbaughman:
@dirk-thomas Can you suggest something different?  The Fibonacci action server program was created specifically as a test fixture for this test, and must provide some mechanism to indicate that it has finished starting up or else the test will race and fail.

>We couldn't recommend the same approach to a user trying to use actions in their own code.

. . . but the current test starts the action server in a process and attempts to communicate with it before it's finished starting up with no feedback whatsoever.  Would you recommend **that** approach to a user?

pbaughman:
@dirk-thomas Would you accept a PR where we disable the test because it's broken?  That would satisfy my need to get [this PR](https://github.com/ros2/ros2cli/pull/376) unstuck

pbaughman:
@dirk-thomas Can you please further clarify your objection to this fix?  The code in question is a test fixture meant to facilitate the testing of the ros2 actions cli tool.  In order for the tests to start, the test fixture must be up and running.  The mechanism I've used here is a message printed over stdout, which I don't think is very unreasonable.

The test fixture in question is not meant as a demo or a suggestion to any user about how to create a ROS2 action server - it's a dummy process for the test to run against.

I would be happy to use some other synchronization method suggested, but thus far no suggestion has been offered.

I see that this issue is also causing CI failures in [this PR](https://github.com/ros2/ros2cli/pull/421) as well as [my other PR](https://github.com/ros2/ros2cli/pull/376).  If we can't agree on a way to improve the test, I suggest disabling it so we can make progress elsewhere.

pbaughman:
@sloretz Do you have any opinions about this?  It should fix the test that's blocking https://github.com/ros2/ros2cli/pull/376 from being merged

jacobperron:
@pbaughman I spoke with @wjwwood offline about this issue and we agree that the correct thing to do in this situation (with the CLI daemon) is to retry the affected tests until discovery happens. I believe the bug in this case is that the `retry_on_failure` decoractor is not waiting between retries and so not enough time is left for discovery to happen. I've proposed a `delay` parameter in https://github.com/ros2/launch/pull/390 that we should be able to use here to help with the flaky tests.

jacobperron:
As an alternative to this PR, I've updated all the tests involving the CLI daemon in https://github.com/ros2/ros2cli/pull/459.

pbaughman:
@jacobperron Sure.  My only dissapointment is that the whole point of having a "Ready To Test" action as part of the launch description was that the test author would be forced to think about things like this and wouldn't have tests that try to run before all of their preconditions are met.  This was a big problem with test actions in launch files using `rostest` in ROS1.  If we're going to put "ReadyToTest" in the wrong spot and then fill the tests with retry logic, it doesn't feel like we've made much progress.

Having said that - let's do whatever we need to do to unstick this.

jacobperron:
@pbaughman I agree with the sentiment, but I think these CLI tests are an edge case because of the underlying daemon that they use. For example, in a different scenario where we have access to the node objects we are testing, it's more feasible to condition a launch event on the nodes discovering each other.

wjwwood:
> If we're going to put "ReadyToTest" in the wrong spot and then fill the tests with retry logic, it doesn't feel like we've made much progress.

Well, this is still an improvement. It just so happens that the thing we're testing for _is_ the precondition. We could write it like this:

- run fixture
- wait for `ros2 action info` to show right info or fail
- ready_fn
- assert true

But instead we're writing it like this:

- run fixture
- ready_fn
- retry until `ros2 action info` shows the the right info or fail

Other tests will be able to do:

- run fixture
- wait for something like `ros2 topic list`/`ros2 service list`/`ros2 action info ...`/etc... to show expected info, or fail
- ready_fn
- check something more complicated, like actually call the service/action

wjwwood:
Also, we could additionally (to what @jacobperron proposes in https://github.com/ros2/ros2cli/pull/459) wait for the server to be available (using an action client), but that would influence the output of `ros2 action info` so we'd have to retry until the client disappears (which is asynchronous as well).

AFAICT, the only reason this pull request helps is that it takes some time to complete your new code in the action server and that allows discovery to happen. If you artificially added a lag to the daemon so it did not report new discovery information until sometime in the future you'd still see this problem with this pull request (imo).

pbaughman:
@wjwwood just an aside:

> * wait for something like ros2 topic list/ros2 service list/ros2 action info .../etc... to show expected info, or fail

I had hoped to write some launch_testing event handlers to make it easy to wait for a certain topic to be available or for data to appear on a certain topic - as these are common things you want to wait for before starting a test, but I never got the chance when I was contracting at Apex.  I'm hopeful that they or someone else will want to make that improvement at some point.

pbaughman:
Closed in favor of https://github.com/ros2/ros2cli/pull/459

PR #459, opened 2020-02-25
Add delay when retrying tests involving the CLI daemon

jacobperron:
This is to give time for discovery to happen between the daemon node and the test fixture nodes.

Alternative to https://github.com/ros2/ros2cli/pull/454

Depends on https://github.com/ros2/launch/pull/390


jacobperron:
rebased on master. CI is here: https://github.com/ros2/launch/pull/390#issuecomment-591141670

PR #460, opened 2020-02-26
Removed rmw_implementation Python dependency

ahcorde:
According with this discussion https://github.com/ros2/rmw_implementation/pull/84#pullrequestreview-364451405 It makes sense to remove the Python dependency and call `ament_index_python.get_resources() where it's needed

### DIFF
--- a/ros2action/test/test_cli.py
+++ b/ros2action/test/test_cli.py
@@ -30,13 +30,15 @@ import launch_testing_ros.tools
 
 import pytest
 
-from rmw_implementation import get_available_rmw_implementations
+import ament_index_python
 
 import yaml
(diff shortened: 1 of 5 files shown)


dirk-thomas:
How about placing a function into `rclpy` beside [get_rmw_implementation_identifier](https://github.com/ros2/rclpy/blob/b27ddc6dee902eb4a7e0c383731acfddf118a3f5/rclpy/rclpy/utilities.py#L65) and it from here?

ahcorde:
closing this in PR in favour of these others https://github.com/ros2/ros2cli/pull/461 and ros2/rclpy#517

Issue #453, opened 2020-02-19
Add a warning in ros2node when dealing with non-uniquely-named nodes


emersonknapp:
## Feature request

Related to https://github.com/ros2/ros2cli/issues/279 - but a much simpler feature request

* When running `ros2 node list`, add a top level warning and a small note next to all nodes that share a name, e.g.

```
$ ros2 node list
# WARNING: There are nodes in the graph that share a name - it is recommended that you use namespaces or different names to differentiate nodes
/talker (not uniquely named)
/listener
/talker (not uniquely named)
```

* When running `ros2 node info`, add a note that you are receiving information about only one node with the given name, if multiple are present, e.g.

```
$ ros2 node info /talker
# WARNING: there is more than one node found that has the name /talker, you are seeing information about only one of them. It is not guaranteed that you will see information about the same node every time
etc...
```

This should help make the non-unique situation more obvious, even though it doesn't fix anything.

#### Implementation considerations
We should already have all the information we need, a sort and pass over the node list should do for both cases.

hidmic:
@emersonknapp as far as I can see, this issue has been fully addressed. I'll close it.

PR #462, opened 2020-02-27
Add warning and annotation to `ros2 node list` when there are non-unique nodes

emersonknapp:
First half of https://github.com/ros2/ros2cli/issues/453 
First half of https://github.com/ros-tooling/aws-roadmap/issues/196

Add a simple warning, and annotate which nodes it applies to, when `ros2 node list` lists multiple nodes with the same exact name.  Within the "request for design" discussion https://github.com/ros2/design/issues/187, there was confusion expressed about where multiple names come from, whether they are artifacts of the printing process. This at least clears that up a little bit -it does not change the functionality at all.

### DIFF
--- a/ros2node/ros2node/api/__init__.py
+++ b/ros2node/ros2node/api/__init__.py
@@ -13,6 +13,8 @@
 # limitations under the License.
 
 from collections import namedtuple
+from typing import Any
+from typing import List
 
 from rclpy.action.graph import get_action_client_names_and_types_by_node
 from rclpy.action.graph import get_action_server_names_and_types_by_node
@@ -49,6 +51,11 @@ def parse_node_name(node_name):
     if namespace == '':
         namespace = '/'
     return NodeName(node_basename, namespace, full_node_name)
+
+
+def has_duplicates(values: List[Any]) -> bool:
+    """Find out if there are any exact duplicates in a list of strings."""
+    return len(set(values)) < len(values)
 
 
 def get_node_names(*, node, include_hidden_nodes=False):
--- a/ros2node/ros2node/verb/list.py
+++ b/ros2node/ros2node/verb/list.py
@@ -12,9 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import sys
+
 from ros2cli.node.strategy import add_arguments
 from ros2cli.node.strategy import NodeStrategy
 from ros2node.api import get_node_names
+from ros2node.api import has_duplicates
 from ros2node.verb import VerbExtension
 
 
@@ -37,4 +40,8 @@ class ListVerb(VerbExtension):
         if args.count_nodes:
             print(len(node_names))
         elif node_names:
-            print(*sorted(n.full_name for n in node_names), sep='\n')
+            sorted_names = sorted(n.full_name for n in node_names)
+            if has_duplicates(sorted_names):
+                print('WARNING: Be aware that are nodes in the graph that share an exact name, '
+                      'this can have unintended side effects.', file=sys.stderr)
+            print(*sorted_names, sep='\n')
--- a/ros2node/test/test_node.py
+++ b/ros2node/test/test_node.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from ros2node.api import has_duplicates
 from ros2node.api import parse_node_name
 
 
@@ -34,3 +35,11 @@ def test_parse_node_name():
     assert name.full_name == '/ns/talker'
     assert name.namespace == '/ns'
     assert name.name == 'talker'
+
+
+def test_has_duplicates():
+    assert not has_duplicates([])
+    assert not has_duplicates(['ns_foo/foo', 'ns_foo/bar'])
+    assert has_duplicates(['ns_foo/foo'] * 2)
+    assert has_duplicates(['ns_foo/foo'] * 3)
+    assert has_duplicates(['ns_foo/foo', 'ns_foo/foo', 'ns_foo/bar'])


emersonknapp:
@ros2/aws-oncall - please run this CI job
Gist: https://gist.githubusercontent.com/emersonknapp/aff09b99d9b062699155ddb86a970b43/raw/8c63209a6cb44e48c7cc71f42e588a10ece4692d/ros2.repos
BUILD args: --packages-up-to ros2node
TEST args: --packages-select ros2node
Job: ci_launcher

piraka9011:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9481/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5209/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7755/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9407/)
* Windows-container [Build Status](http://ci.ros2.org/job/ci_windows-container/247/)

emersonknapp:
@mjcarroll perhaps, review?

ros2node/test/test_node.py
```diff
@@ -34,3 +36,28 @@ def test_parse_node_name():
     assert name.full_name == '/ns/talker'
     assert name.namespace == '/ns'
     assert name.name == 'talker'
+
+
+def test_duplicate_names():
```
dabonnie: Any sanity checking needed, e.g., empty list input to ```annotate_duplicate_nodes```? What about using more than 2 nodes?

ros2node/test/test_node.py
```diff
@@ -34,3 +36,28 @@ def test_parse_node_name():
...
+    assert not any_duplicates
+    assert printable_names == [
+        'ns_foo/bar',
+        'ns_foo/foo',
+    ]
+
+    all_unique_names = [
+        NodeName('foo', 'ns_foo', 'ns_foo/foo'),
+        NodeName('foo', 'ns_foo', 'ns_foo/foo'),
+    ]
```
zmichaels11: I think this would be more readable as 2 separate tests;
* `test_annotate_duplicate_nodes_names_all_unique`
* `test_annotate_duplicate_nodes_names_with_duplicates`

emersonknapp:
@chapulina who should we ask for OSRF-side review on this PR?

ros2node/test/test_node.py
```diff
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from ros2node.api import annotate_duplicate_nodes
+from ros2node.api import NodeName
 from ros2node.api import parse_node_name
```
zmichaels11: Is this the right order? I think `N` < `a` in ASCII

emersonknapp: It is the right order, flake8-import-order does case-insensitive alphasort

zmichaels11: works for me

emersonknapp:
@hidmic can you review this?

ros2node/ros2node/api/__init__.py
```diff
@@ -47,6 +49,32 @@ def parse_node_name(node_name):
     return NodeName(node_basename, namespace, full_node_name)
 
 
+def annotate_duplicate_nodes(node_names: List[NodeName]) -> Tuple[bool, List[str]]:
+    """
+    Annotate a list of node names with a warnings if names are not unique.
```
wjwwood:
```suggestion
    Annotate a list of node names with warnings if names are not unique.
```

or

```suggestion
    Annotate a list of node names with a warning if names are not unique.
```

ros2node/ros2node/api/__init__.py
```diff
@@ -47,6 +49,32 @@ def parse_node_name(node_name):
...
+    """
+    annotation_string = ' (non-unique name)'
+
+    any_duplicates = False
+    name_count = {}
+    for n in node_names:
+        count = name_count.get(n.full_name, 0) + 1
+        if count > 1:
+            any_duplicates = True
+        name_count[n.full_name] = name_count.get(n.full_name, 0) + 1
```
wjwwood:
```suggestion
        name_count[n.full_name] = count
```

Why not this?

ros2node/ros2node/api/__init__.py
```diff
@@ -47,6 +49,32 @@ def parse_node_name(node_name):
...
+
+    sorted_names = sorted(
```
wjwwood: Would it not be better to return a list in the same order as the input list, i.e. do not sort it?

ros2node/ros2node/verb/list.py
```diff
@@ -37,4 +39,8 @@ def main(self, *, args):
         if args.count_nodes:
             print(len(node_names))
         elif node_names:
-            print(*sorted(n.full_name for n in node_names), sep='\n')
+            has_duplicates, sorted_names = annotate_duplicate_nodes(node_names)
```
wjwwood: This seems like it would be annoying for people using this as a script. They're expecting a single complete node name per line in stdout, but now we are unconditionally adding something to some of the names. Maybe there needs to be an option to avoid this annotation, even if there are duplicate names. The warning to stderr is fine in any case.

ros2node/ros2node/api/__init__.py
emersonknapp: Fixed

ros2node/ros2node/verb/list.py
emersonknapp: I hadn't thought of that use case - do you think the per-node annotation is valuable? Or, should we just leave it with the top level warning and call it a day? I could be convinced either way. Since the names are sorted anyways, probably there's not that much additional value added

ros2node/ros2node/api/__init__.py
emersonknapp: you're right, it's better that way

emersonknapp:
@wjwwood uploaded a simplified version of this PR

ros2node/ros2node/verb/list.py
wjwwood: The warning only seems fine to me.

wjwwood:
I approved the changes, can one of you all do CI again (since there were new changes) and then I (or anyone really) can merge it.

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9641/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5328/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7869/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9538/)

emersonknapp:
Hypothesizing it's flaky preexisting tests, retrying osx build 

[Build Status](https://ci.ros2.org/job/ci_osx/7871/)

emersonknapp:
Hypothesis confirmed, @wjwwood i think we're good to go?

PR #463, opened 2020-03-02
Add a warning to ros2 node info when there is more than one node with the same name

emersonknapp:
Second half of https://github.com/ros2/ros2cli/issues/453 
Second half of https://github.com/ros-tooling/aws-roadmap/issues/196

Add a simple warning for `ros2 node info` when there is more than one node with the same name. This does not change functionality at all, just informs users of a potentially confusing situation.

Signed-off-by: Emerson Knapp <emerson.b.knapp@gmail.com>

### DIFF
--- a/ros2node/ros2node/verb/info.py
+++ b/ros2node/ros2node/verb/info.py
@@ -46,7 +46,12 @@ class InfoVerb(VerbExtension):
     def main(self, *, args):
         with NodeStrategy(args) as node:
             node_names = get_node_names(node=node, include_hidden_nodes=args.include_hidden)
-        if args.node_name in (n.full_name for n in node_names):
+        count = [n.full_name for n in node_names].count(args.node_name)
+        if count > 1:
+            print('# WARNING: There are {} nodes in the graph with this exact name "{}". '
+                  'You are seeing information about only one of them.'.format(
+                    count, args.node_name))
+        if count > 0:
             with DirectNode(args) as node:
                 print(args.node_name)
                 subscribers = get_subscriber_info(
--- a/ros2node/ros2node/verb/info.py
+++ b/ros2node/ros2node/verb/info.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import sys
 
 from ros2cli.node.direct import DirectNode
 from ros2cli.node.strategy import add_arguments
@@ -22,6 +23,7 @@ from ros2node.api import get_publisher_info
 from ros2node.api import get_service_client_info
 from ros2node.api import get_service_server_info
 from ros2node.api import get_subscriber_info
+from ros2node.api import INFO_NONUNIQUE_WARNING_TEMPLATE
 from ros2node.api import NodeNameCompleter
 from ros2node.verb import VerbExtension
 
@@ -48,9 +50,9 @@ class InfoVerb(VerbExtension):
             node_names = get_node_names(node=node, include_hidden_nodes=args.include_hidden)
         count = [n.full_name for n in node_names].count(args.node_name)
         if count > 1:
-            print('# WARNING: There are {} nodes in the graph with this exact name "{}". '
-                  'You are seeing information about only one of them.'.format(
-                    count, args.node_name))
+            print(
+                INFO_NONUNIQUE_WARNING_TEMPLATE.format(num_nodes=count, node_name=args.node_name),
+                file=sys.stderr)
         if count > 0:
             with DirectNode(args) as node:
                 print(args.node_name)
--- a/ros2node/ros2node/verb/info.py
+++ b/ros2node/ros2node/verb/info.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
 import sys
 
 from ros2cli.node.direct import DirectNode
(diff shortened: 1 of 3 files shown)


emersonknapp:
@chapulina ping for routing to the correct reviewers

ros2node/ros2node/verb/info.py
```diff
@@ -46,7 +46,12 @@ def add_arguments(self, parser, cli_name):
     def main(self, *, args):
         with NodeStrategy(args) as node:
             node_names = get_node_names(node=node, include_hidden_nodes=args.include_hidden)
-        if args.node_name in (n.full_name for n in node_names):
+        count = [n.full_name for n in node_names].count(args.node_name)
+        if count > 1:
+            print('# WARNING: There are {} nodes in the graph with this exact name "{}". '
```
hidmic: @emersonknapp nit: 
```suggestion
            print('WARNING: There are {} nodes in the graph with this exact name "{}". '
```
(just to be consistent with other CLI warnings' format)

hidmic: Well, it's not consistent at all, but at least this way we don't introduce yet another variation.

ros2node/ros2node/verb/info.py
```diff
@@ -46,7 +46,12 @@ def add_arguments(self, parser, cli_name):
...
+                  'You are seeing information about only one of them.'.format(
+                    count, args.node_name))
```
hidmic: @emersonknapp consider printing to `stderr` instead of `stdout`.

emersonknapp: good idea

emersonknapp:
@hidmic I've added a test to expect this warning line output in a launch scenario with this node. I was thinking of adding an inverse test, "this line is _not_ present when running against a normal node" - but I couldn't think of how to do that "expectation of absence" - any ideas? Also, maybe it's not necessary because the existing tests handle the regular-case scenario. What do you think?

emersonknapp:
@hidmic can you review this?

emersonknapp:
@ros2/aws-oncall - please run this CI job
Gist: https://gist.githubusercontent.com/emersonknapp/3f56000a8f88a272334b7c4c2e53828d/raw/cecff37a502392f6efb96f404c5bdd96b31a2ad2/ros2.repos
BUILD args: --packages-up-to ros2node
TEST args: --packages-select ros2node
Job: ci_launcher

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9588/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5277/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7822/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9503/)

emersonknapp:
Restarting after updating branch
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9590/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5279/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7825/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9505/)

emersonknapp:
I've just learned that `rmw_cyclonedds_cpp` uses a `std::set<>` under the hood to return node names, so it will not return a list with multiple nodes of the same name, therefore these checks don't help us in CycloneDDS - @hidmic do you have thoughts on this? I'm wondering if I should go in and change the rmw_cyclonedds_cpp implementation to use a `std::vector` instead, or if we should disable these tests for Cylone, or a third option (don't print this warning?)

emersonknapp:
After updating rmw_cyclonedds to provide nodes with duplicate names

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9612/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5308/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7835/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9517/)
Triggering a new build of ci_windows

ros2node/ros2node/verb/info.py
```diff
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import sys
```
wjwwood: You should have a blank line between the first import and the copyright/license header, as it was before.

```suggestion

import sys
```

wjwwood: I tried to fix this up myself, but it won't let me push to your branch.

emersonknapp: Yeah it's on our dev fork. Added the commit. Should we update ament-copyright to catch this particular nit so you don't have to be the human spellcheck? =P

hidmic:
Alright, looking good! Thanks for your contribution @emersonknapp, merging.

ros2node/ros2node/verb/info.py
wjwwood: To be honest, I thought that PEP8 (flake8) would catch it.

emersonknapp: It seems we are not enforcing it, and as best I can tell it's not one of the ones we're disabling in ament_flake8.ini, most of those are docstring-related

wjwwood: It's probably a corner case, since flake8 assumes it's a comment about the import statement, which it isn't, but it cannot know that. I guess the copyright checker could assert an empty line after the statement.

PR #467, opened 2020-03-19
Add /usr/lib/<pkg> and /usr/libexec/<pkg> to search paths

sloretz:
This is a proposal following https://github.com/ros2/ci/issues/400#issuecomment-601319086. This PR adds `<ws>/usr/lib/<pkg>` and `<ws>/usr/libexec/<pkg>` to the search paths used by `ros2 run` to comply with the [Linux Filesystem Hierarchy Standard](https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.pdf#[{%22num%22%3A242%2C%22gen%22%3A0}%2C{%22name%22%3A%22XYZ%22}%2C72%2C720%2Cnull]). It looks like the current practice of `<ws>/lib/<pkg>` doesn't match, but I left that in to make this be an enhancement rather than a disruptive change.

### DIFF
--- a/ros2pkg/ros2pkg/api/__init__.py
+++ b/ros2pkg/ros2pkg/api/__init__.py
@@ -41,17 +41,22 @@ def get_executable_paths(*, package_name):
     prefix_path = get_prefix_path(package_name)
     if prefix_path is None:
         raise PackageNotFound(package_name)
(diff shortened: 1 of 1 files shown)


audrow:
Hey @sloretz, are you still interested in getting this feature in? Perhaps you or I can make PR with a small addition to REP 122.

PR #469, opened 2020-03-23
[ros2param] Add timeout to ros2param list

DongheeYe:
If the node use default node option ( start_parameter_services is ture) and use not spin() but spin_until_future_complete(), then ros2param list send a request to the node but the callback of parameter service list parameter request can not be called. and ros2param spin_until_future_complete can not be finished.

### DIFF
--- a/ros2param/ros2param/verb/list.py
+++ b/ros2param/ros2param/verb/list.py
@@ -87,7 +87,7 @@ class ListVerb(VerbExtension):
 
             # wait for all responses
             for future in futures.values():
-                rclpy.spin_until_future_complete(node, future)
+                rclpy.spin_until_future_complete(node, future, timeout_sec=1.0)
 
             # print responses
             for node_name in sorted(futures.keys()):


ros2param/ros2param/verb/list.py
```diff
@@ -87,7 +87,7 @@ def main(self, *, args):  # noqa: D102
 
             # wait for all responses
             for future in futures.values():
-                rclpy.spin_until_future_complete(node, future)
+                rclpy.spin_until_future_complete(node, future, timeout_sec=1.0)
```
fujitatomoya: not sure just ignoring timeout is good enough here. i would add check the FutureReturnCode, and if gets TIMEOUT, notify the user what node's parameter cannot be queried. i would like to hear from others.

jacobperron: I agree, if there is a problem communicating with one of the parameter servers then we should warn the user, but I think we are getting some feedback with the "exception" printed at L104.

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10076/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5663/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8194/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9978/)

PR #472, opened 2020-03-25
[ros2doctor] Only report topic warnings if publisher or subscription count is zero

jacobperron:
Before the tool was generating many false positives as it is a valid configuration
to have an unequal number of publishers and subscriptions on a given topic.

This change makes it so we only provide a warning if one of the counts is zero and the
other is not. Although this is still a legitimate configuration, it seems more likely to be
a problem and worth reporting.

### DIFF
--- a/ros2doctor/ros2doctor/api/topic.py
+++ b/ros2doctor/ros2doctor/api/topic.py
@@ -49,10 +49,10 @@ class TopicCheck(DoctorCheck):
             for topic in to_be_checked:
                 pub_count = node.count_publishers(topic)
                 sub_count = node.count_subscribers(topic)
-                if pub_count > sub_count:
+                if pub_count > 0 and sub_count == 0:
                     doctor_warn(f'Publisher without subscriber detected on {topic}.')
                     result.add_warning()
-                elif pub_count < sub_count:
+                elif sub_count > 0 and pub_count == 0:
                     doctor_warn(f'Subscriber without publisher detected on {topic}.')
                     result.add_warning()
         return result


dirk-thomas:
> Although this is still a legitimate configuration, it seems more likely to be
a problem and worth reporting.

Should this be reported as a warning? I would rather lean to no - otherwise a valid system will print plenty of warnings which will teach users to ignore them anyway.

claireyywang:
> > Although this is still a legitimate configuration, it seems more likely to be
> > a problem and worth reporting.
> 
> Should this be reported as a warning? I would rather lean to no - otherwise a valid system will print plenty of warnings which will teach users to ignore them anyway.

I think it makes sense to remove the warnings if valid system could also have unequal # of pub/subs. Is there a illegitimate situation I could check instead, or is it best to remove this check altogether?

jacobperron:
IMO, it's still valuable to have a tool to detect dangling subscriptions or publishers (ie. when the count of one is zero). Maybe there should be an option to whitelist topics or turn the check off altogether for those systems that consider it valid. @claireyywang Is there a mechanism for passing in a configuration file to ros2doctor or have you thought about it?

claireyywang:
@jacobperron Currently there's only a variable `whitelist` that holds some topic names, but yes passing a config file sounds like a much better mechanism. We can merge this for now and I'll propose another PR with config file option later this/next week.

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9893/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5510/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8050/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9749/)

Issue #480, opened 2020-04-09
ros2param test failures (cross-platform)


jacobperron:
## Bug report

**Required Info:**

- Operating System:
  - All
- Installation type:
  - source
- Version or commit hash:
  - c9e9e2f9e932c8775d59aff3a5ac833f918afdbd
- DDS implementation:
  - N/A

#### Steps to reproduce issue

See nightly failures:

- https://ci.ros2.org/view/nightly/job/nightly_linux_repeated/1836/testReport/junit/ros2param.src.ros2.ros2cli.ros2param.test.test_verb_dump/TestVerbDump/test_verb_dump/
- https://ci.ros2.org/view/nightly/job/nightly_linux_repeated/1836/testReport/junit/ros2param.src.ros2.ros2cli.ros2param.test.test_verb_dump/TestVerbDump/test_verb_dump_invalid_path/
- https://ci.ros2.org/view/nightly/job/nightly_linux_repeated/1836/testReport/junit/ros2param.src.ros2.ros2cli.ros2param.test.test_verb_dump/TestVerbDump/test_verb_dump_print/

#### Expected behavior

Tests pass

#### Actual behavior

Tests fail

#### Additional information
Roughly, it looks like the failures started on April 5th ([based on the nightly Linux repeated build](https://ci.ros2.org/view/nightly/job/nightly_linux_repeated/1832/)).

jacobperron:
It looks like the failures are because not enough time is given for discover to happen between the test node and the CLI node (ie. the daemon).

If I switch the RMW implementation to `rmw_connext_cpp`, then I am able to reproduce reliably (I think because it takes longer for Connext to initialize).

jacobperron:
#481 isn't the fix we were looking for. Now we can see a timeout happening during the newly added wait for discovery: https://ci.ros2.org/view/nightly/job/nightly_linux-aarch64_repeated/1158/testReport/ros2param.src.ros2.ros2cli.ros2param.test.test_verb_dump/TestVerbDump/test_verb_dump/

jacobperron:
I can reproduce the timeouts locally by first starting the daemon with an RMW implementation that is different than the one used by the tests:

```
RMW_IMPLEMENTATION=rmw_connext_cpp ros2 daemon start
```

And then  running the tests
```
colcon test --packages-select ros2param
```

hidmic:
[...]

Issue #482, opened 2020-04-11
Test failures on unexpected output


rotu:
## Bug report

As originally mentioned in https://github.com/ros2/rmw_cyclonedds/pull/147 by @hidmic:

test failures as seen in https://ci.ros2.org/view/nightly/job/nightly_osx_debug/1588/testReport/.
> 
> CI up to `ros2topic`, `ros2action` and `ros2service`:
> 
> * Debug MacOS [![Build Status](https://camo.githubusercontent.com/a082fdbcc16377a551e96b6de411349a847e093e/687474703a2f2f63692e726f73322e6f72672f6275696c645374617475732f69636f6e3f6a6f623d63695f6f7378266275696c643d38313933)](http://ci.ros2.org/job/ci_osx/8193/)

> This pull request registers regular expressions for output filtering in launch tests that use `rmw_cyclonedds_cpp` as their RMW implementation of choice. It aims to solve test failures as seen in https://ci.ros2.org/view/nightly/job/nightly_osx_debug/1588/testReport/.
> 
> CI up to `ros2topic`, `ros2action` and `ros2service`:
> 
> * Debug MacOS [![Build Status](https://camo.githubusercontent.com/a082fdbcc16377a551e96b6de411349a847e093e/687474703a2f2f63692e726f73322e6f72672f6275696c645374617475732f69636f6e3f6a6f623d63695f6f7378266275696c643d38313933)](http://ci.ros2.org/job/ci_osx/8193/)
> 
> * https://ci.ros2.org/view/nightly/job/nightly_osx_debug/1588/testReport/ros2lifecycle.src.ros2.ros2cli.ros2lifecycle.test/test_cli/test_cli/
> * https://ci.ros2.org/view/nightly/job/nightly_osx_debug/1588/testReport/ros2service.src.ros2.ros2cli.ros2service.test/test_cli/test_cli/
> * https://ci.ros2.org/view/nightly/job/nightly_osx_debug/1588/testReport/ros2topic.src.ros2.ros2cli.ros2topic.test/test_cli/test_cli/
> * https://ci.ros2.org/view/nightly/job/nightly_osx_debug/1588/testReport/ros2action.src.ros2.ros2cli.ros2action.test/test_cli/test_cli/

Although not seen in the test failure messages, the failures are caused by unexpected output from Cyclone DDS. In particular, "`using network interface en0 (udp/192.168.1.120) selected arbitrarily from: en0, en1`". This is followed by the output the program expects.

[...]

PR #485, opened 2020-04-13
[ros2param] Convert test_verb_dump into launch test

jacobperron:
Fixes https://github.com/ros2/ros2cli/issues/480

The actual tests are the same, except with the use of launch_testing we ensure the CLI daemon
is restarted between tests. This follows a similar pattern as the other ros2cli tests.

I've left a TODO to test with all RMW implementations. I did not enable this now  since I do
not expect the tests to pass with all RMW implementations.
See https://github.com/ros2/ros2cli/issues/482

### DIFF
--- a/ros2param/package.xml
+++ b/ros2param/package.xml
@@ -20,6 +20,10 @@
   <test_depend>ament_flake8</test_depend>
   <test_depend>ament_pep257</test_depend>
   <test_depend>ament_xmllint</test_depend>
+  <test_depend>launch</test_depend>
+  <test_depend>launch_ros</test_depend>
+  <test_depend>launch_testing</test_depend>
+  <test_depend>launch_testing_ros</test_depend>
   <test_depend>python3-pytest</test_depend>
 
   <export>
--- a/ros2param/test/test_verb_dump.py
+++ b/ros2param/test/test_verb_dump.py
@@ -28,6 +28,7 @@ import launch_testing.actions
 import launch_testing.asserts
 import launch_testing.markers
 import launch_testing.tools
+import launch_testing_ros.tools
 
 import pytest
 
@@ -102,6 +103,9 @@ class TestVerbDump(unittest.TestCase):
         proc_output,
         rmw_implementation
     ):
+        rmw_implementation_filter = launch_testing_ros.tools.basic_output_filter(
+            filtered_rmw_implementation=rmw_implementation
+        )
 
         @contextlib.contextmanager
         def launch_param_dump_command(self, arguments):
@@ -114,7 +118,8 @@ class TestVerbDump(unittest.TestCase):
                 output='screen'
             )
             with launch_testing.tools.launch_process(
-                launch_service, param_dump_command_action, proc_info, proc_output
+                launch_service, param_dump_command_action, proc_info, proc_output,
+                output_filter=rmw_implementation_filter
             ) as param_dump_command:
                 yield param_dump_command
         cls.launch_param_dump_command = launch_param_dump_command
--- a/ros2param/test/test_verb_dump.py
+++ b/ros2param/test/test_verb_dump.py
@@ -56,6 +56,13 @@ EXPECTED_PARAMETER_FILE = (
     '    use_sim_time: false\n'
 )
 
+# Skip cli tests on Windows while they exhibit pathological behavior
+# https://github.com/ros2/build_farmer/issues/248
+if sys.platform.startswith('win'):
+    pytest.skip(
+            'CLI tests can block for a pathological amount of time on Windows.',
+            allow_module_level=True)
+
 
 @pytest.mark.rostest
 @launch_testing.parametrize('rmw_implementation', get_available_rmw_implementations())
--- a/ros2param/test/test_verb_dump.py
+++ b/ros2param/test/test_verb_dump.py
@@ -233,9 +233,6 @@ class TestVerbDump(unittest.TestCase):
                 text=param_dump_command.output,
                 strict=True
             )
-
-            not_generated_param_file = os.path.join(tmpdir, self._output_file())
-            with self.assertRaises(OSError) as context:
-                open(not_generated_param_file, 'r')
             # Make sure the file was not create, thus '--print' did preempt
-            assert '[Errno 2] No such file or directory' in str(context.exception)
+            not_generated_param_file = os.path.join(tmpdir, self._output_file())
+            assert not os.path.exists(not_generated_param_file)
--- a/ros2param/test/test_verb_dump.py
+++ b/ros2param/test/test_verb_dump.py
@@ -159,7 +159,7 @@ class TestVerbDump(unittest.TestCase):
             self.fail(f'CLI daemon failed to find test node after {TEST_TIMEOUT} seconds')
 
     def _output_file(self):
-        return f'{TEST_NAMESPACE}/{TEST_NODE}'[1:].replace('/', '__') + '.yaml'
+        return f'{TEST_NAMESPACE}/{TEST_NODE}'.lstrip('/').replace('/', '__') + '.yaml'
 
     def test_verb_dump_invalid_node(self):
         with self.launch_param_dump_command(arguments=['invalid_node']) as param_dump_command:
--- a/ros2param/test/test_verb_dump.py
+++ b/ros2param/test/test_verb_dump.py
@@ -12,20 +12,27 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from io import StringIO
+import contextlib
 import os
+import sys
 import tempfile
-import threading
 import time
 import unittest
-from unittest.mock import patch
 import xmlrpc
 
-import rclpy
-from rclpy.executors import MultiThreadedExecutor
-from rclpy.parameter import PARAMETER_SEPARATOR_STRING
+from launch import LaunchDescription
+from launch.actions import ExecuteProcess
+from launch_ros.actions import Node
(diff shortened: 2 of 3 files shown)


ros2param/package.xml
```diff
@@ -20,6 +20,10 @@
   <test_depend>ament_flake8</test_depend>
   <test_depend>ament_pep257</test_depend>
   <test_depend>ament_xmllint</test_depend>
+  <test_depend>launch</test_depend>
+  <test_depend>launch_ros</test_depend>
+  <test_depend>launch_testing</test_depend>
+  <test_depend>launch_testing_ros</test_depend>
```
jacobperron, line 26: Even though it doesn't seem like we need to depend on `launch_testing_ros` (e.g. it is not explicitly used anywhere), if I remove this dependency, then I get the error:

```
TypeError: setUpClass() missing 4 required positional arguments: 'launch_service', 'proc_info', 'proc_output', and 'rmw_implementation'
```

@hidmic Any idea why this is the case?

jacobperron:
Testing with additional `ros2cli` since the failures this PR is trying to fix involves left-over daemons from other tests:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10127/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5694/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8229/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10021/)

jacobperron:
I can reproduce the Windows failure locally. It looks like that launch_testing is timing out waiting for the `ros2 param` commands to finish executing. Indeed on my machine `ros2 param dump` can take several seconds to complete.

jacobperron:
Rebuild on Windows with timeout increased (8bbe967): [Build Status](https://ci.ros2.org/job/ci_windows/10080/)

ros2param/test/test_verb_dump.py
```diff
@@ -109,52 +146,83 @@ def setUp(self):
...
+                assert param_dump_command.wait_for_shutdown(timeout=TEST_TIMEOUT)
+            assert param_dump_command.exit_code == launch_testing.asserts.EXIT_OK
+            assert launch_testing.tools.expect_output(
+                expected_text=EXPECTED_PARAMETER_FILE + '\n',
+                text=param_dump_command.output,
+                strict=True
+            )
 
             not_generated_param_file = os.path.join(tmpdir, self._output_file())
-
```
nuclearsandwich, line 155: Are these whitespace changes are intentional cleanup or artifacts?

ros2param/test/test_verb_dump.py
```diff
@@ -12,75 +12,112 @@
...
+    '      str_param: foo\n'
+    '    int_param: 42\n'
+    '    str_param: Hello World\n'
+    '    use_sim_time: false\n'
+)
+
+# TODO(jacobperron): Enable testing with all RMW implementations when they pass
+# @launch_testing.parametrize('rmw_implementation', get_available_rmw_implementations())
+@pytest.mark.rostest
+@launch_testing.parametrize('rmw_implementation', ['rmw_fastrtps_cpp'])
```
nuclearsandwich, line 60: What's the failure mode if Fast-RTPS is not part of the build? Does it make more sense to parameterize now and remove cyclonedds?

jacobperron: Good point. We can just remove the parameterization altogether for now.

ros2param/test/test_verb_dump.py
jacobperron: Intentional. I thought it looked better this way; grouping together the LOC related to checking that the directory does not exist. It's subjective, and I can revert if you prefer.

ros2param/test/test_verb_dump.py
jacobperron: To answer your question, I'd expect the test to fail because of the unknown RMW implementation.

jacobperron: Actually, with #482 resolved, it looks like the test should work with all of our existing RMWs. It was easier to parameterize now than not (56e34da)

jacobperron:
Anecdotally, I'll note that converting to launch tests adds significant overhead. In particular on my machine, the time it takes to run the tests has doubled, from ~1min to ~2min (with one RMW). I think this extra time is mainly coming from waiting for the daemon to restart (which I think is unavoidable for the moment).

jacobperron:
Rebased to get changes from #486

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10219/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5767/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8307/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10107/)

jacobperron:
:disappointed: Now I'm noticing test failures locally with fastrtps because of a warning introduced in https://github.com/ros2/rclpy/pull/536. I think it's going to cause a lot of other CLI tests to fail too (e.g. `ros2 topic echo` also produces the new warning).

ros2param/package.xml
jacobperron: With e166f03, we need the test dependency on `launch_testing_ros` anyways. But it still would
be good to know the answer to the original issue for posterity.

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10261/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5799/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8342/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10146/)

jacobperron:
Jenkins doesn't warn about it, but there was a deprecation warning that I've fixed in 8fef210
Rerunning Linux for sanity: [Build Status](https://ci.ros2.org/job/ci_linux/10285/)

ros2param/test/test_verb_dump.py
```diff
@@ -110,52 +159,83 @@ def setUp(self):
...
     def _output_file(self):
-
-        ns = self.node.get_namespace()
-        name = self.node.get_name()
-        if '/' == ns:
-            fqn = ns + name
-        else:
-            fqn = ns + '/' + name
-        return fqn[1:].replace('/', '__') + '.yaml'
+        return f'{TEST_NAMESPACE}/{TEST_NODE}'[1:].replace('/', '__') + '.yaml'
```
hidmic, line 162: @jacobperron would 
```suggestion
        return f'{TEST_NAMESPACE}/{TEST_NODE}'.lstrip('/').replace('/', '__') + '.yaml'
```
be clearer?

ros2param/test/test_verb_dump.py
```diff
@@ -110,52 +159,83 @@ def setUp(self):
...
+                strict=True
+            )
 
             not_generated_param_file = os.path.join(tmpdir, self._output_file())
-
             with self.assertRaises(OSError) as context:
                 open(not_generated_param_file, 'r')
-
             # Make sure the file was not create, thus '--print' did preempt
             assert '[Errno 2] No such file or directory' in str(context.exception)
```
hidmic, line 241: @jacobperron would `os.path.exists()` and `os.path.isfile()` serve the same purpose?

ros2param/test/test_verb_dump.py
jacobperron: Sure. This is left-over from the previous implementation.

ros2param/test/test_verb_dump.py
jacobperron: Yes, that sounds much better. This is just left over from the previous implementation, but I can change it if you like.

ros2param/test/test_verb_dump.py
```diff
@@ -12,75 +12,124 @@
...
+    '    str_param: Hello World\n'
+    '    use_sim_time: false\n'
+)
+
+# Skip cli tests on Windows while they exhibit pathological behavior
+# https://github.com/ros2/build_farmer/issues/248
+if sys.platform.startswith('win'):
+    pytest.skip(
+            'CLI tests can block for a pathological amount of time on Windows.',
+            allow_module_level=True)
```
nuclearsandwich, line 64: Glad/sad this is here. Thanks.

ros2param/test/test_verb_dump.py
nuclearsandwich: Adding to the above question, is the intent to test the behavior (that no file exists) or the error message (that the exception contains '...No such file...'. I see them as somewhat distinct and I think checking the behavior is less brittle than checking the error message.

ros2param/test/test_verb_dump.py
jacobperron: 4af4a47

ros2param/test/test_verb_dump.py
jacobperron: The former. It's not so useful to check the error message, IMO, since it is not a message created by the CLI tool itself. Changed in 43701e0

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10287/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5822/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8367/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10184/)

PR #489, opened 2020-04-15
Skip CLI tests on Windows until we resolve the blocking/hanging isuse.

nuclearsandwich:
We've been lacking on Windows test data for months due to the blocking
or hung state that these tests can get into. I think the best thing to
do is skip these tests on windows to allow other builds to complete
while we continue investigating the problem.

I've opted for skipping entire test modules as I think the underlying issue is in the test apparatus rather than any specific test.

### DIFF
--- a/ros2action/test/test_cli.py
+++ b/ros2action/test/test_cli.py
@@ -35,6 +35,14 @@ from rclpy.utilities import get_available_rmw_implementations
 import yaml
 
 
+# Skip cli tests on Windows while they exhibit pathological behavior
+# https://github.com/ros2/build_farmer/issues/248
+if sys.platform.startswith('win'):
+    pytest.skip(
+            'CLI tests can block for a pathological amount of time on Windows.',
+            allow_module_level=True)
+
+
 @pytest.mark.rostest
 @launch_testing.parametrize('rmw_implementation', get_available_rmw_implementations())
 def generate_test_description(rmw_implementation):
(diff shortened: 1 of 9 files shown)


nuclearsandwich:
CI up to these packages and testing them. On Windows I want to check that the CLI tests are skipped and on other platforms I want to verify that they run as expected.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10235/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5783/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8322/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10122/)

nuclearsandwich:
I'm pretty sure the test failures above were introduced by https://github.com/ros2/rclpy/pull/536 ([discussion](https://github.com/ros2/rclpy/pull/536#issuecomment-614338473)). I am going to merge in order to improve the situation on Windows.

dirk-thomas:
Please ticket the follow up task to revert this change.

nuclearsandwich:
> Please ticket the follow up task to revert this change.

Do you feel an issue separate from ros2/build_farmer#248 is warranted?

dirk-thomas:
I would be concerned that the task of reverting this change gets lost in the length and broader scope of that ticket.

nuclearsandwich:
> I would be concerned that the task of reverting this change gets lost in the length and broader scope of that ticket.

https://github.com/ros2/ros2cli/issues/490

PR #491, opened 2020-04-17
Implement times for ros2 topic pub

clalancette:

While doing some testing, it turned out to be useful to tell `ros2 topic pub` exactly how many times to publish a message.  This is basically a generic version of the `-1` flag that already exists, though that flag is left for backwards-compatibility.

I should point out that this is *not* the same as doing `for i in seq 1 10 ; do ros2 topic pub ... ; done`.  The difference is that with the bash loop, a new node and publisher is created every time.  This can be problematic for discovery, since it can happen that we bring up the node, publish, and destroy the node before connections are properly made.  Doing it inside of `ros2 topic pub` itself neatly avoids this issue.

The one potentially flaky part of this is the test, which expects *exactly* 5 messages to be published.  Depending on the platform and load, that may not be a reasonable assumption.  Thoughts on how to improve that, while still testing that we are publishing a fixed number of times, are appreciated.

### DIFF
--- a/ros2topic/ros2topic/verb/pub.py
+++ b/ros2topic/ros2topic/verb/pub.py
@@ -40,7 +40,7 @@ def nonnegative_int(inval):
     ret = int(inval)
     if ret < 0:
         # The error message here gets completely swallowed by argparse
-        raise ValueError('Value must be positive')
+        raise ValueError('Value must be positive or zero')
     return ret
 
 
--- a/ros2topic/ros2topic/verb/pub.py
+++ b/ros2topic/ros2topic/verb/pub.py
@@ -93,9 +93,6 @@ class PubVerb(VerbExtension):
             parser, is_publisher=True, default_preset='system_default')
 
     def main(self, *, args):
-        if args.once and args.times > 0:
-            raise RuntimeError('Cannot pass both -1 and -t <times>')
-
         return main(args)
 
 
--- a/ros2topic/ros2topic/verb/pub.py
+++ b/ros2topic/ros2topic/verb/pub.py
@@ -79,12 +79,18 @@ class PubVerb(VerbExtension):
         if args.rate <= 0:
             raise RuntimeError('rate must be greater than zero')
 
+        if args.once and args.times > 0:
+            raise RuntimeError('Cannot pass both -1 and -t <times>')
+
         return main(args)
 
 
 def main(args):
     qos_profile = qos_profile_from_short_keys(
         args.qos_profile, reliability=args.qos_reliability, durability=args.qos_durability)
+    times = args.times
+    if args.once:
+        times = 1
     with DirectNode(args, node_name=args.node_name) as node:
         return publisher(
             node.node,
@@ -93,8 +99,7 @@ def main(args):
             args.values,
             1. / args.rate,
             args.print,
-            args.once,
-            args.times,
+            times,
             qos_profile)
 
 
@@ -110,7 +115,6 @@ def publisher(
     values: dict,
     period: float,
     print_nth: int,
-    once: bool,
     times: int,
     qos_profile: QoSProfile,
 ) -> Optional[str]:
@@ -119,11 +123,6 @@ def publisher(
     values_dictionary = yaml.safe_load(values)
     if not isinstance(values_dictionary, dict):
         return 'The passed value needs to be a dictionary in YAML format'
-
-    if once:
-        if times > 0:
-            return 'Cannot pass both -1 and -t <times>'
-        times = 1
 
     publisher_callbacks = PublisherEventCallbacks(
         incompatible_qos=handle_incompatible_qos_event)
--- a/ros2topic/ros2topic/verb/pub.py
+++ b/ros2topic/ros2topic/verb/pub.py
@@ -36,6 +36,22 @@ import yaml
 MsgType = TypeVar('MsgType')
 
 
+def positive_int(inval):
+    ret = int(inval)
+    if ret < 0:
+        # The error message here gets completely swallowed by argparse
+        raise ValueError('Value must be positive')
+    return ret
+
+
+def positive_float(inval):
+    ret = float(inval)
+    if ret <= 0.0:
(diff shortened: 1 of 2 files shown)


mjcarroll:
> The one potentially flaky part of this is the test, which expects exactly 5 messages to be published. Depending on the platform and load, that may not be a reasonable assumption. Thoughts on how to improve that, while still testing that we are publishing a fixed number of times, are appreciated.

Would it be sufficient to check the console output like: 

```
publishing #1: std_msgs.msg.String(data='hello')
publishing #2: std_msgs.msg.String(data='hello')
publishing #3: std_msgs.msg.String(data='hello')
publishing #4: std_msgs.msg.String(data='hello')
publishing #5: std_msgs.msg.String(data='hello')
```

While that doesn't guarantee that the messages have been published, it would be a bit more robust.

clalancette:
> While that doesn't guarantee that the messages have been published, it would be a bit more robust.

Good idea!  I'll go with that instead.  Thanks.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10282/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5818/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8363/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10178/)

ros2topic/ros2topic/verb/pub.py
```diff
@@ -115,6 +120,11 @@ def publisher(
     if not isinstance(values_dictionary, dict):
         return 'The passed value needs to be a dictionary in YAML format'
 
+    if once:
+        if times > 0:
+            return 'Cannot pass both -1 and -t <times>'
```
dirk-thomas, line 125: This check should happen and `parse_args` as part of a custom type function.

clalancette: I ended up moving this check into main, alongside the other check for args.rate <= 0.

There is actually further work that could be done here to check for errors earlier.  For instance, we could look up the message type/module and fill out the message all within main, before passing it down to the `publish` function.  That would do less setup/teardown work in the case of an error.  But I'll leave that for a future cleanup.

dirk-thomas: I would still suggest to move both checks into custom type functions to determine valid input at argument parsing time.

ros2topic/ros2topic/verb/pub.py
```diff
@@ -69,19 +69,28 @@ def add_arguments(self, parser, cli_name):
...
+            '-t', '--times', type=int, default=0,
+            help='Publish this number of times and then exit')
         add_qos_arguments_to_argument_parser(
             parser, is_publisher=True, default_preset='system_default')
 
     def main(self, *, args):
         if args.rate <= 0:
             raise RuntimeError('rate must be greater than zero')
 
+        if args.once and args.times > 0:
```
fujitatomoya, line 82:
```suggestion
        if args.once and args.times > 1:
```

i think that with once and times == 1, we could proceed.

clalancette: While that is true, and would be logically consistent, in general passing both `-1` and `-t` isn't logically consistent.  Thus, I would rather always throw an error here saying "you can't pass both of these at the same time".

ros2topic/ros2topic/verb/pub.py
clalancette: As far as I can tell, this isn't possible.  You can define a custom [type callable](https://docs.python.org/3/library/argparse.html#type) in argparse, but that callable does not have access to the overall `args` structure.  Even if it did, there is no way of knowing that you parsed all of the arguments by the time you got there.  You can really only do this particular check at the end.

That being said, I did implement a custom type check for both rate and times that ensures that that they are positive float and non-negative integer, respectively.

clalancette:
New CI:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10315/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5844/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8400/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10220/)

ros2topic/ros2topic/verb/pub.py
```diff
@@ -36,6 +36,22 @@
 MsgType = TypeVar('MsgType')
 
 
+def positive_int(inval):
+    ret = int(inval)
+    if ret < 0:
+        # The error message here gets completely swallowed by argparse
+        raise ValueError('Value must be positive')
+    return ret
```
dirk-thomas, line 44: Both functions use the term `positive` but handle differently. That seems pretty confusing to me. Positive integers do not include zero. So I would suggest to rename this function to `non_negative_int` and update the error message accordingly.

ros2topic/ros2topic/verb/pub.py
dirk-thomas: I was referring to both individual checks which have been moved into type functions :+1:

If both arguments can't be passed at the same time the arguments should be added to an exclusive group.

clalancette: Good call about the function names; fixed now.  Also, I didn't know about argparse's mutually exclusive groups, thanks for that.  I've now changed it to use them.  I'll run another CI here, and this is ready for another review.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10376/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5891/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8456/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10273/)

clalancette:
The macOS failure is a flake in topic_hz test, unrelated to this PR.  So I think this is ready to go now.

Issue #488, opened 2020-04-15
ros2 topic echo should use the daemon to look up type names


clalancette:
## Bug report

When trying to run `ros2 topic echo /foo` for a topic that already exists, sometimes it prints:

```
Could not determine the type for the passed topic
```

This happens because `ros2 topic echo` creates a node, then has to wait for discovery to happen.  On some DDS implementations where discovery is slower, it can timeout waiting for the discovery.  What we should do instead is utilize the ros2 daemon and ask it for the type of the topic.  If the daemon knows the type of the topic, it can return it immediately.  If it isn't running, then we should probably start the daemon.  And if it is running, but doesn't know the type of the topic, echo can then wait to discover it.

**Required Info:**

- Operating System:
  - Any
- Installation type:
  - From source
- Version or commit hash:
  - bf4941afd06064d20914c19687a8ca98186687e6
- DDS implementation:
  - Any, though it happens more on RTI Connext which has slower discovery
- Client library (if applicable):
  - rclpy

#### Steps to reproduce issue
Terminal 1:
```
ros2 run demo_nodes_cpp talker
```

Terminal 2 (this has to be done very quickly after Terminal 1 to show the issue):
```
ros2 topic echo /chatter
```

#### Expected behavior

The topic type is fetched (invisibly), and `topic echo` starts echoing the topic.

#### Actual behavior

Sometimes an error is shown:

```
Could not determine the type for the passed topic
```

hidmic:
This has been addressed by #494.

PR #494, opened 2020-04-22
Make CLI more robust to discovery latency.

hidmic:
Split from #421. This pull request builds on top of #493 and ensures the daemon gets used throughout all CLI tools. It also updates some tests which showed to be flakey nonetheless (with and without this patch).

### DIFF
--- a/ros2action/test/test_cli.py
+++ b/ros2action/test/test_cli.py
@@ -72,7 +72,7 @@ def generate_test_description(rmw_implementation):
     ])
 
 
-def get_fibonacci_send_goal_output(*, order=1, with_feedback=False):
+def get_fibonacci_send_goal_output(*, order=1, with_feedback=0):
     assert order > 0
     output = [
         'Waiting for an action server to become available...',
@@ -85,12 +85,13 @@ def get_fibonacci_send_goal_output(*, order=1, with_feedback=False):
     sequence = [0, 1]
     for _ in range(order - 1):
         sequence.append(sequence[-1] + sequence[-2])
-        if with_feedback:
+        if with_feedback > 0:
             output.append('Feedback:')
             output.extend(('    ' + yaml.dump({
                 'sequence': sequence
             })).splitlines())
             output.append('')
+            with_feedback -= 1
     output.append('Result:'),
     output.extend(('    ' + yaml.dump({
         'sequence': sequence
@@ -251,8 +252,6 @@ class TestROS2ActionCLI(unittest.TestCase):
             assert action_command.wait_for_shutdown(timeout=10)
         assert action_command.exit_code == launch_testing.asserts.EXIT_OK
         assert launch_testing.tools.expect_output(
-            expected_lines=get_fibonacci_send_goal_output(
-                order=5, with_feedback=True
-            ),
-            text=action_command.output, strict=True
+            expected_lines=get_fibonacci_send_goal_output(order=5, with_feedback=1),
+            text=action_command.output, strict=False
         )
--- a/ros2component/ros2component/api/__init__.py
+++ b/ros2component/ros2component/api/__init__.py
@@ -24,7 +24,6 @@ import rcl_interfaces.msg
 
 import rclpy
 
-from ros2cli.node.direct import DirectNode
 from ros2cli.node.strategy import NodeStrategy
 from ros2node.api import get_node_names
 from ros2node.api import get_service_server_info
@@ -82,7 +81,10 @@ def get_container_components_info(*, node, remote_container_node_name):
         '{}/_container/list_nodes'.format(remote_container_node_name)
     )
     try:
-        list_nodes_client.wait_for_service()
+        if not list_nodes_client.wait_for_service(timeout_sec=5.0):
+            raise RuntimeError(
+                f"No list_nodes service found for '{remote_container_node_name}' container"
+            )
         future = list_nodes_client.call_async(
             composition_interfaces.srv.ListNodes.Request()
         )
@@ -128,7 +130,10 @@ def load_component_into_container(
         '{}/_container/load_node'.format(remote_container_node_name)
     )
     try:
-        load_node_client.wait_for_service()
+        if not load_node_client.wait_for_service(timeout_sec=5.0):
+            raise RuntimeError(
+                f"No load_node service found for '{remote_container_node_name}' container"
+            )
         request = composition_interfaces.srv.LoadNode.Request()
         request.package_name = package_name
         request.plugin_name = plugin_name
@@ -177,7 +182,10 @@ def unload_component_from_container(*, node, remote_container_node_name, compone
         '{}/_container/unload_node'.format(remote_container_node_name)
     )
     try:
-        unload_node_client.wait_for_service()
+        if not unload_node_client.wait_for_service(timeout_sec=5.0):
+            raise RuntimeError(
+                f"No unload_node service found for '{remote_container_node_name}' container"
+            )
         for uid in component_uids:
             request = composition_interfaces.srv.UnloadNode.Request()
             request.unique_id = uid
@@ -193,7 +201,7 @@ def find_container_node_names(*, node, node_names):
     """
     Identify container nodes from a list of node names.
 
-    :param node: a `ros2cli.node.DirectNode` instance
+    :param node: a node-like instance
     :param node_names: list of `ros2node.api.NodeName` instances, as returned
         by `ros2node.api.get_node_names()`
     :return: list of `ros2node.api.NodeName` instances for nodes that are
@@ -230,11 +238,9 @@ def package_with_components_name_completer(prefix, parsed_args, **kwargs):
 def container_node_name_completer(prefix, parsed_args, **kwargs):
     """Callable returning a list of container node names."""
     with NodeStrategy(parsed_args) as node:
-        node_names = get_node_names(node=node)
-    with DirectNode(parsed_args) as node:
         return [
             n.full_name for n in find_container_node_names(
-                node=node, node_names=node_names
+                node=node, node_names=get_node_names(node=node)
             )
         ]
 
--- a/ros2component/ros2component/api/__init__.py
+++ b/ros2component/ros2component/api/__init__.py
@@ -83,7 +83,7 @@ def get_container_components_info(*, node, remote_container_node_name):
     try:
         if not list_nodes_client.wait_for_service(timeout_sec=5.0):
             raise RuntimeError(
-                f"No list_nodes service found for '{remote_container_node_name}' container"
+                f"No 'list_nodes' service found for '{remote_container_node_name}' container"
             )
         future = list_nodes_client.call_async(
             composition_interfaces.srv.ListNodes.Request()
@@ -132,7 +132,7 @@ def load_component_into_container(
     try:
         if not load_node_client.wait_for_service(timeout_sec=5.0):
             raise RuntimeError(
-                f"No load_node service found for '{remote_container_node_name}' container"
+                f"No 'load_node' service found for '{remote_container_node_name}' container"
             )
         request = composition_interfaces.srv.LoadNode.Request()
         request.package_name = package_name
@@ -184,7 +184,7 @@ def unload_component_from_container(*, node, remote_container_node_name, compone
     try:
         if not unload_node_client.wait_for_service(timeout_sec=5.0):
             raise RuntimeError(
-                f"No unload_node service found for '{remote_container_node_name}' container"
+                f"No 'unload_node' service found for '{remote_container_node_name}' container"
             )
         for uid in component_uids:
             request = composition_interfaces.srv.UnloadNode.Request()
--- a/ros2component/ros2component/verb/standalone.py
+++ b/ros2component/ros2component/verb/standalone.py
@@ -16,8 +16,8 @@ import platform
 import signal
 import uuid
 
+from ros2cli.node.direct import add_arguments
 from ros2cli.node.direct import DirectNode
-from ros2cli.node.strategy import add_arguments
 
 from ros2component.api import add_component_arguments
 from ros2component.api import load_component_into_container
--- a/ros2component/ros2component/verb/standalone.py
+++ b/ros2component/ros2component/verb/standalone.py
@@ -16,8 +16,8 @@ import platform
 import signal
 import uuid
 
-from ros2cli.node.direct import add_arguments
 from ros2cli.node.direct import DirectNode
+from ros2cli.node.strategy import add_arguments
 
 from ros2component.api import add_component_arguments
 from ros2component.api import load_component_into_container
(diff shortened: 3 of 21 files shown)


hidmic:
With https://github.com/ros2/ros2cli/pull/494/commits/fac0a491aa2bd10c85ed295141f7d39fe20bb4bd, https://github.com/ros2/ros2cli/issues/488 should be now addressed.

hidmic:
CI up to `ros2topic`, `ros2service`, `ros2action`, `ros2node` and `ros2component`:

- Linux [Build Status](https://ci.ros2.org/job/ci_linux/10531/)
- Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/5989/)
- OSX [Build Status](https://ci.ros2.org/job/ci_osx/8578/)
- Windows [Build Status](https://ci.ros2.org/job/ci_windows/10398/)

ros2component/ros2component/api/__init__.py
```diff
@@ -82,7 +81,10 @@ def get_container_components_info(*, node, remote_container_node_name):
         '{}/_container/list_nodes'.format(remote_container_node_name)
     )
     try:
-        list_nodes_client.wait_for_service()
+        if not list_nodes_client.wait_for_service(timeout_sec=5.0):
```
dirk-thomas, line 84: When used in a loop (as in `component list`) a blocking timeout of 5s can add up to a very long time for the user call.

ros2topic/ros2topic/verb/echo.py
```diff
@@ -95,16 +97,20 @@ def main(args):
         callback = subscriber_cb_csv(truncate_length, args.no_arr, args.no_str)
     qos_profile = qos_profile_from_short_keys(
         args.qos_profile, reliability=args.qos_reliability, durability=args.qos_durability)
+    if args.message_type is None:
+        with NodeStrategy(args) as node:
```
dirk-thomas, line 101: This potentially creates a second direct node (beside the one created a few lines below) which doesn't seem like a good idea.

Same pattern below in multiple cases (even though they preexisted before).

dirk-thomas:
> With fac0a49, #488 should be now addressed.

Please use a format which will ensure the referenced ticket is automatically closed when the PR is merged.

ros2action/test/test_cli.py
```diff
@@ -251,8 +252,6 @@ def test_send_fibonacci_goal_with_feedback(self):
             assert action_command.wait_for_shutdown(timeout=10)
         assert action_command.exit_code == launch_testing.asserts.EXIT_OK
         assert launch_testing.tools.expect_output(
-            expected_lines=get_fibonacci_send_goal_output(
-                order=5, with_feedback=True
-            ),
-            text=action_command.output, strict=True
+            expected_lines=get_fibonacci_send_goal_output(order=5, with_feedback=1),
+            text=action_command.output, strict=False
```
dirk-thomas, line 256: Why not `strict` anymore?

hidmic: TL; DR because we cannot guarantee all action feedback will be received.

This bit is covered by https://github.com/ros2/ros2cli/pull/494/commits/b82bcd697c9bc5359694a558dcbd4475051d6e41. Test was flaky because it expected all published goal feedback to be printed. But the `ros2 action send_goal` exits once the goal result arrives, but actions' design does not (it cannot) guarantee all feedback will have arrived by the time the result response arrives. So I had to change the test to only expect the first piece of expected feedback and then look for the result -- which still does not provide any guarantees but it's much less likely to fail.

Having said that, we could try to match as many feedback messages as possible. But `expect_output()` can't do that now.

ros2component/ros2component/api/__init__.py
hidmic: Fair enough. Timeout is a somewhat arbitrary, larger than zero quantity. What about `2.0` seconds?

ros2topic/ros2topic/verb/echo.py
hidmic: That is true, you might end up creating node, destroying it and creating another one. We do want to use the daemon if available though.

A solution could be to extend `NodeStrategy`'s API to ensure you're dealing with a direct node in case you have to. Something like a `force_direct()` method (or a context manager?). WDYT?

dirk-thomas: > A solution could be to extend NodeStrategy's API to ensure you're dealing with a direct node in case you have to.

What is the difference to a `DirectNode` then?

ros2component/ros2component/api/__init__.py
dirk-thomas: I don't mind the value. For this PR you can choose whatever.

The problem is that with the current API all waits happen sequentially. E.g. if the service of `N` nodes isn't available the CLI will hang for `N * timeout`. I would prefer that the CLI only blocks for `timeout` seconds (whatever that value is). But that is pretty much our of scope for this PR.

ros2topic/ros2topic/verb/echo.py
dirk-thomas: I think this should create direct node, and conditionally if a daemon is running query some information from that to not have to wait for "complete" discovery for the direct node.

hidmic: > What is the difference to a `DirectNode` then?

That it'd use the daemon outside the scope where you actually need it to be direct, e.g.:

```python
    # ...
    with node.force_direct():
        node.subscribe()
```

> I think this should create direct node, and conditionally if a daemon is running query some information from that to not have to wait for "complete" discovery for the direct node.

That sounds a lot like replicating `NodeStrategy`'s implementation N times. I'd rather improve it in a single place.

hidmic: See https://github.com/ros2/ros2cli/pull/499 for one alternative.

dirk-thomas: > See #499 for an alternative.

This looks much better :+1:

hidmic:
CI up to `ros2topic`, `ros2service`, `ros2action`, `ros2node` and `ros2component`:

- Linux [Build Status](https://ci.ros2.org/job/ci_linux/10697/)

ros2component/ros2component/api/__init__.py
dirk-thomas: Please create a follow up ticket for this.

ros2component/ros2component/verb/standalone.py
```diff
@@ -16,8 +16,8 @@
 import signal
 import uuid
 
+from ros2cli.node.direct import add_arguments
 from ros2cli.node.direct import DirectNode
-from ros2cli.node.strategy import add_arguments
```
dirk-thomas, line 20: This change doesn't seem to be related to the topic of the PR. Please consider separating it.

ros2component/ros2component/verb/types.py
```diff
@@ -25,7 +24,6 @@ class TypesVerb(VerbExtension):
     """Output a list of components registered in the ament index."""
 
     def add_arguments(self, parser, cli_name):
-        add_arguments(parser)
```
dirk-thomas, line 28: This change doesn't seem to be related to the topic of the PR. Please consider separating it.

ros2service/ros2service/verb/find.py
```diff
@@ -34,6 +35,7 @@ def add_arguments(self, parser, cli_name):
         parser.add_argument(
             '--include-hidden-services', action='store_true',
             help='Consider hidden services as well')
+        add_strategy_node_arguments(parser)
```
dirk-thomas, line 38: This change doesn't seem to be related to the topic of the PR. Please consider separating it.

ros2service/ros2service/verb/list.py
```diff
@@ -22,7 +22,7 @@ class ListVerb(VerbExtension):
     """Output a list of available services."""
 
     def add_arguments(self, parser, cli_name):
-        add_arguments(parser)
+        add_strategy_node_arguments(parser)
```
dirk-thomas, line 25: This change doesn't seem to be related to the topic of the PR. Please consider separating it.

ros2service/ros2service/verb/type.py
```diff
@@ -27,6 +28,7 @@ def add_arguments(self, parser, cli_name):
             help="Name of the ROS service to get type (e.g. '/talker/list_parameters')")
         arg.completer = ServiceNameCompleter(
             include_hidden_services_key='include_hidden_services')
+        add_strategy_node_arguments(parser)
```
dirk-thomas, line 31: This change doesn't seem to be related to the topic of the PR. Please consider separating it.

ros2topic/ros2topic/verb/find.py
```diff
@@ -33,6 +35,7 @@ def add_arguments(self, parser, cli_name):
         parser.add_argument(
             '--include-hidden-topics', action='store_true',
             help='Consider hidden topics as well')
+        add_strategy_node_arguments(parser)
```
dirk-thomas, line 38: This change doesn't seem to be related to the topic of the PR. Please consider separating it.

ros2topic/ros2topic/verb/list.py
```diff
@@ -22,7 +23,7 @@ class ListVerb(VerbExtension):
     """Output a list of available topics."""
 
     def add_arguments(self, parser, cli_name):
-        add_arguments(parser)
+        add_strategy_node_arguments(parser)
```
dirk-thomas, line 26: This change doesn't seem to be related to the topic of the PR. Please consider separating it.

ros2component/ros2component/api/__init__.py
hidmic: Oh, good point. See https://github.com/ros2/ros2cli/issues/507.

ros2component/ros2component/verb/standalone.py
hidmic: Unrelated, yes. Reverted in https://github.com/ros2/ros2cli/pull/494/commits/69363670e4c14ede0a1cd48e80cd4b4dce4fed2f. Same for all other cases below.

ros2topic/ros2topic/verb/list.py
jacobperron: I think this comment still applies.

hidmic: Argh, I thought I had changed them all. See https://github.com/ros2/ros2cli/pull/494/commits/ba6248b11201436e21e9cccee83ac291e29698c2.

ros2component/ros2component/api/__init__.py
```diff
@@ -82,7 +81,10 @@ def get_container_components_info(*, node, remote_container_node_name):
...
+            raise RuntimeError(
+                f"No list_nodes service found for '{remote_container_node_name}' container"
```
dirk-thomas, line 86: The service name should be quoted for better readability.

Same below.

ros2topic/ros2topic/verb/list.py
```diff
@@ -22,7 +22,6 @@ class ListVerb(VerbExtension):
     """Output a list of available topics."""
 
     def add_arguments(self, parser, cli_name):
-        add_arguments(parser)
```
dirk-thomas, line 25: This change doesn't seem to be related to the topic of the PR. Please consider separating it.

dawonn-haval:
Just wanted to chime in and mention that this patch resolves the cli instability for me across multiple hosts that I've been seeing for a long time. 

I wonder if the same changes need to be made to the  ros2bag package?

ros2component/ros2component/api/__init__.py
hidmic: See 5a28d9da97d673a795a3e911cd6e843765399065.

ros2topic/ros2topic/verb/list.py
hidmic: See 5a28d9da97d673a795a3e911cd6e843765399065.

hidmic:
Running CI once again, got linter test failures due to recent `flake8` updates.

clalancette:
I might also suggest a full CI run on Linux, just to see what the system_tests stuff looks like with this in place.

dirk-thomas:
> I might also suggest a full CI run on Linux, just to see what the system_tests stuff looks like with this in place.

Which part of `ros2cli` do the `system_tests` use?

clalancette:
> Which part of `ros2cli` do the `system_tests` use?

I'm honestly not sure, but this is a big enough change that it makes me wary of the overnights.  I'd prefer to see what it does on a full run, as I'd like not to regress at this point (we are actually getting pretty close to green for Linux CI, for instance).

dirk-thomas:
> I'd prefer to see what it does on a full run,

A more complete run always makes sense. But also only for downstream packages which actually depend on something in the `ros2cli` repo. And in `system_tests` the only package depending on `ros2cli` is [test_security](https://github.com/ros2/system_tests/blob/2f157db5cf47abaf8f3ef0b163c78c57a6dba14f/test_security/package.xml#L29) - so running CI on e.g. `test_communication` would be not helpful and a waste of time / resources.

hidmic:
CI up to and above `ros2cli`, `ros2topic`, `ros2service`, `ros2action`, `ros2node` and `ros2component`:

- Linux [Build Status](https://ci.ros2.org/job/ci_linux/10729)

hidmic:
@clalancette @dirk-thomas all test failures here are unrelated to this patch. I've sent separate PRs to fix flake8 errors and `sros2` test failures. Do you want to move forward or do you want to wait for those to land and make another CI run, this time across platforms?

hidmic:
> It would be good to investigate why the test failure happened in the PR build.

It's a strange one, I cannot reproduce locally.

ros2topic/ros2topic/verb/echo.py
```diff
@@ -105,33 +107,8 @@ def subscriber(
...
-            if n == expanded_name:
-                if len(t) > 1:
-                    raise RuntimeError(
-                        "Cannot echo topic '%s', as it contains more than one type: [%s]" %
-                        (topic_name, ', '.join(t))
-                    )
-                message_type = t[0]
-                break
-        else:
-            raise RuntimeError(
```
jacobperron, line 128: I think we need to add back a similar guard to avoid passing `None` to `rclpy.Node.create_subscription`.
This introduced a bug reported in https://github.com/ros2/ros2cli/issues/519

jacobperron: See fix in https://github.com/ros2/ros2cli/pull/520

aitazhixin:
In dashing 0.7.11 version, 'Could not determine the type for the passed topic' also happens.

hidmic:
@aitazhixin I don't think these changes were backported to Dashing. It doesn't seem like it'd hurt either, as it only extends API. @nuclearsandwich thoughts?

aitazhixin:
> @aitazhixin I don't think these changes were backported to Dashing. This patch (and connected ones) only extends and improves API, but I will say that anything bigger than a bugfix is less likely to be considered for a backport. @nuclearsandwich thoughts?

image
so, I need a version which is bigger than 0.9.3?

aitazhixin:
@hidmic so, I need a version which is bigger than 0.9.3?
image

Issue #492, opened 2020-04-20
ros2topic test_echo_pub can't handle new QoS mismatch warnings


eboasson:
**Required Info:**

- Operating System:
  - macOS 10.15.4 (should be platform independent)
- Installation type:
  - from source
- Version or commit hash:
  - 75e0ea1f5da57321dd3d6e78e842a4f8fe68a3d6 (master as of 19-4-2020)
- DDS implementation:
  - rmw_cyclonedds_cpp

A bit of experimentation shows that it is the new warning when incompatible QoS's are detected that throws off the test. They depend on a feature supported by Cyclone DDS but not yet by Fast-RTPS.

#### Steps to reproduce issue
```
(cd src/ros2/ros2cli/ros2topic/test && pytest test_echo_pub.py)
```

#### Expected behavior

Test passes

#### Actual behavior

Test fails with

```
======================================================================
FAIL: test_echo_basic (test_echo_pub.TestROS2TopicEchoPub) (topic='/clitest/topic/echo_incompatible_qos', provide_qos=True, compatible_qos=False)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/erik/ros2_ws/src/ros2/ros2cli/ros2topic/test/test_echo_pub.py", line 250, in test_echo_basic
    assert not command.output, (
AssertionError: Echo CLI should not have received anything with incompatible QoS
assert not 'Incompatible QoS Policy detected: RELIABILITY_QOS_POLICY\n'
 +  where 'Incompatible QoS Policy detected: RELIABILITY_QOS_POLICY\n' = <launch_testing.tools.process.ProcessProxy object at 0x1090be7d0>.output
```

mm318:
> * DDS implementation:
>   * rmw_cyclonedds_cpp

> They depend on a feature supported by Cyclone DDS but not yet by Fast-RTPS.

So is this a problem with `rmw_cyclonedds_cpp` or `rmw_fastrtps_cpp`?

I wasn't able to reproduce the test failure with just:
> `cd src/ros2/ros2cli/ros2topic/test && pytest test_echo_pub.py`

Is there some environment variable that needs to be set?

dirk-thomas:
> So is this a problem with `rmw_cyclonedds_cpp` or `rmw_fastrtps_cpp`?

As the above description clearly mentions:

    DDS implementation:
        rmw_cyclonedds_cpp

> Is there some environment variable that needs to be set?

To choose a non-default RMW you have to set e.g.:

[...]

PR #496, opened 2020-04-23
Filter out "Incompatible QoS" messages from unit test

mm318:
Closes https://github.com/ros2/ros2cli/issues/492

The `test_echo_pub.py` unit test is failing due to interpreting newly added warning messages as received messages from a publisher. This pull request filters out those warning messages.


dirk-thomas:
Should the existence of such a warning really be ignored in the test? Should the test not either ensure that no such warning is present or that an excepted warning is present?

mm318:
That test's purpose is to test for the presence or non-presence of topic messages received from a publisher.

There could be another test that tests for the presence or non-presence of a warning, but that would be more reasonable if the default `rmw` implementation FastRTPS also supports it.

hidmic:
> That test's purpose is to test for the presence or non-presence of topic messages received from a publisher.

Regardless, why does that warning pop up in the first place?

mm318:
> Regardless, why does that warning pop up in the first place?

The test is purposely using a publisher with a QoS policy that's incompatible with `ros2 topic echo`'s subscription's QoS policy.

dirk-thomas:
> The test is purposely using a publisher with a QoS policy that's incompatible with the custom set `ros2 topic echo` subscription QoS policy.

Then the test should also cover that the user see the desired warning.

mm318:
> Then the test should also cover that the user see the desired warning.

Okay @dirk-thomas, I have updated the test to do so.

ros2topic/test/test_echo_pub.py
```diff
@@ -247,9 +247,15 @@ def publish_message():
                             'Echo CLI did not print expected message'
                         )
                     else:
-                        assert not command.output, (
-                            'Echo CLI should not have received anything with incompatible QoS'
-                        )
+                        if 'rmw_fastrtps' in get_rmw_implementation_identifier():
```
hidmic, line 250: @mm318 mind to add a TODO for removal once https://github.com/ros2/rmw_fastrtps/issues/356 is addressed?

mm318: Done.

hidmic:
@mm318 there're some flake8 issues.

mm318:
Hi @hidmic, I have resolved the `flake8` issues. Thanks!

mm318:
Hi @hidmic, I think this pull request is ready to be merged. Thanks!

mm318:
Let's hold off on merging this pull request until we've reached a conclusion on https://github.com/ros2/ros2cli/issues/500#issuecomment-621348241, because if we were to revert the changes in https://github.com/ros2/ros2cli/pull/410, then the contents of the expected warning message will be different.

ivanpauno:
CI for both this PR and https://github.com/ros2/rclpy/pull/553:

- Linux [Build Status](https://ci.ros2.org/job/ci_linux/10603/)
- Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6044/)
- macOS [Build Status](http://ci.ros2.org/job/ci_osx/8634/)
- Windows [Build Status](http://ci.ros2.org/job/ci_windows/10499/)

ivanpauno:
@hidmic is this ready to be merged?

hidmic:
It is, assuming the discussion in https://github.com/ros2/ros2cli/issues/500#issuecomment-621348241 
 has settled. Are you planning on making more changes @mm318?

ivanpauno:
> Are you planning on making more changes @mm318?

#553 is enough to solve that discussion.
Reverting #410 is optional, and would modify the messages.

IMO, this PR is ready to be merged. @mm318 WDYT?

mm318:
I don't plan on making anymore changes to this pull request. It should be merged to also fix https://github.com/ros2/ros2cli/issues/492. Thanks!

PR #499, opened 2020-04-28
Improve NodeStrategy to use the right node seamlessly.

hidmic:


ros2cli/ros2cli/node/strategy.py
```diff
@@ -25,21 +25,47 @@ class NodeStrategy:
     def __init__(self, args):
         use_daemon = not getattr(args, 'no_daemon', False)
         if use_daemon and is_daemon_running(args):
-            self.node = DaemonNode(args)
+            self._daemon_node = DaemonNode(args)
```
dirk-thomas, line 28: This looks like a race between `is_daemon_running` and creating the `DaemonNode`. You can end up with a not-working daemon node (and access of its `None` proxy would raise) but no fallback.

hidmic: > This looks like a race between is_daemon_running and creating the DaemonNode

There is a race, but I don't see a way around it.

> You can end up with a not-working daemon node (and access of its None proxy would raise) but no fallback.

That is true for the current [`NodeStrategy` implementation](https://github.com/ros2/ros2cli/blob/b5b53263410a594d6211cc2a7b6ee827f966b1aa/ros2cli/ros2cli/node/strategy.py#L28). With this patch, [this condition](https://github.com/ros2/ros2cli/pull/499/files#diff-8a54fb07e6acc0ff6c8e4d8c731dca06R59) would evaluate to `False` in that case and a `DirectNode` would get instantiated instead.

dirk-thomas: But that fallback is opaque to the caller and the function might return incomplete information since the direct node didn't have any time to discover, no?

hidmic:
> Atm the StrategyNode allows you to treat it as a node and behind the scene you get the speed of a daemon or a direct node (with potential discovery delays). For the referenced issue I think we want a node which uses daemon methods when available and otherwise use the methods on a direct node. Maybe that kind of functionality shouldn't go into StrategyNode but into a new class since it is a pretty different usage / expected API?

Such a new class would provide the exact same functionality that `NodeStrategy` provides for all existing use cases, plus 

> a node which uses daemon methods when available and otherwise use the methods on a direct node.

Considering that this patch is completely transparent for all uses cases that rely on the current `NodeStrategy` behavior, I don't see the point in creating yet another class. I can imagine `NodeStrategy` being progressively replaced by it over time.

ros2cli/ros2cli/node/strategy.py
hidmic: Right. But that's expected. Same goes if the daemon wasn't running at the moment and had to be spawned, in which case there's no real benefit in using the daemon (which isn't a silver bullet to cope with discovery latency anyways, we simply _hope_ the daemon and the rest of the ROS ecosystem have been around long enough for that not to be a problem).

dirk-thomas:
> Is the goal here to address #488?

Can you still answer this question?

hidmic:
> Is the goal here to address #488?

This PR sets up the scene. I'll update https://github.com/ros2/ros2cli/pull/494 to make use of these features, which will ultimately address #488.

hidmic:
CI up to `ros2cli`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10695/) (canary build)

hidmic:
Alright, this one's ready to go. How you feel about it @jacobperron ? It doesn't change API, but it affects most `ros2` CLI verbs.

jacobperron:
IIUC, we're affecting most CLI verbs for the better, so I'm okay with this (and the connected PR) landing for the Foxy release candidate.

hidmic:
Alright, going in!

PR #501, opened 2020-05-01
add support for get_node_names_and_namespaces_with_enclaves

mikaelarguedas:
to enable https://github.com/ros2/sros2/pull/203


ivanpauno:
@jacobperron with the same rationale that in https://github.com/ros2/sros2/pull/203#pullrequestreview-404135526, I think we can get this in.
WDYT?

ivanpauno:
Based on comment here https://github.com/ros2/sros2/pull/203#issuecomment-622457750 and CI results here https://github.com/ros2/sros2/pull/203#issuecomment-622325401, I will merge this one.

Thanks @mikaelarguedas !

Issue #503, opened 2020-05-04
ros2cli daemon test hanging with Connext


jacobperron:
## Bug report

The Foxy CI build with Connext-only is timing out (5 hours) at what looks like `test_daemon.py`.

Example build: http://build.ros2.org/view/Fci/job/Fci__nightly-connext_ubuntu_focal_amd64/98

Note, it appears to be a flake since there has been a successful build in between timeouts:

http://build.ros2.org/view/Fci/job/Fci__nightly-connext_ubuntu_focal_amd64/97/

dirk-thomas:
The first build timeout happened on Apr 18th and since then failed 5 of 7 times. That makes PR ros2/ros2cli#493 merged the day before look suspicious.

dirk-thomas:
Please see an example build where the `test_daemon` tests fails and is being repeated: https://ci.ros2.org/job/ci_linux/10624/consoleFull#console-section-204

```
_____________________ test_get_subscriptions_info_by_topic _____________________
test/test_daemon.py:204: in test_get_subscriptions_info_by_topic
    subscriptions_info = daemon_node.get_subscriptions_info_by_topic(TEST_TOPIC_NAME)
/usr/lib/python3.8/xmlrpc/client.py:1109: in __call__
    return self.__send(self.__name, args)
/usr/lib/python3.8/xmlrpc/client.py:1450: in __request
    response = self.__transport.request(
/usr/lib/python3.8/xmlrpc/client.py:1153: in request
    return self.single_request(host, handler, request_body, verbose)
/usr/lib/python3.8/xmlrpc/client.py:1169: in single_request
    return self.parse_response(resp)
/usr/lib/python3.8/xmlrpc/client.py:1341: in parse_response
    return u.close()
/usr/lib/python3.8/xmlrpc/client.py:655: in close
    raise Fault(**self._stack[0])
E   xmlrpc.client.Fault: <Fault 1: "<class 'OverflowError'>:Total nanoseconds value is too large to store in C time point.">
```

hidmic:
Alright, https://github.com/ros2/rclpy/pull/555 and 504 fixes test failures. Curious thing is that no hangs occur if the test doesn't fail. I'll try to figure out why.

PR #504, opened 2020-05-05
Make test_daemon.py robust to discovery latency.

hidmic:
Fixes https://github.com/ros2/ros2cli/issues/503. Some RMW implementations are slow enough to make this graph API-intensive test flakey. This pull requests attempts to circumvent that problem.

### DIFF
--- a/ros2cli/test/test_daemon.py
+++ b/ros2cli/test/test_daemon.py
@@ -110,6 +110,33 @@ def daemon_node():
     assert spawn_daemon(args=[], wait_until_spawned=5.0)
     with DaemonNode(args=[]) as node:
         yield node
+        node.system.shutdown()
+
+
+def retry(*, times=1, delay=0):
+    import functools
+    import inspect
+    import time
+
+    assert times > 0
+    assert delay >= 0
+
+    def decorator(func):
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            nonlocal times
+            while True:
+                times -= 1
+                try:
+                    return func(*args, **kwargs)
+                except AssertionError:
+                    if times == 0:
+                        raise
+                    if delay > 0:
+                        time.sleep(delay)
+        wrapper.__signature__ = inspect.signature(func)
+        return wrapper
+    return decorator
 
 
 def test_get_name(daemon_node):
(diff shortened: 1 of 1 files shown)


dirk-thomas:
I don't think we should add yet another layer of retry logic here. Why can't the module level fixture not have a sleep instead?

ros2cli/test/test_daemon.py
```diff
@@ -110,6 +110,33 @@ def daemon_node():
     assert spawn_daemon(args=[], wait_until_spawned=5.0)
     with DaemonNode(args=[]) as node:
         yield node
```
jacobperron, line 112: Is it sufficient to wait for discovery of the test node at the fixture level, here? E.g. wait until `get_node_names_and_namespaces()` discovers the test node.

hidmic:
> Why can't the module level fixture not have a sleep instead?

A single sleep is brittle.

> Is it sufficient to wait for discovery of the test node at the fixture level, here? E.g. wait until get_node_names_and_namespaces() discovers the test node.

That's something we can try, indeed.

ros2cli/test/test_daemon.py
hidmic: See https://github.com/ros2/ros2cli/pull/504/commits/559de8eba42069fe46e3e40bf6af9c5e9b68d1b8.

ros2cli/test/test_daemon.py
```diff
@@ -104,12 +104,25 @@ def noop_execute_callback(goal_handle):
 
 @pytest.fixture(scope='module')
 def daemon_node():
+    import time
```
dirk-thomas, line 107: Nitpick: why a local import?

hidmic: No particular reason, see https://github.com/ros2/ros2cli/pull/504/commits/9d7f5d34dc75bcdb01e2e05e49cd98f81f5ba2c0.

hidmic:
CI up to `ros2cli`, `rclpy`, and `rmw_connext` (when possible):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10632/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6064/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8657/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10538/)

Issue #502, opened 2020-05-04
ROS2 daemon terminates silently


cnboonhan:
**Context:** 
I am currently using the Server-Client discovery mechanism from FastRTPS, as documented [here](https://fast-rtps.docs.eprosima.com/en/latest/advanced.html#server-client-discovery), for a ROS2 network. This "enterprise" network connects multiple devices, both wired and wireless, via Cisco Catalyst 3850 switches and Cisco Access Points. The wired and wireless devices are separated into two VLANs. For this example, only 3 devices are live: a Discovery Server, Talker and Listener.

**Issue:**
Initial runs with the 3 device setup ( Discovery Server, Talker publishing, Listener subscribing ) work well. However, after a certain amount of time ( ranging from 1 hour to more ), the `ros2_daemon` process on the Discovery Server will eventually terminate. The consequence is that any new subscriptions from Listener to Talker will not work. The `ros2_daemon` process does not automatically restart, thus crippling the discovery mechanism.

I have not been able to replicate this issue on a simpler ( flat ) wired switched network with single ASUS router the Discovery Server, Talker and Listener, within a similar amount of time.

**Follow up:**
I am currently using a `cron` job that checks for `ros2_daemon` and restarts it if it is down. 

However, I would also like to know what sort of actions I can take to search for any relevant information that might reveal information about why the process terminates. 

## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - ROS2 Eloquent, installed from binaries
- Version or commit hash:
  - Latest Eloquent 
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue

For all devices, we expect RO2 Eloquent to be sourced: `source /opt/ros/eloquent/setup.bash`

**Network Hardware**
It is not possible to document the specific configuration of the network. Some details are specified in the primer above, but let me know if there is any more information I could provide. Again, I was not able to replicate this one a simpler network, within the time I had.

**On Discovery Server**
1. add the following as `$HOME/server_config.xml`
```
<?xml version="1.0" encoding="utf-8" ?>
<dds xmlns="http://www.eprosima.com/XMLSchemas/fastRTPS_Profiles">
  <profiles>
    <participant profile_name="UDP SERVER" is_default_profile="true">
      <rtps>
        <prefix>72.61.73.70.66.61.72.6d.74.65.73.74</prefix>
        <builtin>
          <discovery_config>
[...]

PR #505, opened 2020-05-09
Feature/timeout config

cnboonhan94:
Configurable daemon timeout option targeted at https://github.com/ros2/ros2cli/issues/502. 

There is no option for disabling timeout entirely yet, save for setting an astronomically large value.
This could be further handled, for example by following @codebot's suggestion of using negative values, but I might defer to the maintainers on design choice.

### DIFF
--- a/ros2cli/ros2cli/verb/daemon/start.py
+++ b/ros2cli/ros2cli/verb/daemon/start.py
@@ -26,6 +26,9 @@ class StartVerb(VerbExtension):
         parser.add_argument(
             '--debug', '-d', action='store_true',
             help='Print debug messages')
+        parser.add_argument(
+            '--timeout', '-t', action='store_const', const=2 * 60 * 60,
+            help='Terminate daemon after t seconds of inactivity. Default of 7200 seconds ( 2 hours )')
 
     def main(self, *, args):
         running = is_daemon_running(args)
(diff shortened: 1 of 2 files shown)


ros2cli/ros2cli/verb/daemon/start.py
```diff
@@ -26,6 +26,10 @@ def add_arguments(self, parser, cli_name):
         parser.add_argument(
             '--debug', '-d', action='store_true',
             help='Print debug messages')
+        parser.add_argument(
+            '--timeout', '-t', default=2 * 60 * 60,
+            help='Terminate daemon after integer t seconds of inactivity. \
+            Default of 7200 seconds(2 hours)')
```
hidmic, line 32: @cnboonhan94 IIRC there's no need to introduce the default value in the help string, the parser usage string will include it.

cnboonhan94:
sorry i have been away for a significant time. I recently tried to add a test to `test_daemon.py` using the latest `ros2cli` sources, however there seems to be significant refactoring going on in the repo, and i cant seem to run any tests for now. i'm getting errors that i'm not sure about when running `test_daemon.py`:

```ModuleNotFoundError: No module named 'rclpy.topic_endpoint_info' ```

As these changes are relatively minor, I'm wondering if its better to add this feature some other time?

PR #508, opened 2020-05-12
Avoid array.array in ros2 param output.

hidmic:
Precisely what the title says. Otherwise parameter YAML dumps won't be load by `rcl`. Also, I think it looks better if we avoid `!!python/object/apply:array.array` altogether.


hidmic:
CI up to `ros2param`:

- Linux [Build Status](https://ci.ros2.org/job/ci_linux/10726/)

hidmic:
Alright going in.

PR #513, opened 2020-05-15
Ensure ros2doctor ROS nodes have valid names.

hidmic:
Precisely what the title says. Came across this while testing `ros2 doctor hello` on a VM that had dashes in its hostname. It's tricky one to test for regression, so I'm open to ideas.

### DIFF
--- a/ros2doctor/ros2doctor/verb/hello.py
+++ b/ros2doctor/ros2doctor/verb/hello.py
@@ -14,6 +14,7 @@
 
 from argparse import ArgumentTypeError
 import os
+import re
 import socket
 import struct
 import threading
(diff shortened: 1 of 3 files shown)


jacobperron:
For a test, we could try using `mock` as described in this [SO answer](https://stackoverflow.com/a/38565885/8539665).

hidmic:
@claireyywang @jacobperron PTAL

On a side note, I find the `hello` verb implementation rather unusual. Particularly when it starts spawning short-lived daemon threads in a loop. IMHO a pool of threads would be cheaper and provide for tighter thread lifetime control.

jacobperron:
>  I'd think there was some confusion w.r.t `pytest` vs. `unittest` and their relationship w/ launch_testing.

Oof, I didn't realize that. I wonder if there are other places in our code base that make a similar mistake.

hidmic:
CI up to `ros2doctor`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10805/)

hidmic:
Focal PR job failure is unrelated. Have two approvals. Merging.

PR #516, opened 2020-05-20
[ros2interface] Remove usage of deprecated std_msgs and std_srvs packages

jacobperron:
Packages were deprecated in https://github.com/ros2/common_interfaces/pull/116.

This also fixes a [test failure](https://ci.ros2.org/view/nightly/job/nightly_linux_release/1548/testReport/ros2interface.src.ros2.ros2cli.ros2interface.test/test_cli/test_cli/) introduced by the deprecation notice.

### DIFF
--- a/ros2interface/ros2interface/verb/package.py
+++ b/ros2interface/ros2interface/verb/package.py
@@ -23,7 +23,7 @@ class PackageVerb(VerbExtension):
     def add_arguments(self, parser, cli_name):
         arg = parser.add_argument(
             'package_name',
-            help="Name of the ROS package (e.g. 'std_msgs, std_srvs, etc.')")
+            help="Name of the ROS package (e.g. 'example_interfaces')")
         arg.completer = package_name_completer
 
     def main(self, *, args):
--- a/ros2interface/test/test_cli.py
+++ b/ros2interface/test/test_cli.py
@@ -38,16 +38,16 @@ if sys.platform.startswith('win'):
             allow_module_level=True)
 
 
-some_messages_from_std_msgs = [
-    'std_msgs/msg/Bool',
-    'std_msgs/msg/Float32',
-    'std_msgs/msg/Float64',
+some_messages_from_test_msgs = [
+    'test_msgs/msg/BasicTypes',
+    'test_msgs/msg/Constants',
+    'test_msgs/msg/Strings',
 ]
 
-some_services_from_std_srvs = [
-    'std_srvs/srv/Empty',
-    'std_srvs/srv/SetBool',
-    'std_srvs/srv/Trigger',
+some_services_from_test_msgs = [
+    'test_msgs/srv/Arrays',
+    'test_msgs/srv/BasicTypes',
+    'test_msgs/srv/Empty',
 ]
 
 some_actions_from_test_msgs = [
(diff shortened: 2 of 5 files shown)


jacobperron:
We should also consider removing/replacing usage in other ros2cli packages.

ros2interface/ros2interface/verb/package.py
```diff
@@ -23,7 +23,7 @@ class PackageVerb(VerbExtension):
     def add_arguments(self, parser, cli_name):
         arg = parser.add_argument(
             'package_name',
-            help="Name of the ROS package (e.g. 'std_msgs, std_srvs, etc.')")
+            help='Name of the ROS package (e.g. "example_interfaces")')
```
dirk-thomas, line 26: Please keep the single quote for the user to see in the help message.

Same below.

jacobperron: Do you mean to swap the single quotes and double quotes? I only made the change to this line for consistency with the other help message strings.

dirk-thomas: > Do you mean to swap the single quotes and double quotes?

Yes, restore to the state it was before the patch.

> I only made the change to this line for consistency with the other help message strings.

I don't see any other help messages with quotes in the touched files so I assume you are referring to some other files. Without knowing which ones I would guess they should be updated for consistency instead.

jacobperron: The other two help messages I touched in the PR have the same pattern: double quotes inside single quotes.

dirk-thomas: Oh, I didn't notice that. Afaik we use show single quotes in user visible message in almost all the code base.

jacobperron: I can make the inverse change (or no change at all since it not very related to this PR) if you prefer.

dirk-thomas: I see 8 cases where double quotes are shown to users in Python argument help strings vs. 65 showing single quotes to users (based on a quick grep). So I would suggest using the later.

dirk-thomas: PRs for the two repos with diverging quotes:
* ros2/demos#447
* ros2/rosbag2#416

jacobperron: I created a separate PR for ros2cli: https://github.com/ros2/ros2cli/pull/517/

I can address any conflicts with this PR.

jacobperron: rebased with #517 merged.

ros2interface/test/test_cli.py
```diff
@@ -229,15 +197,13 @@ def test_package_on_test_msgs(self):
             lines=output_lines,
             strict=True
         )
-        assert all(action in output_lines for action in some_actions_from_test_msgs)
+        assert all(action in output_lines for action in some_interfaces)
```
hidmic, line 200: @jacobperron nit: 
```suggestion
        assert all(interface in output_lines for interface in some_interfaces)
```

jacobperron: see ae8649c

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10869/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6233/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8854/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10759/)

PR #518, opened 2020-05-21
[ros2interface] Allow input from stdin for `show`

### LINKED ISSUE #349: [ros2interface] does not allow input from stdin for `show`

audrow:
Fixes #349, allowing `ros2 interface show` to read from stdin.
```console
$ echo std_msgs/msg/String | ros2 interface show -
# This was originally provided as an example message.
# It is deprecated as of Foxy
# It is recommended to create your own semantically meaningful message.
# However if you would like to continue using this please use the equivalent in example_msgs.

string data
```

# Design decisions

Unlike `rosmsg`, this solution uses a dash (`-`) to denote that the program should look to stdin, which is a common convention in many programs ([see here](https://unix.stackexchange.com/a/16364)).

There are two main reasons for using a dash rather than automatically looking for stdin:
1. It allows for the argparse help message to be shown in the event the user enters `ros2 interface show` (with no stdin piped in).
2. It required changing only the `show` verb of `ros2interface`.  

# Error handling

No stdin supplied:
```console
$ ros2 interface show -
stdin is empty
```

stdin is not exclusively an interface type:
```console
$ echo "Not in the right form" | ros2 interface show -
Could not determine message type from stdin
``` 

No dash shows help:
```console
$ ros2 interface show
usage: ros2 interface show [-h] type
ros2 interface show: error: the following arguments are required: type
```

Package related errors are differed to `rosidl_runtime_py`'s `get_interface_path`:
```console
$ echo "not/a/message/type" | ros2 interface show -
Unknown package 'not'
```
```console
$ echo "std_msgs/msg/Foo" | ros2 interface show -
Could not find the interface '/root/ros2_foxy/install/std_msgs/share/std_msgs/msg/Foo.idl'
```
```console
$ echo "not/a/message/type/" | ros2 interface show -
[...]
ValueError: Invalid name 'not/a/message/type/'. Must not contain empty parts
```
```console
$ echo "not_a_message_type" | ros2 interface show -
[...]
ValueError: Invalid name 'not_a_message_type'. Expected at least two parts separated by '/'
```

### DIFF
--- a/ros2interface/ros2interface/verb/show.py
+++ b/ros2interface/ros2interface/verb/show.py
@@ -17,16 +17,35 @@ from ros2interface.verb import VerbExtension
 from rosidl_runtime_py import get_interface_path
 
 
+import sys
+import re
+
+
 class ShowVerb(VerbExtension):
     """Output the interface definition."""
 
     def add_arguments(self, parser, cli_name):
         arg = parser.add_argument(
                 'type',
-                help='Show an interface definition (e.g. "std_msgs/msg/String")')
+                help='Show an interface definition (e.g. "std_msgs/msg/String"; \'-\' for stdin)',
+        )
         arg.completer = type_completer
 
     def main(self, *, args):
+        if args.type == "-":
+            if sys.stdin.isatty():
+                return str("stdin is empty")
+
+            # The regex pattern is permissive so that
+            # `get_interface_path` can handle errors
+            regex_pattern = "([a-zA-Z0-9_\/]+)"
+            stdin_str = "".join(sys.stdin.readlines())
+            matches = re.findall(regex_pattern, stdin_str)
+
+            if len(matches) != 1:
+                return str("Could not determine message type from stdin")
+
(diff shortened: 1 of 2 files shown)


hidmic:
@audrow there was a previous PR addressing #349, #387. It got lost in my queue to go back to it after running CI, my bad.

audrow:
@hidmic I think we should go with the PR in #387.  My PR and theirs both do pretty much the same thing, but I think theirs is a little cleaner.  I just tested #387 on my computer and it worked well.  The tests all passed except for one related to #516.

PR #522, opened 2020-06-01
[ros2interface] show "expanded" message definition

### LINKED ISSUE #348: [ros2interface] does not show "expanded" message definition
#### Feature description

In ROS 1 `rosmsg show` would show you the recursively expanded message fields, whereas in ROS 2 at the moment `ros2 interface show` only shows you the fields of the immediate message you specify.

For example, in ROS 1 (from https://wiki.ros.org/ROS/Tutorials/UnderstandingTopics#Using_rostopic_type):

```
$ rosmsg show geometry_msgs/Twist
geometry_msgs/Vector3 linear
  float64 x
  float64 y
  float64 z
geometry_msgs/Vector3 angular
  float64 x
  float64 y
  float64 z
```

And the equivalent in ROS 2:

```
% ros2 interface show geometry_msgs/msg/Twist
# This expresses velocity in free space broken into its linear and angular parts.

Vector3  linear
Vector3  angular
```

It does, however, show the comments in the message file, which may be nice to have some times.

Also, as a side note, this command takes almost 2 seconds to run on my macbook, but I'm not sure why...

audrow:
Fixes #348

### DIFF
--- a/ros2interface/ros2interface/verb/show.py
+++ b/ros2interface/ros2interface/verb/show.py
@@ -25,7 +25,7 @@ from rosidl_parser.definition import BASIC_TYPES
 
 
 PRIMITIVE_FIELD_TYPES = [
-    'bool'
+    'bool',
     'byte',
     'char',
     'float32',
@@ -74,7 +74,7 @@ class InterfaceTextLine:
     ):
 
         if not raw:
-            if self._is_comment_or_whitespace():
+            if self._is_comment_or_whitespace() and not self._is_separator():
                 return ""
             text = remove_comment_from_string(
                 self._text,
@@ -122,6 +122,9 @@ class InterfaceTextLine:
 
     def _is_comment_or_whitespace(self) -> bool:
         return self._get_interface_string_from_text() is None
+
+    def _is_separator(self) -> bool:
+        return re.match(r"^[-]{3}\s*$", self._text) is not None
 
     def _is_basic_idl_type(self) -> bool:
         return self._get_interface_string_from_text() in PRIMITIVE_FIELD_TYPES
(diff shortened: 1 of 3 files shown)


ros2interface/ros2interface/verb/show.py
```diff
@@ -43,12 +242,105 @@ def add_arguments(self, parser, cli_name):
...
+                          help="show raw message text, including comments")
         arg.completer = type_completer
 
     def main(self, *, args):
-        try:
-            file_path = get_interface_path(args.type)
-        except LookupError as e:
-            return str(e)
+
+        InterfaceExpander(args.type).print(raw=args.raw)
```
sloretz, line 252: It looks like this is parsing the legacy message definition. I think this should reuse the legacy message parsing code in `rosidl`. This lives in a package called [`rosidl_adapter`](https://github.com/ros2/rosidl/tree/master/rosidl_adapter). In particular I think the methods [`parse_message_string`](https://github.com/ros2/rosidl/blob/b4a4431b49f0aa786f9cf990acf2bcccc5d3c98a/rosidl_adapter/rosidl_adapter/parser.py#L444), [`parse_service_string`](https://github.com/ros2/rosidl/blob/b4a4431b49f0aa786f9cf990acf2bcccc5d3c98a/rosidl_adapter/rosidl_adapter/parser.py#L822), and [`parse_action_string()`](https://github.com/ros2/rosidl/blob/b4a4431b49f0aa786f9cf990acf2bcccc5d3c98a/rosidl_adapter/rosidl_adapter/parser.py#L866) could be reused here depending on the type of message given in.

Those give `ActionSpecification`, `ServiceSpecification`, and `MessageSpecification` instances. The most important one here is the `MessageSpecification`. I think if `raw` is `False` then this code should use those functions to parse the legacy definition, modify the `MessageSpecification` instance (say by removing comments), then finally calling [`str(msg_spec_instance)`](https://github.com/ros2/rosidl/blob/b4a4431b49f0aa786f9cf990acf2bcccc5d3c98a/rosidl_adapter/rosidl_adapter/parser.py#L424) to write out the result.

PR #524, opened 2020-06-03
[ros2interface] show "expanded" message definition

audrow:
Fixes #348

PR #527, opened 2020-06-05
[ros2interface] Add option to include/remove whitespace and comments

audrow:
Adds the ability to show an interface's whitespace and comments with the argument `--raw` or `-r`, like in ROS1's [`rosmsg`](https://github.com/ros/ros_comm/blob/9162b32a42b5569ae42a94aa6426aafcb63021ae/tools/rosmsg/src/rosmsg/__init__.py#L585-L587).

### DIFF
--- a/ros2interface/ros2interface/verb/show.py
+++ b/ros2interface/ros2interface/verb/show.py
@@ -58,6 +58,12 @@ class InterfaceTextLine:
     def __str__(self) -> str:
         return self._raw_line_text
 
+    def is_comment(self) -> bool:
+        return self._msg_spec and self._msg_spec.annotations['comment']
+
+    def is_trailing_comment(self) -> bool:
+        return self._field and self._field.annotations['comment']
+
     @property
     def nested_type(self) -> typing.Optional[str]:
         if self._field and self._is_nested():
@@ -65,6 +71,13 @@ class InterfaceTextLine:
             if self._field.type.is_array:
                 interface_type = interface_type[:interface_type.find('[')]
             return interface_type.replace('/', '/msg/')
+
+    @property
+    def trailing_comment(self) -> typing.Optional[str]:
+        if self.is_trailing_comment():
+            return self._field.annotations['comment'][0]
+        else:
+            return None
 
     @property
     def _field(self) -> typing.Optional[Field]:
@@ -97,18 +110,31 @@ def _get_interface_lines(interface_identifier: str) -> typing.Iterable[Interface
             )
 
 
-def _show_interface(interface_identifier: str, indent_level: int = 0):
+def _print_interface_line(line: InterfaceTextLine, raw: bool, indent_level: int):
+    text = str(line)
+    if not raw:
+        if not text or line.is_comment():
+            return
+        elif line.is_trailing_comment():
+            comment_start_idx = text.find(line.trailing_comment)
+            text = text[:comment_start_idx - 1].strip()
+    if text:
+        indent_string = indent_level * '\t'
+        print(f'{indent_string}{text}')
+    else:
+        print()
+
+
+def _show_interface(interface_identifier: str, raw: bool = False, indent_level: int = 0):
+
     for line in _get_interface_lines(interface_identifier):
 
-        if str(line):
-            indent_string = indent_level*'\t'
-            print(f'{indent_string}{str(line)}')
-        else:
-            print()
+        _print_interface_line(line, raw=raw, indent_level=indent_level)
 
         if line.nested_type:
             _show_interface(
                 line.nested_type,
+                raw=raw,
                 indent_level=indent_level+1,
             )
 
@@ -130,6 +156,11 @@ class ShowVerb(VerbExtension):
     """Output the interface definition."""
 
     def add_arguments(self, parser, cli_name):
+        parser.add_argument(
+            '-r', '--raw',
+            dest='raw', default=False, action='store_true',
+            help='show raw interface text, including whitespace and comments'
+        )
         arg = parser.add_argument(
             'type',
             action=ReadStdinPipe,
@@ -140,6 +171,6 @@ class ShowVerb(VerbExtension):
 
     def main(self, *, args):
         try:
-            _show_interface(args.type)
+            _show_interface(args.type, raw=args.raw)
         except (ValueError, LookupError) as e:
             return str(e)
--- a/ros2interface/test/test_cli.py
+++ b/ros2interface/test/test_cli.py
@@ -260,6 +260,36 @@ class TestROS2InterfaceCLI(unittest.TestCase):
             strict=True
         )
 
+    def test_show_raw_message(self):
+        with self.launch_interface_command(
+                arguments=['show', 'test_msgs/msg/Builtins', '--raw']
+        ) as interface_command:
+            assert interface_command.wait_for_shutdown(timeout=2)
+        assert interface_command.exit_code == launch_testing.asserts.EXIT_OK
+        assert launch_testing.tools.expect_output(
+            expected_lines=[
+                'builtin_interfaces/Duration duration_value',
+                '\t# Duration defines a period between two time points. It is comprised of a',
+                '\t# seconds component and a nanoseconds component.',
+                '',
+                '\t# Seconds component, range is valid over any possible int32 value.',
+                '\tint32 sec',
+                '',
+                '\t# Nanoseconds component in the range of [0, 10e9).',
+                '\tuint32 nanosec',
+                'builtin_interfaces/Time time_value',
+                "\t# Time indicates a specific point in time, relative to a clock's 0 point.",
+                '',
+                '\t# The seconds component, valid over all int32 values.',
+                '\tint32 sec',
+                '',
+                '\t# The nanoseconds component, valid in the range [0, 10e9).',
+                '\tuint32 nanosec',
+            ],
+            text=interface_command.output,
+            strict=True
+        )
+
     def test_show_service(self):
         with self.launch_interface_command(
             arguments=['show', 'test_msgs/srv/BasicTypes']
@@ -310,13 +340,10 @@ class TestROS2InterfaceCLI(unittest.TestCase):
         assert interface_command.exit_code == launch_testing.asserts.EXIT_OK
         assert launch_testing.tools.expect_output(
             expected_lines=[
-                '#goal definition',
                 'int32 order',
                 '---',
-                '#result definition',
                 'int32[] sequence',
                 '---',
-                '#feedback',
                 'int32[] sequence',
             ],
             text=interface_command.output,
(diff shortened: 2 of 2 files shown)


dirk-thomas:
> changes the default behavior so that comments are stripped out.

I am not sure that is better default behavior. The comments often provide a lot of context. I would rather see them shown by default and let uses hide them if they don't mind.

sloretz:
> I am not sure that is better default behavior. The comments often provide a lot of context.

I found it pretty difficult to understand `geometry_msgs/msg/TwistStamped` with comments: https://github.com/ros2/ros2cli/pull/524#pullrequestreview-423928859 . The repeated comments added noise that made it difficult for my eyes to find the message fields.

audrow:
I ran CI after Jacob's review.  Here are the results.
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11048/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6372/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8999/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10971/)

audrow:
I'm happy to make any changes depending on what we think the default behavior should be.

dirk-thomas:
> I found it pretty difficult to understand geometry_msgs/msg/TwistStamped with comments: #524 (review) . The repeated comments added noise that made it difficult for my eyes to find the message fields.

And there are probably countless counter examples where without any comments the semantic is not obvious to the average user. That is why there certainly should be options to choose.

In your example a lot of the overhead comes from the fact that the comments are shown repeatedly for submessages. Maybe not showing comments for submessages would be a reasonable option or only showing it once per type?

sloretz:
> And there are probably countless counter examples where without any comments the semantic is not obvious to the average user. That is why there certainly should be options to choose.

Agreed that the option should be there

> In your example a lot of the overhead comes from the fact that the comments are shown repeatedly for submessages. Maybe not showing comments for submessages would be a reasonable option or only showing it once per type?

what that would look like:

```
$ ros2 interface show geometry_msgs/msg/TwistStamped
# A twist with reference coordinate frame and timestamp

std_msgs/Header header
	# Standard metadata for higher-level stamped data types.
	# This is generally used to communicate timestamped data
	# in a particular coordinate frame.

	# Two-integer timestamp that is expressed as seconds and nanoseconds.
	builtin_interfaces/Time stamp
		# Time indicates a specific point in time, relative to a clock's 0 point.

		# The seconds component, valid over all int32 values.
		int32 sec

		# The nanoseconds component, valid in the range [0, 10e9).
		uint32 nanosec

	# Transform frame with which this data is associated.
	string frame_id
Twist twist
	# This expresses velocity in free space broken into its linear and angular parts.

	Vector3  linear
		# This represents a vector in free space.

		# This is semantically different than a point.
		# A vector is always anchored at the origin.
		# When a transform is applied to a vector, only the rotational component is applied.

		float64 x
		float64 y
		float64 z
	Vector3  angular
		float64 x
		float64 y
		float64 z
```

It certainly helps, though there are still a lot of comments that might not help me understand the twist message, especially if I'm already familiar with the nested message (like std_msgs/msg/Header).

Maybe showing only the comments from the top level message by default would reduce the noise without losing the meaning? Presumably the top level message definition has enough comments to understand its semantics. Then the fields of the nested messages are there only to show how to access the data.

```
$ ros2 interface show geometry_msgs/msg/TwistStamped
# A twist with reference coordinate frame and timestamp

std_msgs/Header header
	builtin_interfaces/Time stamp
		int32 sec
		uint32 nanosec
	string frame_id
Twist twist
	Vector3  linear
		float64 x
		float64 y
		float64 z
	Vector3  angular
		float64 x
		float64 y
		float64 z
```

dirk-thomas:
> Maybe showing only the comments from the top level message by default would reduce the noise without losing the meaning? Presumably the top level message definition has enough comments to understand its semantics.

:+1:

audrow:
I can make it so only top level comments and whitespace are shown.  Do we want to keep a `--raw` option to show the comments and whitespace for nested definitions?

dirk-thomas:
> Do we want to keep a `--raw` option to show the comments and whitespace for nested definitions?

:+1: for having options to customize the output. Imo `--raw` doesn't convey what the option is supposed to do. Options like `--all-comments` / `--no-comments` seem much more intuitive.

audrow:
I've change `interface show` to use `--all-comments` and `--no-comments`.  Below are some examples of printout.

<details><summary>No flags (top level comments included) </summary>

```
$ ros2 interface show geometry_msgs/msg/TwistStamped 
# A twist with reference coordinate frame and timestamp

std_msgs/Header header
        builtin_interfaces/Time stamp
                int32 sec
                uint32 nanosec
        string frame_id
Twist twist
        Vector3  linear
                float64 x
                float64 y
                float64 z
        Vector3  angular
                float64 x
                float64 y
                float64 z
```

</details>

<details><summary>No comments flag </summary>

```
$ ros2 interface show geometry_msgs/msg/TwistStamped --no-comments
std_msgs/Header header
        builtin_interfaces/Time stamp
                int32 sec
                uint32 nanosec
        string frame_id
Twist twist
        Vector3  linear
                float64 x
                float64 y
                float64 z
        Vector3  angular
                float64 x
                float64 y
                float64 z
```

</details>

<details><summary>All comments flag </summary>

```
$ ros2 interface show geometry_msgs/msg/TwistStamped --all-comments
# A twist with reference coordinate frame and timestamp

std_msgs/Header header
        # Standard metadata for higher-level stamped data types.
        # This is generally used to communicate timestamped data
        # in a particular coordinate frame.

        # Two-integer timestamp that is expressed as seconds and nanoseconds.
        builtin_interfaces/Time stamp
                # Time indicates a specific point in time, relative to a clock's 0 point.

                # The seconds component, valid over all int32 values.
                int32 sec

                # The nanoseconds component, valid in the range [0, 10e9).
                uint32 nanosec

        # Transform frame with which this data is associated.
        string frame_id
Twist twist
        # This expresses velocity in free space broken into its linear and angular parts.

        Vector3  linear
                # This represents a vector in free space.

                # This is semantically different than a point.
                # A vector is always anchored at the origin.
                # When a transform is applied to a vector, only the rotational component is applied.

                float64 x
                float64 y
                float64 z
        Vector3  angular
                # This represents a vector in free space.

                # This is semantically different than a point.
                # A vector is always anchored at the origin.
                # When a transform is applied to a vector, only the rotational component is applied.

                float64 x
                float64 y
                float64 z
```

</details>

<details><summary>Error message with no comments and all comments flag </summary>

```
$ ros2 interface show geometry_msgs/msg/TwistStamped --all-comments --no-comments 
Cannot use the flags '--all-comments' and '--no-comments' at the same time
```

</details>

ros2interface/ros2interface/verb/show.py
```diff
@@ -139,7 +186,25 @@ def add_arguments(self, parser, cli_name):
         arg.completer = type_completer
 
     def main(self, *, args):
+
+        if args.all_comments and args.no_comments:
+            return f"Cannot use the flags '{self._all_comments_flag}' and " \
+                   f"'{self._no_comments_flag}' at the same time"
```
dirk-thomas, line 192: Instead of a custom check the two arguments should be in a mutual exclusive group (`add_mutually_exclusive_group`).

ros2interface/ros2interface/verb/show.py
```diff
@@ -139,7 +186,25 @@ def add_arguments(self, parser, cli_name):
...
+        elif args.all_comments:
+            is_show_top_level_comments = True
+            is_show_nested_comments = True
+        elif args.no_comments:
+            is_show_top_level_comments = False
+            is_show_nested_comments = False
+        else:
+            is_show_top_level_comments = True
+            is_show_nested_comments = False
```
dirk-thomas, line 201: These lines can be condensed into:

```
is_show_top_level_comments = args.all_comments or not args.no_comments
is_show_nested_comments = args.all_comments
```

ros2interface/ros2interface/verb/show.py
```diff
@@ -97,18 +104,45 @@ def _get_interface_lines(interface_identifier: str) -> typing.Iterable[Interface
             )
 
 
-def _show_interface(interface_identifier: str, indent_level: int = 0):
+def _print_interface_line(
+        line: InterfaceTextLine,
```
dirk-thomas, line 108: Nitpick: should be single level indent - not two.

Same below.

ros2interface/ros2interface/verb/show.py
```diff
@@ -139,7 +186,15 @@ def add_arguments(self, parser, cli_name):
         arg.completer = type_completer
 
     def main(self, *, args):
+
+        is_show_top_level_comments = args.all_comments or not args.no_comments
```
jacobperron, line 190: I think the this can be simplified further, right?
```suggestion
        is_show_top_level_comments = not args.no_comments
```

audrow:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11083/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6392/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9039/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10994/)

ros2interface/ros2interface/verb/show.py
```diff
@@ -97,18 +104,45 @@ def _get_interface_lines(interface_identifier: str) -> typing.Iterable[Interface
...
+        print()
+
+
+def _show_interface(
+    interface_identifier: str,
+    is_show_comments: bool = False,
+    is_show_nested_comments: bool = False,
+    indent_level: int = 0
+):
+
```
dirk-thomas, line 132: Nitpick: remove empty line at the beginning of the function.

Same below.

ros2interface/ros2interface/verb/show.py
```diff
@@ -97,18 +104,45 @@ def _get_interface_lines(interface_identifier: str) -> typing.Iterable[Interface
...
 
-        if str(line):
-            indent_string = indent_level*'\t'
-            print(f'{indent_string}{str(line)}')
-        else:
-            print()
+        _print_interface_line(
+            line,
+            is_show_comments=is_show_comments,
+            indent_level=indent_level
```
dirk-thomas, line 138: Maybe reduce the wrapping here?

```
        _print_interface_line(
            line, is_show_comments=is_show_comments, indent_level=indent_level)
```

ros2interface/ros2interface/verb/show.py
```diff
@@ -139,7 +186,15 @@ def add_arguments(self, parser, cli_name):
         arg.completer = type_completer
 
     def main(self, *, args):
+
+        is_show_top_level_comments = not args.no_comments
+        is_show_nested_comments = args.all_comments
```
dirk-thomas, line 191: Maybe just get rid of these single-use variables?

Issue #523, opened 2020-06-02
[ros2 topic echo] does not echo a published message by [ros2 topic pub] if joining late, even with --qos-durability=transient_local


craigh92:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04 
- Installation type:
  - Binary
- Version or commit hash:
  - Eloquent
- DDS implementation:
  - Default
- Client library (if applicable):
  - rclpy

#### Steps to reproduce issue
If I start a publisher with the ros2cli tool at 0.1Hz, with the option `--qos-durability=transient_local`, and then wait 10 seconds for it to publish a message, and then I start a topic echoer with the same qos settings, I would expect to see the last message published by the publisher, even though the echoer wasn't around when the message was published. However, this doesn't happen.

i.e: First:

```ros2 topic pub -r 0.1 --qos-durability=transient_local hello std_msgs/msg/String```

Then after 10 seconds in a new terminal:

```ros2 topic echo --qos-durability=transient_local hello std_msgs/msg/String``` 

#### Expected behavior

As soon as the ros2 topic echo node starts, it should print the message that was published by the publisher before this node was started.

#### Actual behavior

It ignores this first message, and only prints a new message after another 10 seconds when the publisher publishes a new message

#### Implementation considerations

I created my own echo node with the transient_local durability option to see if this would work. This behaves with the expected behavior I mentioned above. 

This is probably an awful implementation, and it does not accept other qos options, but it proves that even when late joining, an echo node can get the last message of a publisher.

```python
import rclpy
from rclpy.node import Node
from rclpy.executors import MultiThreadedExecutor
from rclpy.qos import QoSProfile, QoSDurabilityPolicy
import sys
from typing import List, Any

"""
Use:
    ros-topic-echo-transient-local.py  <namespace/path/topic_name> <fully/qualified/message_type>

Theres a bug with the command line utility `ros2 topic echo` when used with the `--qos-durability=transient_local` policy.
This script behaves in the same way that the command:
    ros2 topic echo --qos-durability=transient_local <namespace/path/topic_name> <fully/qualified/message_type>
should.

It uses dynamic programing to modify the source code at run time so that the node can subscribe to the type passed
in as an argument.
"""


def main(topicName : str, messageType : Any):
    """
    Start the 'topic-echo-node' and spin untill exception is thrown
    """
    rclpy.init()
    echoNode = Node('topic_echo_node')

    latching_qos = QoSProfile(depth=1,
[...]

PR #528, opened 2020-06-09
support QoS Depth and History via ros2 topic pub/echo.

fujitatomoya:
close https://github.com/ros2/ros2cli/issues/523

Signed-off-by: Tomoya.Fujita <Tomoya.Fujita@sony.com>

### DIFF
--- a/ros2topic/ros2topic/api/__init__.py
+++ b/ros2topic/ros2topic/api/__init__.py
@@ -166,6 +166,8 @@ def qos_profile_from_short_keys(
         profile.history = rclpy.qos.QoSHistoryPolicy.get_from_short_key(history)
     if durability:
         profile.durability = rclpy.qos.QoSDurabilityPolicy.get_from_short_key(durability)
+        if profile.durability == rclpy.qos.QoSDurabilityPolicy.TRANSIENT_LOCAL and profile.depth == 0:
+            profile.depth = 1
     if reliability:
         profile.reliability = rclpy.qos.QoSReliabilityPolicy.get_from_short_key(reliability)
 
--- a/ros2topic/ros2topic/api/__init__.py
+++ b/ros2topic/ros2topic/api/__init__.py
@@ -160,16 +160,17 @@ def qos_profile_from_short_keys(
     """Construct a QoSProfile given the name of a preset, and optional overrides."""
     # Build a QoS profile based on user-supplied arguments
     profile = rclpy.qos.QoSPresetProfiles.get_from_short_key(preset_profile)
-    if depth >= 0:
-        profile.depth = depth
     if history:
         profile.history = rclpy.qos.QoSHistoryPolicy.get_from_short_key(history)
     if durability:
         profile.durability = rclpy.qos.QoSDurabilityPolicy.get_from_short_key(durability)
-        if profile.durability == rclpy.qos.QoSDurabilityPolicy.TRANSIENT_LOCAL and profile.depth == 0:
-            profile.depth = 1
     if reliability:
         profile.reliability = rclpy.qos.QoSReliabilityPolicy.get_from_short_key(reliability)
+    if depth >= 0:
+        profile.depth = depth
+    else:
+        if profile.durability == rclpy.qos.QoSDurabilityPolicy.TRANSIENT_LOCAL and profile.depth == 0:
+            profile.depth = 1
 
     return profile
 
@@ -190,7 +191,8 @@ def add_qos_arguments_to_argument_parser(
     parser.add_argument(
         '--qos-depth', metavar='N', type=int, default=-1,
         help='Queue size setting to {} with '
-             '(overrides depth value of --qos-profile option)')
+             '(overrides depth value of --qos-profile option)'
+             .format(verb))
     parser.add_argument(
         '--qos-history',
         choices=rclpy.qos.QoSHistoryPolicy.short_keys(),
--- a/ros2topic/ros2topic/api/__init__.py
+++ b/ros2topic/ros2topic/api/__init__.py
@@ -26,6 +26,15 @@ from rosidl_runtime_py import get_message_interfaces
 from rosidl_runtime_py import message_to_yaml
 from rosidl_runtime_py.utilities import get_message
 
+def unsigned_int(string):
+    try:
+        value = int(string)
+    except ValueError:
+        value = -1
+    if value < 0:
+        raise ArgumentTypeError('value must be non-negative integer')
+    return value
+
 
 def get_topic_names_and_types(*, node, include_hidden_topics=False):
     topic_names_and_types = node.get_topic_names_and_types()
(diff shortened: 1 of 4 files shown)


fujitatomoya:
@clalancette 

could you take a look at this?

clalancette:
> could you take a look at this?

I will take a look, but I'd prefer to first fix the depth problem as discussed in https://github.com/ros2/ros2cli/issues/523#issuecomment-641316590 in first.

ros2topic/ros2topic/api/__init__.py
```diff
@@ -146,12 +155,19 @@ def __call__(self, prefix, parsed_args, **kwargs):
 
 def qos_profile_from_short_keys(
     preset_profile: str, reliability: str = None, durability: str = None,
+    depth: unsigned_int = 0, history: str = None,
 ) -> rclpy.qos.QoSProfile:
     """Construct a QoSProfile given the name of a preset, and optional overrides."""
     # Build a QoS profile based on user-supplied arguments
     profile = rclpy.qos.QoSPresetProfiles.get_from_short_key(preset_profile)
+    if depth > 0:
+        profile.depth = depth
```
clalancette, line 164: I'm not so sure about this logic.  If the user passes `--qos-depth 0`, then we'll ignore their preference.  I think we'll need to use some sentinel value (maybe -1) to correspond to "the user didn't set a depth", at which point we'd use the default for the profile that they chose.

ros2topic/ros2topic/api/__init__.py
```diff
@@ -171,6 +187,17 @@ def add_qos_arguments_to_argument_parser(
              .format(verb, default_preset))
     default_profile = rclpy.qos.QoSPresetProfiles.get_from_short_key(
         default_preset)
+    parser.add_argument(
+        '--qos-depth', metavar='N', type=unsigned_int, default = default_profile.depth,
```
clalancette, line 191:
```suggestion
        '--qos-depth', metavar='N', type=unsigned_int, default=default_profile.depth,
```

ros2topic/ros2topic/api/__init__.py
```diff
@@ -171,6 +187,17 @@ def add_qos_arguments_to_argument_parser(
...
+        help='Queue size setting to {} with '
+             '(overrides depth value of --qos-profile option, default: {})'
+             .format(verb, default_profile.depth))
```
clalancette, line 194: The depth here isn't necessarily a default of 0 in the case of transient_local.  Not sure how we want to communicate that, though.

ros2topic/ros2topic/api/__init__.py
fujitatomoya: makes sense to me, fixing.

ros2topic/ros2topic/api/__init__.py
fujitatomoya: we could check the default profile and and if the durability is transient_local AND depth==0, we can show depth=1 instead?

fujitatomoya: I will set -1 as default value, and not printing out.

ros2topic/ros2topic/api/__init__.py
```diff
@@ -171,6 +187,16 @@ def add_qos_arguments_to_argument_parser(
              .format(verb, default_preset))
     default_profile = rclpy.qos.QoSPresetProfiles.get_from_short_key(
         default_preset)
+    parser.add_argument(
+        '--qos-depth', metavar='N', type=int, default=-1,
+        help='Queue size setting to {} with '
+             '(overrides depth value of --qos-profile option)')
```
clalancette, line 193:
```suggestion
        help='Queue size setting to {} with '
             '(overrides depth value of --qos-profile option)'
             .format(verb))
```

ros2topic/ros2topic/api/__init__.py
```diff
@@ -146,12 +155,19 @@ def __call__(self, prefix, parsed_args, **kwargs):
...
+    if depth >= 0:
+        profile.depth = depth
+    if history:
+        profile.history = rclpy.qos.QoSHistoryPolicy.get_from_short_key(history)
     if durability:
         profile.durability = rclpy.qos.QoSDurabilityPolicy.get_from_short_key(durability)
+        if profile.durability == rclpy.qos.QoSDurabilityPolicy.TRANSIENT_LOCAL and profile.depth == 0:
+            profile.depth = 1
     if reliability:
         profile.reliability = rclpy.qos.QoSReliabilityPolicy.get_from_short_key(reliability)
```
clalancette, line 172: I just realized this has the same problem of overriding the users' preference.  That is, if the user says `--qos-durability transient_local --qos-depth 0`, then we'll automatically override to a depth of 1.

Arguably, that's a somewhat silly thing to do, but it is still something I think we should support.  I think you can fix this by rearranging this whole block to be something more like this:

```
    if history:
        profile.history = rclpy.qos.QoSHistoryPolicy.get_from_short_key(history)
    if durability:
        profile.durability = rclpy.qos.QoSDurabilityPolicy.get_from_short_key(durability)
    if reliability:
        profile.reliability = rclpy.qos.QoSReliabilityPolicy.get_from_short_key(reliability)
    if depth >= 0:
        profile.depth = depth
    else:
        if profile.durability == rclpy.qos.QoSDurabilityPolicy.TRANSIENT_LOCAL:
            profile.depth = 1
```

ros2topic/ros2topic/api/__init__.py
fujitatomoya: my bad, nice catch! thanks,

ros2topic/ros2topic/api/__init__.py
fujitatomoya: that makes sense to me, i think it is better for user. appreciate for your considering!

fujitatomoya:
@clalancette 

sorry to disturb you once again, could you take a look at https://github.com/ros2/ros2cli/pull/528/files/11020132e4efc1ecf7d4b0f96fcf099b6678b95c..a4e73056a2fad85525f715e666fff73341a62e01 ?

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11180/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6459/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9130/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11094/)

clalancette:
@fujitatomoya It looks like tests are failing with flake8 errors.  Can you take a look?

fujitatomoya:
@clalancette 

i will check it and get back to you.

fujitatomoya:
@clalancette 

fix flake8 code style, confirmed with `colcon test --packages-select ros2topic` locally.

fujitatomoya:
looks all green, do we have a go?

clalancette:
Here's new CI.  Assuming it goes green, I'll merge:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11200/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6477/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9162/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11121/)

PR #538, opened 2020-06-22
Make ros2doctor depend on ros_environment and fix platform.py bug on error.

clalancette:
The ros2doctor [package check](https://github.com/ros2/ros2cli/blob/d109a695b0630b139f30c979ec9133836ae77fb0/ros2doctor/ros2doctor/api/package.py#L38) requires ROS_DISTRO to be defined in the environment.  If it is not, you get warnings like:

```
UserWarning: ERROR: ROS_DISTRO is not set
```

Since that environment variable comes from the `ros_environment` package, that means this package has an `exec_depend` on `ros_environment`.  Add that here.

While we are in here, also notice that platform.py returns a "None" value if `_check_platform_helper` fails.  However, it's contract states:

```
:return: Report object storing report content.
```

Thus, make sure to return an empty report on error, which fixes possible crashes and also conforms to what other reports do.

### DIFF
--- a/ros2doctor/ros2doctor/command/doctor.py
+++ b/ros2doctor/ros2doctor/command/doctor.py
@@ -65,7 +65,8 @@ class DoctorCommand(CommandExtension):
         if args.report_failed and fail != 0:
             fail_reports = generate_reports(categories=fail_category)
             for report_obj in fail_reports:
-                format_print(report_obj)
+                if report_obj is not None:
+                    format_print(report_obj)
 
 
 class WtfCommand(DoctorCommand):
--- a/ros2doctor/ros2doctor/command/doctor.py
+++ b/ros2doctor/ros2doctor/command/doctor.py
@@ -65,8 +65,7 @@ class DoctorCommand(CommandExtension):
         if args.report_failed and fail != 0:
             fail_reports = generate_reports(categories=fail_category)
             for report_obj in fail_reports:
-                if report_obj is not None:
-                    format_print(report_obj)
+                format_print(report_obj)
 
 
 class WtfCommand(DoctorCommand):
--- a/ros2doctor/package.xml
+++ b/ros2doctor/package.xml
@@ -14,7 +14,8 @@
   <exec_depend>python3-ifcfg</exec_depend>
   <exec_depend>python3-rosdistro-modules</exec_depend>
   <exec_depend>rclpy</exec_depend>
-  
+  <exec_depend>ros_environment</exec_depend>
+
   <test_depend>ament_copyright</test_depend>
   <test_depend>ament_flake8</test_depend>
   <test_depend>ament_pep257</test_depend>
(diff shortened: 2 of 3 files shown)


ros2doctor/ros2doctor/command/doctor.py
```diff
@@ -65,7 +65,8 @@ def main(self, *, parser, args):
         if args.report_failed and fail != 0:
             fail_reports = generate_reports(categories=fail_category)
             for report_obj in fail_reports:
-                format_print(report_obj)
+                if report_obj is not None:
```
dirk-thomas, line 68: I don't see what this check is needed for. Can you please provide a reproducible example when this is needed.

nuclearsandwich:
> Since it technically needs ROS_DISTRO defined for it to work,

You mentioned it in an offline discussion but since the diff doesn't include any mention of the ROS_DISTRO variable a description of the failure mode without the ros_environment package (which is the canonical provider of the ROS_DISTRO environment variable) would be welcome in the PR description. To answer another offline question, there is no reason not to add a dependency on ros_environment and we've been recommending a build dependency on it for anyone whose build process wants to check the ROS_DISTRO, ROS_VERSION, or ROS_PYTHON_VERSION.

clalancette:
> You mentioned it in an offline discussion but since the diff doesn't include any mention of the ROS_DISTRO variable a description of the failure mode without the ros_environment package (which is the canonical provider of the ROS_DISTRO environment variable) would be welcome in the PR description.

Good point, will add that.

> To answer another offline question, there is no reason not to add a dependency on ros_environment and we've been recommending a build dependency on it for anyone whose build process wants to check the ROS_DISTRO, ROS_VERSION, or ROS_PYTHON_VERSION.

:+1:

ros2doctor/ros2doctor/command/doctor.py
clalancette: The easiest way to reproduce this is to build a workspace, source it, `unset ROS_DISTRO`, and then run `ros2 doctor -r`.  This somewhat artificially causes the `package.py` check to fail, and it to place a `None` value in the list of Report objects.

But actually, you are right.  Looking into it further, this actually highlighted a different bug in platform.py.  See 99d2d9b for a fix there.

dirk-thomas:
(Maybe update the title to mention the additional change happening in the patch.)

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11255/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6515/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9214/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11164/)

clalancette:

clalancette:
PR tests are flakey, so going ahead and merging this one.

PR #540, opened 2020-06-23
Update quoted comments in the test.

tfoote:
Followup to ros2/rcl_interfaces#103 that collided with #527 

@audrow This test is quite fragile and depends on very specific elements from a package that's only an indirect dependency. For robustness it might be worth specifically having controlled test messages that can be versioned with the code to avoid the fragility of relying on things like comments in other packages to remain static. I don't think that there's a significant benefit/purpose to test builtin_interfaces/test_msgs specifically vs any message you define with the expected structure. Otherwise this package will be required to be released in lockstep with any change to the message definitions in builtin_interfaces or test_msgs.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11267/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6525/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9225/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11174/)


audrow:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11258/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6518/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9217/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11167/)

jacobperron:
> I don't think that there's a significant benefit/purpose to test builtin_interfaces/test_msgs specifically vs any message you define with the expected structure.

I agree. I think `test_msgs` was originally used since it is our go-to interface package for use in unit tests in a lot of packages. We inherently have the issue that changes to messages in `test_msgs` may require changes in a lot of downstream packages (e.g. [`system_tests`](https://github.com/ros2/system_tests)), but it's a trade-off as we avoid maintaining separate interface packages for each package we want to test. For `ros2interface` in particular, it may make sense to maintain a separate interface package since the tests are especially sensitive towards the exact contents of interface definitions.

audrow:
> For ros2interface in particular, it may make sense to maintain a separate interface package since the tests are especially sensitive towards the exact contents of interface definitions.

I agree.  I made an issue #541 and assigned it to myself.

PR #544, opened 2020-06-30
add --keep-alive option to 'topic pub'

dirk-thomas:
The new option allows to make the currently hard coded sleep before exiting configurable. It also change the logic to apply that sleep for `times != 1` since even when publishing multiple messages it makes sense to wait after the last one to ensure it is published to the wire.

CI builds testing `ros2topic`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11339/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6576/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9302/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11236/)

### DIFF
--- a/ros2topic/ros2topic/verb/pub.py
+++ b/ros2topic/ros2topic/verb/pub.py
@@ -84,6 +84,10 @@ class PubVerb(VerbExtension):
             '-t', '--times', type=nonnegative_int, default=0,
             help='Publish this number of times and then exit')
         parser.add_argument(
+            '--keep-alive', metavar='N', type=positive_float, default=0.1,
+            help='Keep publishing node alive for N seconds after the last msg '
+                 '(default: 0.1)')
+        parser.add_argument(
             '-n', '--node-name',
             help='Name of the created publishing node')
         add_qos_arguments_to_argument_parser(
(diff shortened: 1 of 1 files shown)


ros2topic/ros2topic/verb/pub.py
```diff
@@ -83,6 +83,10 @@ def add_arguments(self, parser, cli_name):
         group.add_argument(
             '-t', '--times', type=nonnegative_int, default=0,
             help='Publish this number of times and then exit')
+        parser.add_argument(
+            '--keep-alive', metavar='N', type=positive_float, default=0.1,
+            help='Keep publishing node alive for N seconds after the last msg '
+                 '(default: 0.1)')
```
hidmic, line 89: @dirk-thomas meta, unrelated: I thought the parser took care of adding a string representation of the given `default` to the `help` string. But it does not. I think it should.

dirk-thomas: We are not configuring `argparse` for that. We are use the raw formatter class in `ros2cli`: https://github.com/ros2/ros2cli/blob/673d4fd22c17e59377f30005a3f82a43e425ecda/ros2cli/ros2cli/cli.py#L30

PR #547, opened 2020-06-30
Add interface definitions and update `ros2interface show` tests to use them

### LINKED ISSUE #541: [ros2interface] Tests rely on `builtin_interfaces`
The tests in `ros2interface` use interface types defined in `builtin_interfaces` and effectively checks some of the interface definition's file contents.  This makes it possible for `ros2interface`'s tests to fail when changes are made to `builtin_interfaces`.

In #540 it was proposed to add interface definitions into `ros2interface`.  Doing so will allow `ros2interface` to better control the messages used in its tests and remove `ros2interface`'s dependency on `builtin_interfaces`.

audrow:
Fixes #541.

### DIFF
--- a/ros2interface/test/test_cli.py
+++ b/ros2interface/test/test_cli.py
@@ -236,76 +236,64 @@ class TestROS2InterfaceCLI(unittest.TestCase):
 
     def test_show_message(self):
         with self.launch_interface_command(
-            arguments=['show', 'test_msgs/msg/BasicTypes']
+            arguments=['show', 'test_msgs/msg/ShortVariedMultiNested']
         ) as interface_command:
             assert interface_command.wait_for_shutdown(timeout=2)
         assert interface_command.exit_code == launch_testing.asserts.EXIT_OK
         assert launch_testing.tools.expect_output(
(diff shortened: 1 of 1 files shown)


dirk-thomas:
> I also fixed a small bug for how inline comments are handled for constants.

Please keep that unrelated change in a separate PR since the title of this PR doesn't give any hint that the patch also addresses that. Both changes might also be considered individually for backporting.

audrow:
Sounds good.  I'll split them up and open a separate PR for the bug fix, as well as consider them separately for backporting.

audrow:
I made the bug fix its own PR in #548.  After it's merged in, I'll update this PR to build off of those changes (since the updated tests assume that the bug is fixed).

audrow:
This set of PRs seems to pass CI without breaking downstream packages.  I'll run the rest of CI once these PRs are reviewed.
[Build Status](http://ci.ros2.org/job/ci_linux/11854/)

audrow:
Closed in favor of https://github.com/ros2/ros2cli/pull/579.

PR #558, opened 2020-07-31
improve the error message for invalid message types

### LINKED ISSUE #557: import_message error when message/module is not found is not informative
I have been going through the tutorials and made a mistake in the [single package interface](https://index.ros.org/doc/ros2/Tutorials/Single-Package-Define-And-Use-Interface) where I built the `more_interfaces` package in one terminal but did not run `. install/setup.bash` in my other terminal to update the workspace.

Running the publisher for the message in the correctly set up terminal works, but when I try to echo the topic in the one where I forgot to source the workspace, I get

```
ros2 topic echo /address_book 
  File "/opt/ros/foxy/bin/ros2", line 11, in <module>
    load_entry_point('ros2cli==0.9.7', 'console_scripts', 'ros2')()
  File "/opt/ros/foxy/lib/python3.8/site-packages/ros2cli/cli.py", line 67, in main
  File "/opt/ros/foxy/lib/python3.8/site-packages/ros2topic/command/topic.py", line 41, in main
  File "/opt/ros/foxy/lib/python3.8/site-packages/ros2topic/verb/echo.py", line 81, in main
    return main(args)
  File "/opt/ros/foxy/lib/python3.8/site-packages/ros2topic/verb/echo.py", line 95, in main
    message_type = get_msg_class(node, args.topic_name, include_hidden_topics=True)
  File "/opt/ros/foxy/lib/python3.8/site-packages/ros2topic/api/__init__.py", line 88, in get_msg_class
    msg_class = _get_msg_class(node, topic, include_hidden_topics)
  File "/opt/ros/foxy/lib/python3.8/site-packages/ros2topic/api/__init__.py", line 133, in _get_msg_class
    return get_message(message_type)
  File "/opt/ros/foxy/lib/python3.8/site-packages/rosidl_runtime_py/utilities.py", line 28, in get_message
    interface = import_message_from_namespaced_type(get_message_namespaced_type(identifier))
  File "/opt/ros/foxy/lib/python3.8/site-packages/rosidl_runtime_py/import_message.py", line 30, in import_message_from_namespaced_type
    module = importlib.import_module(
  File "/usr/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 961, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 973, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'more_interfaces'
```

The crash is [here](https://github.com/ros2/rosidl_runtime_py/blob/master/rosidl_runtime_py/import_message.py#L30).

In ros1, when a module/message type can't be found, a message like

```
ERROR: Cannot load message class for [my_package/MyMessage]. Are your messages built?
```

is shown. Rather than crashing and showing a backtrace, a message like this should be displayed, perhaps something like

```
ERROR: Cannot load message class for [my_package/MyMessage]. Check that the message is built and the workspace is sourced.
```

For reference, the ros1 code is [here](https://github.com/ros/ros_comm/blob/noetic-devel/tools/rostopic/src/rostopic/__init__.py#L1041). The error is printed in the main function after catching the exception, which is why the traceback is not printed.

Perhaps the solution to this is to raise a more informative error message here, and then add some specific handling code in the echo verb.

dirk-thomas:
Closes #557.

CI builds testing only `ros2topic` with Fast DDS:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11725/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6809/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9564/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11663/)

### DIFF
--- a/ros2topic/ros2topic/api/__init__.py
+++ b/ros2topic/ros2topic/api/__init__.py
@@ -141,7 +141,10 @@ def _get_msg_class(node, topic, include_hidden_topics):
         # Could not determine the type for the passed topic
         return None
 
-    return get_message(message_type)
+    try:
+        return get_message(message_type)
+    except (AttributeError, ModuleNotFoundError, ValueError):
+        raise RuntimeError("The message type '%s' is invalid" % message_type)
 
 
 class TopicMessagePrototypeCompleter:
--- a/ros2topic/ros2topic/verb/echo.py
+++ b/ros2topic/ros2topic/verb/echo.py
@@ -96,7 +96,10 @@ def main(args):
             message_type = get_msg_class(
                 node, args.topic_name, include_hidden_topics=True)
         else:
-            message_type = get_message(args.message_type)
+            try:
+                message_type = get_message(args.message_type)
+            except (AttributeError, ModuleNotFoundError, ValueError):
+                raise RuntimeError('The passed message type is invalid')
 
         if message_type is None:
             raise RuntimeError(
--- a/ros2topic/ros2topic/verb/pub.py
+++ b/ros2topic/ros2topic/verb/pub.py
@@ -129,7 +129,10 @@ def publisher(
     keep_alive: float,
 ) -> Optional[str]:
     """Initialize a node with a single publisher and run its publish loop (maybe only once)."""
-    msg_module = get_message(message_type)
+    try:
+        msg_module = get_message(message_type)
+    except (AttributeError, ModuleNotFoundError, ValueError):
+        raise RuntimeError('The passed message type is invalid')
     values_dictionary = yaml.safe_load(values)
     if not isinstance(values_dictionary, dict):
         return 'The passed value needs to be a dictionary in YAML format'


heuristicus:
Thanks for this. I think it would be good to be specific about which message type was not found - I often find myself echoing things without looking at the type first. Information about how to resolve the issue would be great too. In this case I was caught out by the need to source the environment after building which if I remember correctly wasn't strictly necessary in ros1.

dirk-thomas:
> I think it would be good to be specific about which message type was not found - I often find myself echoing things without looking at the type first.

If the argument was passed explicitly to the CLI we commonly don't repeat the information in error messages (e.g. https://github.com/ros2/ros2cli/blob/f03938e903902890043ac8f5d7b88aafc6b1f563/ros2service/ros2service/verb/call.py#L75, https://github.com/ros2/ros2cli/blob/f03938e903902890043ac8f5d7b88aafc6b1f563/ros2action/ros2action/verb/send_goal.py#L86).

> Information about how to resolve the issue would be great too. In this case I was caught out by the need to source the environment after building which if I remember correctly wasn't strictly necessary in ros1.

Since the reason why it fails isn't obvious - it could be a misspelling, it could missing environment for a newly created package - I don't think there is a good recommendation to include. Imo an incorrect recommendation can be quite misleading for users.

The situation is the same in ROS 1 - if you are looking at isolated builds for both ROS versions. For a non-isolated builds you often get away without (re-)source the environment since all packages share the same prefix path.

dirk-thomas:

dirk-thomas:

dirk-thomas:

dirk-thomas:
> This improves the case where the user explicitly gives the message type, but not the case where the user lets the tool infer the type. 

Good point. Added in c37fa77ad860fc045a588ce22532b9c5d14644ca.

dirk-thomas:

dirk-thomas:

dirk-thomas:

dirk-thomas:

dirk-thomas:

PR #559, opened 2020-07-31
check that passed type is actually a service

dirk-thomas:
Fixes ros2/rclpy#608.

If this passes the PR build I don't plan to run CI builds on ci.ros2.org since they won't test the effect of this patch.


dirk-thomas:

dirk-thomas:

PR #561, opened 2020-08-14
* catch additional exception in get_parameter_value()

travipross:
If `string_value` contains a leading `%` sign, this is interpreted as a YAML directive, which would throw a ParserError as expected. However, if an additional `%` exists elsewhere in `string_value`, then `PyYAML` throws a `ScannerError`.

Without catching this exception, `get_parameter_value()` will fail under these circumstances. With this fix, the intended behaviour is maintained.

Example: `ros2 param set /my_node my_param "%string"` succeeds, but  `ros2 param set /my_node my_param "%string%"` throws an uncaught `yaml.scanner.ScannerError`.

This commit should resolve the issue.

PR #562, opened 2020-08-20
Fix issue when setting parameter values containing percent symbols.

travipross:
When setting a parameter value with a leading percent symbol, `yaml.safe_load()` raises a `ParserError` which is caught, and the value is returned as a string. When a second percent symbol is added to the input value, `yaml.safe_load()` raises a `ScannerError` instead. This was previously uncaught.

Signed-off-by: Travis Prosser <travisprosser@gmail.com>

### DIFF
--- a/ros2service/ros2service/verb/call.py
+++ b/ros2service/ros2service/verb/call.py
@@ -69,10 +69,13 @@ def requester(service_type, service_name, values, period):
         module = importlib.import_module('.'.join(parts[:-1]))
         srv_name = parts[-1]
         srv_module = getattr(module, srv_name)
-        if not package_name or not srv_module:
-            raise ValueError()
-    except ValueError:
+    except (AttributeError, ModuleNotFoundError, ValueError):
         raise RuntimeError('The passed service type is invalid')
+    try:
+        srv_module.Request
+        srv_module.Response
+    except AttributeError:
+        raise RuntimeError('The passed type is not a service')
 
     values_dictionary = yaml.safe_load(values)
 


travipross:
@fujitatomoya Please continue the discussion regarding #561 here. In the previous PR the change was mistakenly applied to the wrong base branch.

Regarding your concern about the intended behaviour: ROS2 string parameters consider percentage symbols a valid character. `ros2param` should be capable of setting parameters with these values.

Because of the current approach of using `yaml.safe_load()` to guess the incoming datatype (despite the incoming text not actually being YAML), we can see some unexpected behaviour. The current approach catches any YAML parsing errors and assumes a string type was passed. This fix catches another `yaml` exception that gets raised in the case of an input value with a leading `%` *and* a successive `%` somewhere else in the string. 

This commit simply matches the intended behaviour of "try to parse with yaml, falling back to string if failing", while handling a previously uncaught Exception.

Example:
```python
>>> import yaml
>>> yaml.safe_load('test')
'test'
>>> yaml.safe_load('test%')
'test%'
>>> yaml.safe_load('%test')
yaml.parser.ParserError: expected '<document start>', but found '<stream end>'
  in "<unicode string>", line 1, column 6:
    %test
>>> yaml.safe_load('%test%')
yaml.scanner.ScannerError: while scanning a directive
  in "<unicode string>", line 1, column 1:
    %test%
```

travipross:
Regarding CI test failures:

Everything works locally on my Ubuntu Bionic machine with ROS2 Eloquent. Colcon can successfully build and all tests pass.

Is there something about the CI environment that has changed which breaks functionality in this branch?

clalancette:
> Everything works locally on my Ubuntu Bionic machine with ROS2 Eloquent. Colcon can successfully build and all tests pass.
> 
> Is there something about the CI environment that has changed which breaks functionality in this branch?

I'm not sure about Eloquent at the moment.  In any case, our typical workflow is to fix the problem on the master branch, and then backport it to released distros as required.  With that in mind, I'm going to retarget this PR to the master branch.

clalancette:
(this will obviously need a rebase now)

PR #564, opened 2020-09-06
Added dependency to python3-argcomplete to ros2cli

ymollard:
Following https://github.com/ros2/ros2_documentation/pull/841#issuecomment-682203981 here's a proposal so that all ros2cli commands no longer require a manual install of python3-argcomplete

TODO: to be updated in https://github.com/ros2/ros2_documentation also

### DIFF
--- a/ros2cli/package.xml
+++ b/ros2cli/package.xml
@@ -9,11 +9,11 @@
   <maintainer email="dthomas@osrfoundation.org">Dirk Thomas</maintainer>
   <license>Apache License 2.0</license>
 
+  <exec_depend>python3-argcomplete</exec_depend>
   <exec_depend>python3-importlib-metadata</exec_depend>
   <exec_depend>python3-netifaces</exec_depend>
   <exec_depend>python3-packaging</exec_depend>
   <exec_depend>python3-pkg-resources</exec_depend>
-  <exec_depend>python3-argcomplete</exec_depend>
   <exec_depend>rclpy</exec_depend>
 
   <test_depend>ament_copyright</test_depend>
--- a/ros2cli/package.xml
+++ b/ros2cli/package.xml
@@ -13,6 +13,7 @@
   <exec_depend>python3-netifaces</exec_depend>
   <exec_depend>python3-packaging</exec_depend>
   <exec_depend>python3-pkg-resources</exec_depend>
+  <exec_depend>python3-argcomplete</exec_depend>
   <exec_depend>rclpy</exec_depend>
 
   <test_depend>ament_copyright</test_depend>


ivanpauno (commented):
Other packages that need the dependency:
- [ament_index_python](https://github.com/ament/ament_index/blob/c9778d8f97e95e8ea344b49442c20fad43d77cd7/ament_index_python/ament_index_python/cli.py#L36)
- [ros2launch](https://github.com/ros2/launch_ros/blob/38407baecc035389794c67247bbf620aa050c320/ros2launch/ros2launch/command/launch.py#L20)
- [sros2](https://github.com/ros2/sros2/blob/5d57584af3db0a5542ccfbf3d9ee43740ecbcca4/sros2/sros2/verb/create_key.py#L16)
- [ros2cli](https://github.com/ros2/ros2cli/blob/591a8aff10975c12d7070a4e0007cb1e147de183/ros2cli/ros2cli/cli.py#L47) (dependency being added here)
- [ros2run](https://github.com/ros2/ros2cli/blob/591a8aff10975c12d7070a4e0007cb1e147de183/ros2run/ros2run/command/run.py#L37)

I'm not sure if we need to all the dependency to all of them, as it's an optional dependency.
Most of the packages in the above list depend on *ros2cli*, so they will get the dependency transitively (except `ament_index_python`).
@clalancette thoughts?

Note: we will probably need to modify the CI scripts before trying this, to make sure the `pip` installed version of argcomplete is not available.

ros2cli/package.xml
```diff
@@ -13,6 +13,7 @@
   <exec_depend>python3-netifaces</exec_depend>
   <exec_depend>python3-packaging</exec_depend>
   <exec_depend>python3-pkg-resources</exec_depend>
+  <exec_depend>python3-argcomplete</exec_depend>
```
ivanpauno, line 16: nit: alpha order

clalancette:
> I'm not sure if we need to all the dependency to all of them, as it's an optional dependency.
> Most of the packages in the above list depend on _ros2cli_, so they will get the dependency transitively (except `ament_index_python`).
> @clalancette thoughts?

I'm usually in favor of 'include-what-you-use'; you never know if a downstream transitive dependency gets changed or removed.  That means for this particular PR, I would add an `exec_depend` on `python3-argcomplete` to both the `ros2cli` and the `ros2run` packages, as they are the ones that directly depend on argcomplete.  I'd also like to see the dependency added to the others in your list, but those would be separate PRs.

dirk-thomas:
> Other packages that need the dependency:

Why should packages which only contribute extension points to `ros2cli` also declare the dependency? They explicitly only conditionally use `argcomplete`.

`ros2cli` determines if argcomplete is being used since it also provides the necessary environment hook.

clalancette:
> Why should packages which only contribute extension points to `ros2cli` also declare the dependency? They explicitly only conditionally use `argcomplete`.

Sure, but generally anything that does an `import` of argcomplete should arguably have the dependency.  For instance, the standalone script https://github.com/ament/ament_index/blob/c9778d8f97e95e8ea344b49442c20fad43d77cd7/ament_index_python/ament_index_python/cli.py#L36 imports it, so it should have the dependency.  Other places that @ivanpauno pointed out are similar.

And while it is conditional (I think in all cases?), I think that most users want it installed by default.  So it makes sense to optimize for those users.  Is there a way to declare a dependency as "Recommends" in the Debian packages?  That way it will get installed by default, but users who don't want it can uninstall it by hand.

dirk-thomas:
> the standalone script https://github.com/ament/ament_index/blob/c9778d8f97e95e8ea344b49442c20fad43d77cd7/ament_index_python/ament_index_python/cli.py#L36 imports it, so it should have the dependency.

Totally agreeing to this case. It is a separate executable and its own environment hook :+1:

>  Other places that @ivanpauno pointed out are similar.

I disagree on that because the do *not* have neither their own executables and nor environment hook.

clalancette (approved):
Regardless of the other conversation going on here, I think we can go ahead with *this* change, so I'll approve.  @dirk-thomas Do you agree?

clalancette:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12336/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7305/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10049/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12245/)

dirk-thomas (approved):

clalancette:

Issue #563, opened 2020-08-21
Autocompletion for message contents broken since foxy


mbuijs:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04 / docker
- Installation type:
  - Apt packages / prebuilt docker image
- Version or commit hash:
  - ros-foxy-ros2topic: 0.9.7-1focal.20200804.003710
  - For the docker images: 
```
$ docker image ls --no-trunc ros
REPOSITORY          TAG                 IMAGE ID                                                                  CREATED             SIZE
ros                 rolling             sha256:ab3c32fdab44aba3069366d90df5f212c1fb233fcd59bd04a16194f7e86aeef2   37 hours ago        722MB
ros                 foxy                sha256:896c8acbfce8030f2550980f38c11d4a33e1e52500ebe7b4ddc3a6a95f85bc05   37 hours ago        722MB
ros                 eloquent            sha256:e5483efb0d9505e6d4d84466d3a7bb78f48e625661abe2552c77eceddd3f8484   37 hours ago        829MB
```
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - rclpy (?)

#### Steps to reproduce issue
Make sure the setup.bash is sources. Then try to publish a message using `ros2 topic pub`:
```bash
$ . /opt/ros/${ROS_DISTRO}/setup.bash
$ ros2 topic pub /cli_demo geometry_msgs/msg/Pose "po<tab>
```

#### Expected behavior

A complete template of the contents of the specified message appears:
```bash
$ ros2 topic pub /cli_demo geometry_msgs/msg/Pose "position:
  x: 0.0
  y: 0.0
  z: 0.0
orientation:
  x: 0.0
  y: 0.0
  z: 0.0
  w: 1.0"
```
This is what happens when using `eloquent` (verified using the mentioned docker image).
In `foxy` this does not happen, neither in docker, nor in a local terminal where the installed apt packages are available.
Same applies for the `rolling` docker image.

#### Actual behavior

No completion suggestion appear. When done without the double quote, the following happens:
```bash
$ ros2 topic pub /cli_demo geometry_msgs/msg/Pose position:\
\ \ x:\ 0.0\
\ \ y:\ 0.0\
\ \ z:\ 0.0\
orientation:\
\ \ x:\ 0.0\
\ \ y:\ 0.0\
\ \ z:\ 0.0\
\ \ w:\ 1.0\ 
```

[...]

PR #570, opened 2020-10-20
bugfix for #563

daisukes:
Use argcomplete's split_line to deal with incomplete quote states.

Signed-off-by: Daisuke Sato <daisukes@cmu.edu>

### DIFF
--- a/ros2cli/ros2cli/command/__init__.py
+++ b/ros2cli/ros2cli/command/__init__.py
@@ -16,7 +16,6 @@
 import argparse
 import inspect
 import os
-import shlex
 import types
 
 from ros2cli.entry_points import get_entry_points
@@ -201,7 +200,9 @@ def add_subparsers_on_demand(
         args = argv
         # for completion use the arguments provided by the argcomplete env var
         if _is_completion_requested():
-            args = shlex.split(os.environ['COMP_LINE'])[1:]
+            from argcomplete import split_line
+            _, _, _, comp_words, _ = split_line(os.environ['COMP_LINE'])
+            args = comp_words[1:]
         try:
             known_args, _ = root_parser.parse_known_args(args=args)
         except SystemExit:


ros2cli/ros2cli/command/__init__.py
```diff
@@ -201,7 +200,9 @@ def add_subparsers_on_demand(
         args = argv
         # for completion use the arguments provided by the argcomplete env var
         if _is_completion_requested():
-            args = shlex.split(os.environ['COMP_LINE'])[1:]
+            from argcomplete import split_line
```
clalancette, line 203: The big reason not to do this is that `argcomplete` is currently an optional dependency of `ros2cli`, and this makes it required.  Is there another way we can accomplish the same thing?

clalancette (changes requested):

ros2cli/ros2cli/command/__init__.py
daisukes: I was wondering if there is a case that we cannot load argcomplete and _is_completion_requested True?

clalancette: I'm honestly not sure.  It looks like the definition of `_is_comletion_requested()` is:

```
def _is_completion_requested():
    return os.environ.get('_ARGCOMPLETE') == '1'
```

I don't know what is expected to set the `_ARGCOMPLETE` environment variable, so I'm not sure what situations get us here.

daisukes: `_ARGCOMPLETE` appears [here](https://github.com/kislyuk/argcomplete/blob/6d33c8e321ffa388c879fec56a80add0181f56b8/argcomplete/shell_integration.py#L44) in argcomplete registration script.

This is called from a ros2cli script, if the script is available meaning argcomplete available I believe,  
https://github.com/ros2/ros2cli/blob/df55a4ed42cdaca1fa103bb1b9807ce2c239055f/ros2cli/completion/ros2-argcomplete.bash#L15-L19

which is invoked when you call setup script
https://github.com/ros2/ros2cli/blob/e5f67d0e0362462644a5530094fac43af84ca938/ros2cli/colcon.pkg#L1-L6

clalancette: Thanks for pointing it out.  This does look OK then.

clalancette (commented):
This looks reasonable enough to me.  @claireyywang @mabelzhang can one of you review this, and maybe test out to make sure things still work both with argcomplete installed and with it not installed?

clalancette (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13609/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8491/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11325/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13670/)

clalancette:
@daisukes I'd love to get this in.  However, it looks like it is causing some failures in the `test_list_` tests in CI.  Would you mind taking a look?

daisukes:
@clalancette, sorry for the late response. I'm going to look into it!

clalancette:
> @clalancette, sorry for the late response. I'm going to look into it!

Thanks, appreciated!

daisukes:
I've checked the log of the CI and found the error. 
```
FAIL: test_cli.TestROS2TopicCLI.test_topic_endpoint_info_verbose[rmw_fastrtps_cpp]
```
However, this can happen with the previous commit 56db2ba9fcc85ead292b0c3403cf27dba41854e2 of mine.
Also, I tested my patch with the current master, and the patch does not introduce any new error. 

The error is caused by the unexpected output of the `ros2 topic info -v /chatter` command.
It seems like this commit fixed the error.
https://github.com/ros2/ros2cli/commit/2b6930e27713e46c70b95997c511a89bf57cc624

claireyywang:
CI 
Linux [Build Status](https://ci.ros2.org/job/ci_linux/13650/) 
Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/8530/)
macOS [Build Status](https://ci.ros2.org/job/ci_osx/11365/)
Windows [Build Status](https://ci.ros2.org/job/ci_windows/13715/)

clalancette:
> The error is caused by the unexpected output of the `ros2 topic info -v /chatter` command.
> It seems like this commit fixed the error.

Ah, good point.  I've just rebased this branch onto master, and here is another run at CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13652/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8532/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11368/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13717/)

clalancette:
All right, CI is happy.  The PR job is failing, but that doesn't seem to be unique to this job.  @daisukes Thanks for the contribution, and sorry this took so long to get in.  Merging.

Issue #569, opened 2020-10-10
param list -t for consistency


JuliusSustarevas:
#96  Feature request
ros2 topic list
and 
ros2 service list
have the -t parameter to show types. But ros2 param list does not. It would be convinient and consistent if it did. 

#### Feature description
"ros2 param list -t" could return param names followed by types

clalancette:
> "ros2 param list -t" could return param names followed by types

While I like the idea of the feature, I think we may want to use a different flag than `-t`.  The `-t` flag implicitly means print the ROS *message* type, which is not what you are asking for.  Instead, you are asking for the internal stored type of the parameters.  I'll just throw out `--param-type` as a flag idea for now, but maybe others disagree.

In any case, while I think this is a great idea, I don't think we're going to have time to work on it right now.  But if you'd like to contribute a change to add it, we'd be happy to review it.

Thanks for the report.

JuliusSustarevas:
Well I personally always wanted params to be ROS Message type which gives a certain nice consistency. You could set some param as a Pose message, set your deadzone speed as a Twist Message etc. But thats a separate feature request i guess

fujitatomoya:
@clalancette 

since this is tagged with `help wanted`, i did come up with quick candidate. could you take a look at https://github.com/ros2/ros2cli/pull/572?

fujitatomoya:
@clalancette 

friengly ping, i think we can close this.

clalancette:
Closing since #572 is now merged.

PR #572, opened 2020-10-28
add "--param-type" option to ros2param list.

fujitatomoya:
address https://github.com/ros2/ros2cli/issues/569

Signed-off-by: Tomoya.Fujita <Tomoya.Fujita@sony.com>

### DIFF
--- a/ros2param/ros2param/verb/list.py
+++ b/ros2param/ros2param/verb/list.py
@@ -22,6 +22,8 @@ from ros2cli.node.strategy import NodeStrategy
 from ros2node.api import get_absolute_node_name
 from ros2node.api import get_node_names
 from ros2node.api import NodeNameCompleter
+from ros2param.api import call_describe_parameters
+from ros2param.api import get_parameter_type_string
 from ros2param.verb import VerbExtension
 from ros2service.api import get_service_names
 
@@ -41,6 +43,9 @@ class ListVerb(VerbExtension):
         parser.add_argument(
             '--param-prefixes', nargs='+', default=[],
             help='Only list parameters with the provided prefixes')
+        parser.add_argument(
+            '--param-type', action='store_true',
+            help='Print parameter types with parameter names')
 
     def main(self, *, args):  # noqa: D102
         with NodeStrategy(args) as node:
@@ -97,7 +102,14 @@ class ListVerb(VerbExtension):
                         print(f'{node_name.full_name}:')
                     response = future.result()
                     for name in sorted(response.result.names):
-                        print(f'  {name}')
+                        if args.param_type is True:
+                            resp = call_describe_parameters(
+                                node=node, node_name=node_name.full_name,
+                                parameter_names={name})
+                            param_type_str = get_parameter_type_string(resp.descriptors[0].type)
+                            print(f'  {name} [{param_type_str}]')
+                        else:
+                            print(f'  {name}')
                 else:
                     e = future.exception()
                     print(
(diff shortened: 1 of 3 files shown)


ros2param/ros2param/verb/list.py
```diff
@@ -97,7 +102,14 @@ def main(self, *, args):  # noqa: D102
                         print(f'{node_name.full_name}:')
                     response = future.result()
                     for name in sorted(response.result.names):
-                        print(f'  {name}')
+                        if args.param_type is True:
+                            resp = call_describe_parameters(
```
fujitatomoya, line 106: Or maybe we can call this before with parameter_names=None, then search the parameter name which matches with 'name' from list service response.

clalancette: Yeah, I think that would be better.  Doing a separate service call for every parameter is going to be expensive and possibly slow (particularly if the node is on a different machine on the network).

ros2param/ros2param/verb/list.py
```diff
@@ -97,7 +102,14 @@ def main(self, *, args):  # noqa: D102
...
+                                node=node, node_name=node_name.full_name,
+                                parameter_names={name})
+                            param_type_str = get_parameter_type_string(resp.descriptors[0].type)
+                            print(f'  {name} [{param_type_str}]')
```
clalancette, line 110: Personally, I would change this to something a little more descriptive:

```
print(f'  {name} (type: {param_type_str})')
```

clalancette (changes requested):
A few things to change in here, but I like the idea.

ros2param/ros2param/verb/list.py
fujitatomoya: Done

fujitatomoya:
@clalancette 

could you check once again?

clalancette:
@fujitatomoya FYI, I rebased this code and did a few more fixups to make the type lookup even faster.

clalancette (commented):
One thing that would be really nice to do here would be to add some "cli" tests for the `list` verb, like we have for `dump` in https://github.com/ros2/ros2cli/blob/master/ros2param/test/test_verb_dump.py .  I don't think it would be too hard to add, and would give us a bit more verification of this work.  @fujitatomoya do you have some time to look into that?

clalancette:
> One thing that would be really nice to do here would be to add some "cli" tests for the `list` verb, like we have for `dump` in https://github.com/ros2/ros2cli/blob/master/ros2param/test/test_verb_dump.py . I don't think it would be too hard to add, and would give us a bit more verification of this work. @fujitatomoya do you have some time to look into that?

Never mind.  I missed that we had https://github.com/ros2/ros2cli/pull/389 already open, and apparently it is harder than I think since we never solved it there.

I'm going to approve this and run CI on it.

clalancette (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12910/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7864/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10627/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12859/)

clalancette:
@claireyywang @mabelzhang I'd appreciate a second look here since I did some of the modifications myself.

claireyywang (approved):
LGTM

clalancette:
All right, we have approvals, green CI.  I'm going to go ahead and merge, thanks @fujitatomoya !

fujitatomoya:
@clalancette 

> I rebased this code and did a few more fixups to make the type lookup even faster.

thanks 👍 this is better!

PR #575, opened 2020-11-13
Add filter option to ros2topic 

jacobperron:
A new '--filter' option makes it easy to echo only part of a message.
For example, to only echo the position part of an odometry message:

    ros2 topic echo /odom --filter pose/pose/position

The filter option takes one value that must not be empty (or only forward slashes '/').

--- 

I found that this feature was a bit easier to add after refactoring the code a little bit. I did the refactor in a separate commit to make the review easier (ae07bd5):

* Move subscription functions into class for easier reuse arguments.
* Prefix functions with an underscore to indicate they are private.

### DIFF
--- a/ros2topic/ros2topic/verb/echo.py
+++ b/ros2topic/ros2topic/verb/echo.py
@@ -58,7 +58,12 @@ class EchoVerb(VerbExtension):
             help='Output all recursive fields separated by commas (e.g. for '
                  'plotting)')
         parser.add_argument(
-            '--filter', type=str, default=None, help='Filter to echo a part of a message.')
+            '--field', type=str, default=None,
+            help='Echo a selected field of a message. '
+                 "Use '.' to select sub-fields. "
+                 'For example, to echo the position field of a nav_msgs/msg/Odometry message: '
+                 "'ros2 topic echo /odom --field pose.pose.position'",
+        )
         parser.add_argument(
             '--full-length', '-f', action='store_true',
             help='Output all elements for arrays, bytes, and string with a '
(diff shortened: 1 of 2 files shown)


jacobperron:

fujitatomoya (approved):
lgtm, I think that this is really useful during debug.

mjeronimo:
Looks good.

One comment on usability: When I first tried the feature, I attempted to use a '.' for the field separator. It took me a minute to figure out that the '/' character was used instead. The '.' is more natural to me (and perhaps other C++ and Python developers); kind of a simple dotting into a structure rather that a more full-featured path-oriented query (where one might expect to use a '/').

mjeronimo (approved):

jacobperron:
> One comment on usability: When I first tried the feature, I attempted to use a '.' for the field separator (i.e., I used: "--filter pose.pose.position.x"). It took me a minute to figure out that the '/' character was used instead. The '.' is more natural to me (and perhaps other C++ and Python developers); kind of a simple dotting into a structure rather that a more full-featured path-oriented query (where one might expect to use a '/').

Makes sense. I think we could use any separator as long as it is going to clash with possible names for message fields. I went with `/` only because it seemed like a natural extension of the topic and what was used in ROS 1's `rostopic echo`. Considering the filter is not actually operating on topic names, I'm open to changing the delimiter (or perhaps supporting multiple delimiters).

I'm also not very attached to the option name, `--filter`. Open to other suggestions, if any.

jacobperron:
The test failures look unrelated; they also happen on the latest [dev job](http://build.ros2.org/view/Rdev/job/Rdev__ros2cli__ubuntu_focal_amd64/21/). It's not clear to me what caused them to start failing. I'm not able to reproduce locally... perhaps they are flaky.

jacobperron:

jacobperron:
After some offline discussion, I've decided to change the option name to `--field`, as it seems more intuitive. I've also changed the delimiter to `.`, also because it seems more intuitive as we are accessing members of a structure.

I've also improved the help message, with an example, and added another test for accessing a nested field.

@mjeronimo PTAL at the most recent changes in a107892

mjeronimo (approved):

jacobperron:
Aha, and CI was in our favor this time :sparkles: 

Here's CI for other platforms:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13115/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8047/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10825/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13092/)

PR #548, opened 2020-06-30
[ros2interface] Fixes inline comments for constants

audrow:
Previously, inline comments on constants were not able to be removed; now they adopt the same display behavior as other comments.

Issue #577, opened 2020-11-18
test_echo_pub.TestROS2TopicEchoPub failing (Rdev and RPr)


jacobperron:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- Installation type:
  - binaries
- Version or commit hash:
  - 0.10.0
- DDS implementation:
  - Fast-DDS

#### Steps to reproduce issue

I haven't been able to reproduce any of the test failures locally.

#### Expected behavior

All tests pass.

#### Actual behavior

test_echo_pub.TestROS2TopicEchoPub is failing in the latest [dev job](http://build.ros2.org/view/Rdev/job/Rdev__ros2cli__ubuntu_focal_amd64/21) and RPr jobs ([for example](http://build.ros2.org/job/Rpr__ros2cli__ubuntu_focal_amd64/121/testReport/ros2topic.ros2topic.test/test_echo_pub/test_echo_pub)).

#### Additional information

I've seen [test_api.TestROS2DoctorAPI](http://build.ros2.org/job/Rpr__ros2cli__ubuntu_focal_amd64/121/testReport/ros2doctor.ros2doctor.test/test_api/test_api/) also fail sometimes during a RPr job. Seems to also be related to an issue connecting with a test publisher.

fujitatomoya:
I cannot reproduce this issue locally either. So far I am not sure what is wrong with test code, any clues? maybe something we can do is to wait for number of publication or subscription before CLI is issued?

jacobperron:
The past couple dev jobs haven't showed this error, but it looks like a recent PR job failure could be related: https://build.ros2.org/job/Rpr__ros2cli__ubuntu_focal_amd64/128/testReport/

Looks like a rare flake.

clalancette:
This doesn't show up in any of the recent Rdev jobs.  I'm going to close this out, but we can reopen if this appears again.

PR #579, opened 2020-12-17
Remove ros2interface test dependencies on builtin interface

### LINKED ISSUE #541: [ros2interface] Tests rely on `builtin_interfaces`

audrow:
Fixes #541 and is an alternative to the PRs associated with #547.

I think this is an important change because, as mentioned in https://github.com/ros2/system_tests/pull/442#issuecomment-671632871, testing against these interface definitions allows us to cover behavior of `ros2 interface show` that is currently not covered, such as the bug fixed by https://github.com/ros2/ros2cli/pull/548. This PR also removes a `ros2interface` test dependency on `builtin_interfaces` (#541). 

I've opted for this approach rather than the one used in #547 to avoid slowing down the test time and for simplicity. As @jacobperron mentions in  https://github.com/ros2/test_interface_files/pull/11#issuecomment-697887296 that adding these interfaces to may increase the time required for all of the tests to run. It also is much more complicated to add messages to `test_msgs` as it requires PRs in four repositories.


ros2cli_test_msgs/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
+cmake_minimum_required(VERSION 3.5)
+project(ros2cli_test_msgs)
```
jacobperron, line 2: nitpick: I prefer the suffix `_interfaces` in the package name, since the package contains more than just messages (ie. also actions and services).

ros2cli_test_msgs/package.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>ros2cli_test_msgs</name>
+  <version>1.0.1</version>
+  <description>A package containing interface definitions for testing ros2cli.</description>
+  <maintainer email="audrow@openrobotics.org">Audrow Nash</maintainer>
```
jacobperron, line 7: I guess we should also add maintainer tags for other maintainers of the ros2cli repository.

ros2cli_test_msgs/CMakeLists.txt
audrow: Sounds good, I think interfaces is better, too - I was following the example of `test_msgs`, which also has services and actions.

jacobperron: Yeah, earlier packages in the development of ROS 2 follow the ROS 1 naming convention with the `_msgs` suffix.

audrow:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13300/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8233/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11033/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13343/)

ros2cli/package.xml
```diff
@@ -24,6 +24,7 @@
   <test_depend>ament_pep257</test_depend>
   <test_depend>ament_xmllint</test_depend>
   <test_depend>python3-pytest</test_depend>
+  <test_depend>ros2cli_test_interfaces</test_depend>
```
jacobperron, line 27: It doesn't look like this test depend is necessary.

jacobperron (commented):
LGTM (one minor comment)

We should also run CI and test the new package `ros2cli_test_interfaces`.

jacobperron:
I believe the PR job failure is the same as https://github.com/ros2/ros2cli/issues/577. It seems safe to ignore, since this change is only touching `ros2interface`.

jacobperron (approved):

audrow:
Here is a new CI. I am testing on `ros2interface` and `ros2cli_test_interfaces`.
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13324/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8254/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11055/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13371/)

audrow:
I reran windows. It seems to pass. I'll rebase on master and then rerun CI.
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13514/)

jacobperron:
> I reran windows. It seems to pass. I'll rebase on master and then rerun CI.

I don't think rebasing is necessary we're using a branch (instead of a repos file) for CI.

audrow:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13456/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8367/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11180/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13520/)

Issue #515, opened 2020-05-18
[ros2 topic delay] use_sim_time


doisyg:
`ros2 topic delay` seems to use the system clock instead of the topic `/clock` when using a simulation (and I guess playing a bag). 
```
$ ros2 topic delay odom
average delay: 1589796651.954
	min: 1589796651.949s max: 1589796651.958s std dev: 0.00260s window: 28

```
How to set `use_sim_time` to true for `ros2 topic` ? Or what should be the appropriate way to work with a simulation ?

hidmic:
Thanks for reporting! It is true there's currently no way to change `use_sim_time` value for `ros2cli` node. Off the top of my head, [`DirectNode`-specific arguments](https://github.com/ros2/ros2cli/blob/b622109d96e16506f1cd31cc9a4664c7df7332ec/ros2cli/ros2cli/node/direct.py#L84) could include a `-s` / `--use-sim-time` flag.

hyunseok-yang:
Hi, anyone solved this question?

clalancette:
> Hi, anyone solved this question?

It is still an open feature request, as far as I know.  If you'd like to jump in and try to implement it, that would be welcome.

fujitatomoya:
@doisyg @hyunseok-yang 

could you kindly try https://github.com/ros2/ros2cli/pull/581 to see if you can have what you want? I did make sure that ros2cli node is constructed with `use_sim_time` is true if option is enabled. it would be really appreciated.

hyunseok-yang:
@fujitatomoya Thanks. I've forgot to check your guys comment :(
I will have try the #581 code!

PR #581, opened 2021-01-07
add option to support use_sim_time.

fujitatomoya:
address https://github.com/ros2/ros2cli/issues/515

### DIFF
--- a/ros2cli/test/test_direct.py
+++ b/ros2cli/test/test_direct.py
@@ -1,0 +1,63 @@
+# Copyright 2020 Sony Corporation.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+
+import pytest
+
+from ros2cli.node.direct import add_arguments as add_direct_node_arguments
+from ros2cli.node.direct import DEFAULT_TIMEOUT
+from ros2cli.node.direct import DirectNode
+
+TEST_NODE_NAME = 'test_node'
+
+
+@pytest.fixture(scope='function')
+def test_arguments_parser():
+    parser = argparse.ArgumentParser()
+    add_direct_node_arguments(parser)
+    return parser
+
+
+def test_default_arguments(test_arguments_parser):
+    args = test_arguments_parser.parse_args()
+    assert not args.use_sim_time
+    assert DEFAULT_TIMEOUT == args.spin_time
+
+
+def test_use_sim_time_arguments(test_arguments_parser):
+    args = test_arguments_parser.parse_args(['--use-sim-time'])
+    assert args.use_sim_time
+    assert DEFAULT_TIMEOUT == args.spin_time
+
+
+def test_spin_time_arguments(test_arguments_parser):
+    args = test_arguments_parser.parse_args(['--spin-time', '10.0'])
+    assert not args.use_sim_time
+    assert 10.0 == args.spin_time
+
+
+def test_use_sim_time_parameter():
+    with DirectNode(args=[], node_name=TEST_NODE_NAME) as direct_node:
+        assert not direct_node.node.get_parameter('use_sim_time').value
+
+    args = argparse.Namespace(use_sim_time=False)
+    with DirectNode(args, node_name=TEST_NODE_NAME) as direct_node:
+        assert not direct_node.node.get_parameter('use_sim_time').value
+
+    # TODO(fujitatomoya): enable this test once /clock callback thread is insulated.
+    #                     see https://github.com/ros2/rclcpp/issues/1542
+    # args = argparse.Namespace(use_sim_time=True)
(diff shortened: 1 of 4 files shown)


fujitatomoya:
CC: @clalancette @hidmic

fujitatomoya:
@hidmic 

friendly ping, could you share what you think?

hidmic (approved):
Seems reasonable to me. A test would be ideal.

audrow (commented):
Looks good to me, too. I also would like to see a test for this.

fujitatomoya:
@hidmic @audrow thanks for the review, will try to add test on this.

fujitatomoya:
@hidmic @audrow friendly ping,

ros2cli/test/test_direct.py
```diff
@@ -0,0 +1,63 @@
...
+
+def test_use_sim_time_parameter():
+    with DirectNode(args=[], node_name=TEST_NODE_NAME) as direct_node:
+        assert not direct_node.node.get_parameter('use_sim_time').value
+
+    args = argparse.Namespace(use_sim_time=False)
+    with DirectNode(args, node_name=TEST_NODE_NAME) as direct_node:
+        assert not direct_node.node.get_parameter('use_sim_time').value
+
+    # TODO(fujitatomoya): enable this test once /clock callback thread is insulated.
```
fujitatomoya, line 59: i did try using threading to run DirectNode in another thread, but the thread cannot be finished anyway. instead of fighting here, it would be nice to address https://github.com/ros2/rclcpp/issues/1542 for rclpy, so that we can have a control for `Node`.

Any idea?

fujitatomoya:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13674/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8553/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11392/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13740/)

ros2cli/test/test_direct.py
audrow: Do I understand the problem correctly?: The commented test does not work because the node does not shutdown since it is not getting the simulation clock signal. If the signal was insulated, then the inner logic of the node would allow it to shutdown after the `with` context for the node has been left in the test, which destroys the node.

The suggestion you've linked to might be a good approach. I know that there are active discussions on how to make clocks more consistent in ROS2. My impression is that more discussion needs to occur before decisions are made and implementation can begin. This may mean that the test may be blocked for some time. 

Perhaps we could proceed with another solution that works for now (e.g., kill the node's thread after you've confirmed the value), or leave the TODO to enable the test, as you have done.

audrow:
It probably also makes sense to run CI tests for `ros2topic`, since you are modifying `ros2topic/ros2topic/verb/delay.py`.

ros2cli/test/test_direct.py
fujitatomoya: > Do I understand the problem correctly?

right, node just waits on rmw_wait forever to subscribe /clock topic. test just keeps running and cannot be finished.

> My impression is that more discussion needs to occur before decisions are made and implementation can begin. This may mean that the test may be blocked for some time.

totally agree,

> Perhaps we could proceed with another solution that works for now (e.g., kill the node's thread after you've confirmed the value), or leave the TODO to enable the test, as you have done.

i think we cannot simply kill the threads with python threading, if you know the idea that would be really help. (i really do not want to introduce multi-processing on this test) probably we could send the SIGINT signal to self-pid and join thread?

fujitatomoya:
> It probably also makes sense to run CI tests for ros2topic, since you are modifying ros2topic/ros2topic/verb/delay.py.

yeah, right. i will do that later.

ros2cli/test/test_direct.py
fujitatomoya: none of them are working to finish the pytest of test_use_sim_time_true_parameter. DirectNode expects timer expire but it just waiting on that event via syscall futex. i'd like to leave this TODO atm.

```
def test_use_sim_time_true_parameter():
    # TODO(fujitatomoya): enable this test once /clock callback thread is insulated.
    #                     see https://github.com/ros2/rclcpp/issues/1542
    direct_node = None

    def spin_direct_node():
        nonlocal direct_node
        args = argparse.Namespace(use_sim_time=True)
        direct_node = DirectNode(args, node_name=TEST_NODE_NAME)

    t = threading.Thread(target=spin_direct_node)
    t.start()
    time.sleep(0.1)
    assert direct_node.node.get_parameter('use_sim_time').value
    #pytest.exit("exit anyway", not direct_node.node.get_parameter('use_sim_time').value)
    #direct_node.node.destroy_timer(direct_node.timer)
    #os.kill(os.getpid(), signal.SIGINT)
    t.join()
```

fujitatomoya:
CI with ros2cli and ros2topic

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13680/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8559/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11397/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13747/)

ros2cli/test/test_direct.py
audrow: > i'd like to leave this TODO atm.

Sounds good to me.

audrow (approved):

audrow:
@fujitatomoya, I think you have permissions to squash and merge. If not, let me know. Thanks for the PR!

PR #583, opened 2021-01-12
Revert "Skip CLI tests on Windows until we resolve the blocking/hanging isuse. (#489)"

hidmic:
Hopefully https://github.com/ros2/launch/pull/476 fixed this :crossed_fingers: 

CI up to `ros2topic`, `ros2action`, `ros2interface`, `ros2lifecycle`, `ros2node`, `ros2pkg`, and `ros2service` (repeated 5 times):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13413/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8333/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11138/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13472/)


ros2interface/test/test_cli.py
```diff
@@ -15,7 +15,6 @@
 import contextlib
 import itertools
 import re
-import sys
```
clalancette, line 18: You still need `sys` for elsewhere in the file.

hidmic: Oops, done in https://github.com/ros2/ros2cli/pull/583/commits/4018dec5fefa467a878414b74083ed890cbd8811.

hidmic:
Rebased to solve failing DCO check.

hidmic:
Another round of CI up to `ros2topic`, `ros2action`, `ros2interface`, `ros2lifecycle`, `ros2node`, `ros2pkg`, and `ros2service` (repeated 5 times):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13414/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8334/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11139/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13473/)

ivanpauno (approved):
LGTM with green CI!

It's worth notifying everybody about these tests being reactivated, just in case they start hanging again.

hidmic:
Once/if Windows CI passes, I'll release `launch` into Rolling to get that Rpr job passing.

hidmic:
Hmm, tests don't hang anymore but they are as flaky as possible. I have to circle back.

ivanpauno:
This change could also be reverted [here](https://github.com/ros2/ros2cli/blob/7bea45f0e7d5e75cff0e57da79ba00d73a8330e2/ros2param/test/test_verb_dump.py#L75-L80) and [here](https://github.com/ros2/ros2cli/blob/7bea45f0e7d5e75cff0e57da79ba00d73a8330e2/ros2param/test/test_verb_load.py#L75-L80).

hidmic:
Oh, I missed those. Thanks @ivanpauno ! I'm still fighting tests to pass on Windows though.

hidmic:
Alright, with last commits I can no longer reproduce flaky tests on Windows.

Spinning up CI up to `ros2topic`, `ros2action`, `ros2interface`, `ros2lifecycle`, `ros2node`, `ros2pkg`, and `ros2service` (repeated up to 5 times until failure):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13699/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8576/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11415/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13770/)

hidmic:
Hmm, some failing `ros2node` tests need an increase in their timeouts, but `ros2topic` test failures are perplexing. I cannot reproduce them in a Windows VM.

As side note, I am a bit worried about the long timeouts. `ros2 action info` takes about 15 seconds to query the daemon and reply on Windows. It's not unusable, but it's not far from it either. FYI @clalancette.

clalancette:
> As side note, I am a bit worried about the long timeouts. `ros2 action info` takes about 15 seconds to query the daemon and reply on Windows. It's not unusable, but it's not far from it either. FYI @clalancette.

Yeah, that's a problem.  It's not useful for users, and the other problem is that adding these test look like they will add ~1 hour to our (already very long) CI times.

https://github.com/osrf/osrf_pycommon/pull/66 will help somewhat .  Other than that, we'd need to go in and do an analysis of what is taking so long.

hidmic:
I'm proactively putting this in the backlog. Ideally, I think we should figure out why these CLIs are so slow on Windows for before Galactic, but I don't know if that's realistic.

hidmic:
One thing I've noticed while doing manual source testing on Windows. CLIs are slow to respond _when the daemon is up_.

hidmic:
The issue with CLI daemons in Windows is tracked by https://github.com/ros2/ros2cli/issues/637.

PR #588, opened 2021-01-25
mod: enable flush output.

### LINKED ISSUE #587: Topic echo uses print function with flush=False
## Bug report

As the print function is the default (flush=False), output buffer blocks piped outputs until filled.
https://github.com/ros2/ros2cli/blob/master/ros2topic/ros2topic/verb/echo.py#L173

I think "echo" should flush output with every message.

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- Installation type:
  - from source
- Version or commit hash:
  - rolling (master branch)

#### Steps to reproduce issue

```
$ ros2 run demo_nodes_cpp talker
$ ros2 topic echo /chatter | grep data # (in other terminal)
```

#### Expected behavior
```
$ ros2 topic echo /chatter | grep data
data: 'Hello World 56' (Output every second)
data: 'Hello World 57'
data: 'Hello World 58'
...
```

#### Actual behavior
```
$ ros2 topic echo /chatter | grep data
(Wait a few minutes)
data: 'Hello World 56'
~~
data: 'Hello World 374'
(Wait a few minutes.)
...
```

hsgwa:
Signed-off-by: hsgwa <hasegawa@isp.co.jp>

Closes #587

### DIFF
--- a/ros2topic/ros2topic/verb/echo.py
+++ b/ros2topic/ros2topic/verb/echo.py
@@ -175,16 +175,16 @@ def _print_yaml(msg, truncate_length, noarr, nostr):
         print(
             message_to_yaml(
                 msg, truncate_length=truncate_length, no_arr=noarr, no_str=nostr),
-            end='---\n')
+            end='---\n', flush=True)
     else:
-        print(msg, end='\n---\n')
+        print(msg, end='\n---\n', flush=True)
 
 
 def _print_csv(msg, truncate_length, noarr, nostr):
     if hasattr(msg, '__slots__'):
-        print(message_to_csv(msg, truncate_length=truncate_length, no_arr=noarr, no_str=nostr))
+        print(message_to_csv(msg, truncate_length=truncate_length, no_arr=noarr, no_str=nostr), flush=True)
     else:
-        print(msg)
+        print(msg, flush=True)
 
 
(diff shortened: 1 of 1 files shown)


clalancette:
It turns out that there is a way to do this without any source changes at all.  If you run:

```
PYTHONUNBUFFERED=1 ros2 topic echo /chatter | grep data
```

That will print out the data on every output.  I think this is a better solution since it will work on *all* of the `ros2` cli commands, and not just `topic echo`.  Because of that, I'm going to close this PR (and the issue) out.  @hsgwa one thing that might be helpful here is to add some documentation somewhere in https://index.ros.org/doc/ros2/Tutorials/#tutorials that explains this.

Feel free to keep commenting or reopen if you disagree.

hsgwa:
Thank you for your comment!

Right. Expanding flush=True to other CLI commands needs source changes.
PYTHONUNBUFFERED solves this issue though, I'm thinking it's better to flush DEFAULT to improve tool usability.

ROS1 also flushes only echo, so there is also the aspect of improving consistency with ROS1.
https://github.com/ros/ros_comm/blob/noetic-devel/tools/rostopic/src/rostopic/__init__.py#L950

I considered a simple fix that would make each command enables flush, but I couldn't come up with a good idea.

Issue #589, opened 2021-01-25
ros2 param load parameters


v-lopez:
#### Feature description
A ros2 param verb for loading a parameter file into a node. Parameters currently can only be dumped from the cli interface, but to load them they must go through the launch interface.

clalancette:
That seems useful.  Please feel free to open a PR with that new functionality.

PR #590, opened 2021-01-26
Add rosparam verb load

v-lopez:
Example execution:

```
ros2 param  load /controller_manager ~/controller_manager.yaml 
Set parameter diff_drive_controller.type successful
Set parameter forward_command_controller_position.type successful
Set parameter joint_state_controller.type successful
```

```
ros2 param  load /controller_manager ~/controller_manager.yaml  --use-wildcard 
Set parameter mystr successful
Set parameter diff_drive_controller.type successful
Set parameter diff_drive_controller.type successful
Set parameter forward_command_controller_position.type successful
Set parameter joint_state_controller.type successful
```

Using the following file:
```yaml
controller_manager:
  ros__parameters:
    diff_drive_controller:
      type: diff_drive_controller/DiffDriveController
    forward_command_controller_position:
      type: forward_command_controller/ForwardCommandController
    joint_state_controller:
      type: joint_state_controller/JointStateController
/**:
  ros__parameters:
    mystr: some_string
    diff_drive_controller:
      type: overriden_string
```

The parameters in the wildcard namespace are only used if `--use-wildcard` is provided, and are loaded before the node parameters, as I assume that should be the logical behavior.

fixes https://github.com/ros2/ros2cli/issues/589

### DIFF
--- a/ros2param/ros2param/api/__init__.py
+++ b/ros2param/ros2param/api/__init__.py
@@ -22,8 +22,9 @@ from rcl_interfaces.srv import SetParameters
 import rclpy
 from rclpy.parameter import PARAMETER_SEPARATOR_STRING
 from ros2cli.node.direct import DirectNode
-import yaml
+
 import sys
+import yaml
 
 
 def get_value(*, parameter_value):
@@ -114,7 +115,7 @@ def parse_parameter_dict(*, namespace, parameter_dict):
 
 def load_parameter_dict(*, node, node_name, parameter_dict):
 
-    parameters = parse_parameter_dict(namespace="", parameter_dict=parameter_dict)
+    parameters = parse_parameter_dict(namespace='', parameter_dict=parameter_dict)
     response = call_set_parameters(
         node=node, node_name=node_name, parameters=parameters)
 
@@ -138,20 +139,20 @@ def load_parameter_dict(*, node, node_name, parameter_dict):
 def load_parameter_file(*, node, node_name, parameter_file):
     # Remove leading slash and namespaces
     internal_node_name = node_name.split('/')[-1]
-    with open(parameter_file, "r") as f:
+    with open(parameter_file, 'r') as f:
         param_file = yaml.safe_load(f)
         if internal_node_name not in param_file:
-            raise RuntimeError("Param file doesn't contain parameters for {}, "
-                               " only for namespaces: {}" .format(internal_node_name,
+            raise RuntimeError('Param file does not contain parameters for {}, '
+                               ' only for namespaces: {}' .format(internal_node_name,
                                                                   param_file.keys()))
 
         value = param_file[internal_node_name]
-        if type(value) != dict or "ros__parameters" not in value:
-            raise RuntimeError("Invalid structure of parameter file in namespace {}"
-                               "expected same format as provided by ros2 param dump"
+        if type(value) != dict or 'ros__parameters' not in value:
+            raise RuntimeError('Invalid structure of parameter file in namespace {}'
+                               'expected same format as provided by ros2 param dump'
                                .format(internal_node_name))
         load_parameter_dict(node=node, node_name=node_name,
-                            parameter_dict=value["ros__parameters"])
+                            parameter_dict=value['ros__parameters'])
 
 
 def call_describe_parameters(*, node, node_name, parameter_names=None):
--- a/ros2param/ros2param/api/__init__.py
+++ b/ros2param/ros2param/api/__init__.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from rcl_interfaces.msg import Parameter
 from rcl_interfaces.msg import ParameterType
 from rcl_interfaces.msg import ParameterValue
 from rcl_interfaces.srv import DescribeParameters
@@ -19,8 +20,10 @@ from rcl_interfaces.srv import GetParameters
 from rcl_interfaces.srv import ListParameters
 from rcl_interfaces.srv import SetParameters
 import rclpy
+from rclpy.parameter import PARAMETER_SEPARATOR_STRING
 from ros2cli.node.direct import DirectNode
 import yaml
+import sys
 
 
 def get_value(*, parameter_value):
@@ -90,6 +93,46 @@ def get_parameter_value(*, string_value):
         value.type = ParameterType.PARAMETER_STRING
         value.string_value = string_value
     return value
+
+
+def parse_parameter_dict(*, namespace, parameter_dict):
+    parameters = []
+    for param_name, param_value in parameter_dict.items():
+        full_param_name = namespace + param_name
+        # Unroll nested parameters
+        if type(param_value) == dict:
+            parameters += parse_parameter_dict(
+                    namespace=full_param_name + PARAMETER_SEPARATOR_STRING,
+                    parameter_dict=param_value)
+        else:
+            parameter = Parameter()
+            parameter.name = full_param_name
+            parameter.value = get_parameter_value(string_value=str(param_value))
+            parameters.append(parameter)
+    return parameters
+
+
+def load_parameter_dict(*, node, node_name, parameter_dict):
+
+    parameters = parse_parameter_dict(namespace="", parameter_dict=parameter_dict)
+    response = call_set_parameters(
+        node=node, node_name=node_name, parameters=parameters)
+
+    # output response
+    assert len(response.results) == len(parameters)
+    for i in range(0, len(response.results)):
+        result = response.results[i]
+        param_name = parameters[i].name
+        if result.successful:
+            msg = 'Set parameter {} successful'.format(param_name)
+            if result.reason:
+                msg += ': ' + result.reason
+            print(msg)
+        else:
+            msg = 'Set parameter {} failed'.format(param_name)
+            if result.reason:
+                msg += ': ' + result.reason
+            print(msg, file=sys.stderr)
 
 
 def call_describe_parameters(*, node, node_name, parameter_names=None):
--- a/ros2param/ros2param/api/__init__.py
+++ b/ros2param/ros2param/api/__init__.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import sys
+
 from rcl_interfaces.msg import Parameter
 from rcl_interfaces.msg import ParameterType
 from rcl_interfaces.msg import ParameterValue
@@ -23,7 +25,6 @@ import rclpy
 from rclpy.parameter import PARAMETER_SEPARATOR_STRING
 from ros2cli.node.direct import DirectNode
 
-import sys
 import yaml
 
 
--- a/ros2param/ros2param/api/__init__.py
+++ b/ros2param/ros2param/api/__init__.py
@@ -135,6 +135,25 @@ def load_parameter_dict(*, node, node_name, parameter_dict):
             print(msg, file=sys.stderr)
 
 
+def load_parameter_file(*, node, node_name, parameter_file):
+    # Remove leading slash and namespaces
+    internal_node_name = node_name.split('/')[-1]
+    with open(parameter_file, "r") as f:
+        param_file = yaml.safe_load(f)
+        if internal_node_name not in param_file:
+            raise RuntimeError("Param file doesn't contain parameters for {}, "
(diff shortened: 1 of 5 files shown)


v-lopez:
@clalancette could you please give this a look? I've got some other PR's that depend on this functionality.

audrow (commented):
This looks like a useful feature, thanks for the PR.

A few thoughts:
* I'd like to see tests for the load verb.
* I don't think that the wildcard behavior should be a part of this PR. There is some (currently stalled) discussion on parameter files and presidence - see https://github.com/ros2/design/pull/289 and https://github.com/ros2/design/pull/303 - that I think we should wait on. The main reason for this is that I think that we should keep the wildcard behavior of `ros2param`, `ros2launch`, and `ros2run` similar.

v-lopez:
@audrow Added tests and removed wildcard behavior. I'm not sure why  those unrelated tests are failing.

audrow:
I'll give this a more thorough review soon (probably Friday). 

By the way, I ran CI. There are a few linter errors.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13638/)

v-lopez:
Thanks, I pushed fixes for the linter. I apologise for that, but I cannot get those errors to show up on my machine. I'm running latest foxy and same flake8 version and never get the bad quotes errors.

audrow:
> I cannot get those errors to show up on my machine

It works for me when I use `colcon test` (Rolling with Ubuntu); for example, `colcon test --packages-select ros2param` from the root of my ROS2 workspace, and `colcon test-result --verbose` to view the test failures. If you're already doing that, I'm not sure why the linter wouldn't run.

Also, if the other tests are passing but you're fixing linter warnings, the colcon test mixin [`linters-only`](https://github.com/colcon/colcon-mixin-repository/blob/1ddb69bedfd1f04c2f000e95452f7c24a4d6176b/test-linters.mixin#L3) is much more convenient, especially for packages with slow tests.

audrow:
It would be nice to make a PR to document this new feature in the [ROS2 documentation on parameters](https://index.ros.org/doc/ros2/Tutorials/Parameters/Understanding-ROS2-Parameters).

I believe it makes sense to add to the ["Load parameter file" section of the Understanding ROS2 parameters](https://github.com/ros2/ros2_documentation/blob/master/source/Tutorials/Parameters/Understanding-ROS2-Parameters.rst#load-parameter-file) tutorial.

If you make a PR there, I can review it.

v-lopez:
I will add the documentation, probably on Friday as well.

ros2param/test/test_verb_load.py
```diff
@@ -0,0 +1,284 @@
...
+            assert param_load_command.wait_for_shutdown(timeout=TEST_TIMEOUT)
+        assert param_load_command.exit_code != launch_testing.asserts.EXIT_OK
+        assert launch_testing.tools.expect_output(
+            expected_lines=['ros2 param load: error: the following arguments are required: '
+                            'parameter_file'],
+            text=param_load_command.output,
+            strict=False
+        )
+
+        # TODO
```
audrow, line 223: Any reason for this TODO, or the one below?

audrow (approved):
Looks good to me. I'll run CI.

audrow:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13656/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8535/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11371/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13720/)

audrow:
@v-lopez, I'm not sure why but I wasn't able to push changes fixing the linter errors to your branch. Anyway, I've made a commit to fix the linter errors: https://github.com/audrow/ros2cli/commit/812a5be8b4a5f099a51d611fc0d6e7d9c213468c. Once you apply these changes, I'll run CI and, if it's green, we can merge in the two PRs.

v-lopez:
Fixes applied!

audrow:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13663/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8542/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11380/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13729/)

audrow:
I've merged both PRs, thanks for the contribution @v-lopez!

v-lopez:
Thanks to you! Should I make a PR for the `foxy` branch to backport it?

ros2param/ros2param/api/__init__.py
```diff
@@ -92,6 +97,65 @@ def get_parameter_value(*, string_value):
...
+        else:
+            msg = 'Set parameter {} failed'.format(param_name)
+            if result.reason:
+                msg += ': ' + result.reason
+            print(msg, file=sys.stderr)
+
+
+def load_parameter_file(*, node, node_name, parameter_file):
+    # Remove leading slash and namespaces
+    internal_node_name = node_name.split('/')[-1]
```
ivanpauno, line 142: why removing the namespace?

The name in a parameter file is a fully qualified name.
IIRC, both having a leading slash or not is accepted.

ros2param/ros2param/api/__init__.py
```diff
@@ -92,6 +97,65 @@ def get_parameter_value(*, string_value):
...
+    with open(parameter_file, 'r') as f:
+        param_file = yaml.safe_load(f)
+        if internal_node_name not in param_file:
```
ivanpauno, line 145: what about `/**`?

I don't see how that is being handled

audrow:
Thanks for pointing those cases out, @ivanpauno. @v-lopez, can you make another PR addressing @ivanpauno's comments, since this has already been merged?

By the way, I checked with the foxy ROS Boss and you can make a PR for a foxy backport. This probably makes sense to do after the new comments are addressed.

Feel free to @ me for both PRs.

v-lopez:
I'll work on a PR tomorrow to address this.

@ivanpauno  I will check for both with and without namespace.

Regarding /** it was in there originally but I removed it on https://github.com/ros2/ros2cli/pull/590/commits/6e0619506c1828e1f5853cf6a78cf36b5baaa18d as discussed in this PR. I'd like to have this or a similar behavior added, but I guess we are waiting for unique criteria on: https://github.com/ros2/design/pull/289 and https://github.com/ros2/design/pull/303

ivanpauno:
> @ivanpauno I will check for both with and without namespace.

You should only check for the fully qualified name.
e.g. `namespace=/asd/bsd` `name=csd`

you should check for `/asd/bsd/csd` and `asd/bsd/csd` (but not for `csd` or `/csd).

> Regarding /** it was in there originally but I removed it on 6e06195 as discussed in this PR

Can you link the discussion? I wasn't able to find it.

> I'd like to have this or a similar behavior added, but I guess we are waiting for unique criteria on: ros2/design#289 and ros2/design#303

IIRC the current behavior is that parameter overrides specific to a node will override the ones passed using the `/**` wildcard, so I think it's fine to imitate that behavior here without waiting for the other discussions.

v-lopez:
> Can you link the discussion? I wasn't able to find it.

Well not a big discussion, just this comment https://github.com/ros2/ros2cli/pull/590#pullrequestreview-584756953

>IIRC the current behavior is that parameter overrides specific to a node will override the ones passed using the /** wildcard, so I think it's fine to imitate that behavior here without waiting for the other discussions.

@audrow Should I put the wildcard behavior back in then?

audrow:
@v-lopez, yes, I think it would make sense to support wildcard behavior. Sorry for the misdirection.

My intention in https://github.com/ros2/ros2cli/pull/590#pullrequestreview-584756953 was that this new CLI verb doesn't get ahead of the discussion I mentioned on presidence in parameter files, by adding the ability to prioritize for parameters defined with a wildcard. I didn't mean to eliminate handling the wildcard case, which I missed in my review. 

It would also be great to update the tests to show that wildcards (and namespaces) are handled :)

v-lopez:
>    You should only check for the fully qualified name.
> e.g. namespace=/asd/bsd name=csd
> 
> you should check for /asd/bsd/csd and asd/bsd/csd (but not for csd or `/csd).

@ivanpauno Is this right? Didn't you mean to check for `csd` and `/csd` but not for the namespace?

As far as I see, the param files do not have the namespace. Only the node name, which could be with or without leading slash.
Using the test node of the ros2param tests, which has a namespace:
`ros2 param dump /foo/test_node`
```
test_node:
  ros__parameters:
    bool_array_param:
    - false
...
```

ivanpauno:
> As far as I see, the param files do not have the namespace. Only the node name, which could be with or without leading slash.

Well, `ros2 param dump` seems to be broken :no_mouth: ....
But the original use case of a parameter file was passing it to a node:

```
node_executable --ros-args --params-file /path/to/my/parameter/file
```

and in that case it will look for `/foo/test_node` (`ros2 param dump` should be generating something compatible to that).

e.g.: rclcpp code [here](https://github.com/ros2/rclcpp/blob/ab743392e4daf546eb0de704936b5f14ef6e0417/rclcpp/src/rclcpp/node_interfaces/node_parameters.cpp#L96-L97)

clalancette:
> and in that case it will look for `/foo/test_node` (`ros2 param dump` should be generating something compatible to that).

Yikes, good call.  It does seem to be broken.  Do you mind opening an issue?

v-lopez:
Leaving aside the implications of the changes, I believe it might be interesting to be able to load params for a node, regardless of the namespace.

Say I have navigation parameters for a robot, and I want to launch a simulation with multiple robots of the same type. Would I have to create multiple identical files, each with its own namespace? Do some magic at launch time to rename them?

ivanpauno:
> Yikes, good call. It does seem to be broken. Do you mind opening an issue?

I'm working on a PR for it.

> Say I have navigation parameters for a robot, and I want to launch a simulation with multiple robots of the same type. Would I have to create multiple identical files, each with its own namespace? Do some magic at launch time to rename them?

That's a good question.
Ideally, `my_robot` would refer to any node named `my_robot` and `/namespace/my_robot` would refer specifically to one.
But we are already using `my_robot` as `/my_robot` in parameter files, so I don't know.

For the moment, I would focus on consistency and not in that particular use case (which I agree is useful).

v-lopez:
>  Ideally, my_robot would refer to any node named my_robot and /namespace/my_robot would refer specifically to one.

That would make sense to us. 

And I understand this is not the place to discuss this, could we move the discussion elsewhere? I think it's going to be important in the near future.

ivanpauno:
> I'm working on a PR for it.

See https://github.com/ros2/ros2cli/pull/600.

ivanpauno:
> And I understand this is not the place to discuss this, could we move the discussion elsewhere? I think it's going to be important in the near future.

Sounds good, maybe you can open an issue with the proposal in `ros2/ros2`, as it will involve changes in different places (rclcpp, rclpy, ros2param, ...).

v-lopez:
I will, but I'd love to have some backing from you guys, otherwise this things get on a huge todo pile.

I'll ping you when I've opened it.

ivanpauno:
> why removing the namespace?
>
> The name in a parameter file is a fully qualified name.
> IIRC, both having a leading slash or not is accepted.

I finally solved this issue in https://github.com/ros2/ros2cli/pull/600, as the tests for `ros2 param load` depend on `ros2 param dump` behavior.

> what about /**?
> 
> I don't see how that is being handled

That one is still pending.

ivanpauno:
> I will, but I'd love to have some backing from you guys, otherwise this things get lost in a huge todo pile.

Sounds good, I understand that pushing new features/enhancements is demanding.

> I'll ping you when I've opened it.

:+1: feel free to tag me.
I think the only point of discussion will be if the change of behavior for relative node names in parameter files is acceptable, or if we need to somehow handle that case in a backwards compatible fashion.

ros2param/ros2param/api/__init__.py
ivanpauno: will be solved by #600

v-lopez:
@ivanpauno Please see https://github.com/ros2/ros2cli/pull/602 for the wildcard support.

v-lopez:
And the issue about node parameters regardless of namespaces: https://github.com/ros2/ros2/issues/1092

PR #611, opened 2021-03-22
Use generate_setuptools_dict

DLu:
See also https://github.com/ament/ament_package/pull/126 and https://github.com/ros2/ros2_documentation/pull/1250


audrow:
We should hold on this until we figure out what to do with helper functions - see https://github.com/ament/ament_package/pull/126.

Issue #594, opened 2021-02-01
Try to autodetect matching QoS in `ros2 topic pub/echo`


nnmm:
## Feature request
It could be a great usability improvement if `ros2 topic echo` would make an effort to use a matching subscription QoS, and similarly for `pub`. For a previous discussion, see #156. The basic idea would be (talking about `echo` but the same applies to `pub`):

* When all publishers have the same/similar enough QoS, it could be possible to find a matching QoS profile and use that.
* When there are multiple QoS, the user has to make a choice.
* This should probably work in combination with the `--qos` flags, which would fix the specified parameters and let the others be deduced
* It should be made the default behavior, if there are no big downsides

But that only talks about a single point in time, whereas for a long-running `echo` command publishers can come and go. That raises the questions:
* Should the subscription "recalculate" its QoS when a new subscription/publisher on the other side joins? That would mean checking that there is a QoS that's consistent with the publishers and if that is not the current QoS, applying that to itself.
  * If the answer is "no", note that it would be very unintuitive to get different output when you run two publishers, depending on which one started first (`pub1`, `echo`, `pub2` vs `pub2`, `echo`, `pub1`). If I understand correctly, there are warnings about this now, which help somewhat.
  * If the answer is "yes", should it prompt the user when there is no consistent QoS, or just exit with a message?
* What should the QoS be when there is nothing to match with yet?
  * This is related to the previous question: If the QoS is recalculated, that "default" QoS profile is meaningless anyway

gonzodepedro:
I'm working on changes to implement partially some of this changes into ros2 topic echo. I have some questions:

1. Should this be circunscribed only to QoS profiles that come with ros2? or should it handle other possible combinations of QoS parameters?
2. With current default QoS parameters (best_effort, volatile), the echo subscriber should be able to receive messages from any publisher. But if the publisher is transient_local, then older messages won't be received. The idea would be to implement a logic that when all the publishers are transient_local, then make the subscriber transient_local as well, in order to receive older messages. Is that a step in the right direction?
3. I cannot find a reasonable way to detect when a new publisher joins, and to change QoS settings without risking to loose messages sent while the QoS change takes place. Any thoughts?

jacobperron:
> 1. Should this be circunscribed only to QoS profiles that come with ros2? or should it handle other possible combinations of QoS parameters?

I think it should handle any possible combination of QoS policies. I.e. we shouldn't limit ourselves to the preset profiles as users are free to create custom profiles.

[...]

PR #613, opened 2021-03-23
QoS autodetection

gonzodepedro:
First iteration. Connects to #594 

### DIFF
--- a/ros2topic/ros2topic/api/__init__.py
+++ b/ros2topic/ros2topic/api/__init__.py
@@ -29,6 +29,7 @@ from rosidl_runtime_py.utilities import get_message
 
 default_preset_str = 'sensor_data'
 
+
 def unsigned_int(string):
     try:
         value = int(string)
@@ -165,7 +166,7 @@ def qos_profile_from_short_keys(
 ) -> rclpy.qos.QoSProfile:
     """Construct a QoSProfile given the name of a preset, and optional overrides."""
     # Build a QoS profile based on user-supplied arguments
-    if not preset_profile :
+    if not preset_profile:
         preset_profile = default_preset_str
     profile = rclpy.qos.QoSPresetProfiles.get_from_short_key(preset_profile)
     if history:
@@ -185,21 +186,20 @@ def qos_profile_from_short_keys(
 
 
 def add_qos_arguments_to_argument_parser(
-    parser: argparse.ArgumentParser, is_publisher: bool = True, default_preset: str = default_preset_str
+    parser: argparse.ArgumentParser, is_publisher: bool = True,
+    default_preset: str = default_preset_str
 ) -> None:
     """Extend an existing ArgumentParser to allow input of QoS policy overrides."""
     verb = 'publish' if is_publisher else 'subscribe'
     parser.add_argument(
         '--qos-profile',
         choices=rclpy.qos.QoSPresetProfiles.short_keys(),
-        #default=default_preset,
         help='Quality of service preset profile to {} with (default: {})'
              .format(verb, default_preset))
     default_profile = rclpy.qos.QoSPresetProfiles.get_from_short_key(
         default_preset)
     parser.add_argument(
         '--qos-depth', metavar='N', type=int,
-        #default=-1,
         help='Queue size setting to {} with '
              '(overrides depth value of --qos-profile option)'
              .format(verb))
--- a/ros2topic/ros2topic/api/__init__.py
+++ b/ros2topic/ros2topic/api/__init__.py
@@ -12,10 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import argparse
-
 from argparse import ArgumentTypeError
 from time import sleep
+from typing import Optional
 
 import rclpy
 
@@ -26,8 +25,6 @@ from ros2cli.node.strategy import NodeStrategy
 from rosidl_runtime_py import get_message_interfaces
 from rosidl_runtime_py import message_to_yaml
 from rosidl_runtime_py.utilities import get_message
-
-default_preset_str = 'sensor_data'
 
 
 def unsigned_int(string):
@@ -162,12 +159,10 @@ class TopicMessagePrototypeCompleter:
 
 def qos_profile_from_short_keys(
     preset_profile: str, reliability: str = None, durability: str = None,
-    depth: int = -1, history: str = None,
+    depth: Optional[int] = None, history: str = None,
 ) -> rclpy.qos.QoSProfile:
     """Construct a QoSProfile given the name of a preset, and optional overrides."""
     # Build a QoS profile based on user-supplied arguments
-    if not preset_profile:
-        preset_profile = default_preset_str
     profile = rclpy.qos.QoSPresetProfiles.get_from_short_key(preset_profile)
     if history:
         profile.history = rclpy.qos.QoSHistoryPolicy.get_from_short_key(history)
@@ -182,41 +177,3 @@ def qos_profile_from_short_keys(
                 and profile.depth == 0):
             profile.depth = 1
     return profile
-
-
-def add_qos_arguments_to_argument_parser(
-    parser: argparse.ArgumentParser, is_publisher: bool = True,
-    default_preset: str = default_preset_str
-) -> None:
-    """Extend an existing ArgumentParser to allow input of QoS policy overrides."""
-    verb = 'publish' if is_publisher else 'subscribe'
-    parser.add_argument(
-        '--qos-profile',
-        choices=rclpy.qos.QoSPresetProfiles.short_keys(),
-        help='Quality of service preset profile to {} with (default: {})'
-             .format(verb, default_preset))
-    default_profile = rclpy.qos.QoSPresetProfiles.get_from_short_key(
-        default_preset)
-    parser.add_argument(
-        '--qos-depth', metavar='N', type=int,
-        help='Queue size setting to {} with '
-             '(overrides depth value of --qos-profile option)'
-             .format(verb))
-    parser.add_argument(
-        '--qos-history',
-        choices=rclpy.qos.QoSHistoryPolicy.short_keys(),
-        help='History of samples setting to {} with '
-             '(overrides history value of --qos-profile option, default: {})'
-             .format(verb, default_profile.history.short_key))
-    parser.add_argument(
-        '--qos-reliability',
-        choices=rclpy.qos.QoSReliabilityPolicy.short_keys(),
-        help='Quality of service reliability setting to {} with '
-             '(overrides reliability value of --qos-profile option, default: {})'
-             .format(verb, default_profile.reliability.short_key))
-    parser.add_argument(
-        '--qos-durability',
-        choices=rclpy.qos.QoSDurabilityPolicy.short_keys(),
-        help='Quality of service durability setting to {} with '
-             '(overrides durability value of --qos-profile option, default: {})'
-             .format(verb, default_profile.durability.short_key))
--- a/ros2topic/ros2topic/api/__init__.py
+++ b/ros2topic/ros2topic/api/__init__.py
@@ -181,7 +181,6 @@ def qos_profile_from_short_keys(
         if (profile.durability == rclpy.qos.QoSDurabilityPolicy.TRANSIENT_LOCAL
                 and profile.depth == 0):
             profile.depth = 1
-
     return profile
 
 
--- a/ros2topic/ros2topic/api/__init__.py
+++ b/ros2topic/ros2topic/api/__init__.py
@@ -27,6 +27,7 @@ from rosidl_runtime_py import get_message_interfaces
 from rosidl_runtime_py import message_to_yaml
 from rosidl_runtime_py.utilities import get_message
 
+default_preset_str = 'sensor_data'
 
 def unsigned_int(string):
     try:
@@ -164,6 +165,8 @@ def qos_profile_from_short_keys(
 ) -> rclpy.qos.QoSProfile:
     """Construct a QoSProfile given the name of a preset, and optional overrides."""
     # Build a QoS profile based on user-supplied arguments
+    if not preset_profile :
+        preset_profile = default_preset_str
     profile = rclpy.qos.QoSPresetProfiles.get_from_short_key(preset_profile)
     if history:
         profile.history = rclpy.qos.QoSHistoryPolicy.get_from_short_key(history)
@@ -171,7 +174,7 @@ def qos_profile_from_short_keys(
         profile.durability = rclpy.qos.QoSDurabilityPolicy.get_from_short_key(durability)
     if reliability:
         profile.reliability = rclpy.qos.QoSReliabilityPolicy.get_from_short_key(reliability)
-    if depth >= 0:
+    if depth and depth >= 0:
         profile.depth = depth
     else:
         if (profile.durability == rclpy.qos.QoSDurabilityPolicy.TRANSIENT_LOCAL
@@ -182,20 +185,21 @@ def qos_profile_from_short_keys(
 
 
 def add_qos_arguments_to_argument_parser(
-    parser: argparse.ArgumentParser, is_publisher: bool = True, default_preset: str = 'sensor_data'
+    parser: argparse.ArgumentParser, is_publisher: bool = True, default_preset: str = default_preset_str
 ) -> None:
     """Extend an existing ArgumentParser to allow input of QoS policy overrides."""
     verb = 'publish' if is_publisher else 'subscribe'
     parser.add_argument(
         '--qos-profile',
         choices=rclpy.qos.QoSPresetProfiles.short_keys(),
-        default=default_preset,
+        #default=default_preset,
         help='Quality of service preset profile to {} with (default: {})'
              .format(verb, default_preset))
     default_profile = rclpy.qos.QoSPresetProfiles.get_from_short_key(
         default_preset)
     parser.add_argument(
-        '--qos-depth', metavar='N', type=int, default=-1,
+        '--qos-depth', metavar='N', type=int,
+        #default=-1,
         help='Queue size setting to {} with '
              '(overrides depth value of --qos-profile option)'
              .format(verb))
--- a/ros2topic/ros2topic/verb/echo.py
+++ b/ros2topic/ros2topic/verb/echo.py
@@ -26,7 +26,6 @@ from rclpy.qos_event import UnsupportedEventTypeError
 from rclpy.utilities import get_rmw_implementation_identifier
 from ros2cli.node.strategy import add_arguments as add_strategy_node_arguments
 from ros2cli.node.strategy import NodeStrategy
-from ros2topic.api import add_qos_arguments_to_argument_parser
 from ros2topic.api import get_msg_class
 from ros2topic.api import qos_profile_from_short_keys
 from ros2topic.api import TopicNameCompleter
@@ -38,6 +37,7 @@ from rosidl_runtime_py.utilities import get_message
 
 DEFAULT_TRUNCATE_LENGTH = 128
 MsgType = TypeVar('MsgType')
+default_profile_str = 'sensor_data'
 
 
 class EchoVerb(VerbExtension):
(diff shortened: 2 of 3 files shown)


gonzodepedro:
Things to do:
- [ ] Autodetect when nodes come and go
- [ ] Consider using 'QoS Profile compatibility" API

ros2topic/ros2topic/verb/echo.py
```diff
@@ -99,14 +154,10 @@ def main(self, *, args):
...
-        qos_profile = qos_profile_from_short_keys(
-            args.qos_profile,
-            reliability=args.qos_reliability,
-            durability=args.qos_durability,
-            depth=args.qos_depth,
-            history=args.qos_history)
-
         with NodeStrategy(args) as node:
+
+            qos_profile = self.chooseQoS(node, args)
```
ivanpauno, line 159: ideally, if the user provided an argument that should be preferred over the autodetection logic.
I mean one of these:

https://github.com/ros2/ros2cli/blob/56db2ba9fcc85ead292b0c3403cf27dba41854e2/ros2topic/ros2topic/api/__init__.py#L189-L219

@jacobperron what do you think?

ivanpauno: to handle that I would delete the defaults from the argument and make them optional, so you can check if they were set or not

ivanpauno (commented):
LGTM after making user provided qos settings have higher precedence than the autodetected ones.
I've also left some minor comments

---

The TODOs can be left for a follow-up, but here some comments:

>  Autodetect when nodes come and go

I would rather do what was commented [here](https://github.com/ros2/ros2cli/issues/594#issuecomment-790123161):

```
Maybe we don't have to detect when a new publisher joins exactly, rather we can detect when there's a new publisher with an incompatible QoS. There is a callback API that can let us know when a new incompatible publisher is detected (see this example).
```

I actually think that we're installing an "incompatible qos" event handler by default, so maybe nothing is needed here (?).
You can try to force an incompatibility and see if you get a warning.

>  Consider using 'QoS Profile compatibility" API

I'm not sure if that will help much here.

ros2topic/ros2topic/verb/echo.py
```diff
@@ -82,6 +84,59 @@ def add_arguments(self, parser, cli_name):
...
+            for info in node.get_publishers_info_by_topic(args.topic_name):
+                node_count += 1
+                if (info.qos_profile.reliability ==
+                    QoSReliabilityPolicy.RELIABLE):
+                    reliability_reliable_endpoints_count += 1
+                if (info.qos_profile.durability ==
+                    QoSDurabilityPolicy.TRANSIENT_LOCAL):
+                    durability_transient_local_endpoints_count += 1
+        except NotImplementedError as e:
+            return str(e)
```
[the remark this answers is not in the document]
jacobperron: If it's not implemented, it seems like a bug and we should just let the exception propagate.

ros2topic/ros2topic/verb/echo.py
```diff
@@ -82,6 +84,59 @@ def add_arguments(self, parser, cli_name):
...
+                print(
+                    'Some, but not all, publisher are offering '
+                    'QoSReliabilityPolicy.RELIABLE. Falling back to '
+                    'QoSReliabilityPolicy.BEST_EFFORT as it will connect '
+                    'to all publishers', end='\n\n'
+                )
+            qos_profile.reliability = QoSReliabilityPolicy.BEST_EFFORT
+
+        # If all endpoints are transient_local, ask for transient_local
+        if  durability_transient_local_endpoints_count == node_count:
```
jacobperron, line 126:
```suggestion
        if durability_transient_local_endpoints_count == node_count:
```

ros2topic/ros2topic/verb/echo.py
jacobperron: I agree. The CLI options should have priority over the auto-detection logic.

jacobperron:
I agree with @ivanpauno about the following two points:

> Autodetect when nodes come and go

Just produce a warning if a new publisher joins with incompatible QoS (don't update the subscriber QoS). Maybe we don't have to implement the warning as Ivan points out.

> Consider using 'QoS Profile compatibility" API

I think the API we could reuse would be this proposed RMW API: https://github.com/ros2/rmw/issues/304

Since that API doesn't exist (yet), we could just add a comment referencing the proposal.

ros2topic/ros2topic/verb/echo.py
```diff
@@ -82,6 +84,59 @@ def add_arguments(self, parser, cli_name):
         parser.add_argument(
             '--raw', action='store_true', help='Echo the raw binary representation')
 
+    def chooseQoS(self, node, args):
```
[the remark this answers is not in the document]
gonzodepedro: Addresed in [8efa6ed](https://github.com/ros2/ros2cli/pull/613/commits/8efa6ede28b8f5a77c912a5ae690f0cf5db7ba75)

ros2topic/ros2topic/verb/echo.py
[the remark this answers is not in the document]
gonzodepedro: Addresed in [8efa6ed](https://github.com/ros2/ros2cli/pull/613/commits/8efa6ede28b8f5a77c912a5ae690f0cf5db7ba75)

ros2topic/ros2topic/verb/echo.py
```diff
@@ -82,6 +84,59 @@ def add_arguments(self, parser, cli_name):
...
+            depth=args.qos_depth,
+            history=args.qos_history)
+
+        reliability_reliable_endpoints_count = 0
+        durability_transient_local_endpoints_count = 0
+        node_count = 0
+
+        try:
+            for info in node.get_publishers_info_by_topic(args.topic_name):
+                node_count += 1
```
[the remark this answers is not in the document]
gonzodepedro: Addresed in [8efa6ed](https://github.com/ros2/ros2cli/pull/613/commits/8efa6ede28b8f5a77c912a5ae690f0cf5db7ba75)

ros2topic/ros2topic/verb/echo.py
gonzodepedro: I made some changes to give CLI options priority over auto-detection logic

Addresed in [8efa6ed](https://github.com/ros2/ros2cli/pull/613/commits/8efa6ede28b8f5a77c912a5ae690f0cf5db7ba75)

ros2topic/ros2topic/verb/echo.py
gonzodepedro: Addresed in [8efa6ed](https://github.com/ros2/ros2cli/pull/613/commits/8efa6ede28b8f5a77c912a5ae690f0cf5db7ba75)

gonzodepedro:
I'm looking at the failing tests

ros2topic/ros2topic/verb/echo.py
```diff
@@ -82,6 +85,61 @@ def add_arguments(self, parser, cli_name):
...
+        publishers_count = 0
+
+        for info in node.get_publishers_info_by_topic(args.topic_name):
+            publishers_count += 1
+            if (info.qos_profile.reliability ==
+                QoSReliabilityPolicy.RELIABLE):
+                    reliability_reliable_endpoints_count += 1
+            if (info.qos_profile.durability ==
+                QoSDurabilityPolicy.TRANSIENT_LOCAL):
+                    durability_transient_local_endpoints_count += 1
```
ivanpauno, line 110: Minor style comment, feel free to ignore:

---

this is a bit more readable (though it iters twice over the list):

```suggestion
        pubs_info = node.get_publishers_info_by_topic(args.topic_name)
        publishers_count = len(pubs_info)
        reliability_reliable_endpoints_count = sum(
            info.qos_profile.reliability == QoSReliabilityPolicy.RELIABLE for info in pubs_info)
        durability_transient_local_endpoints_count = sum(
            info.qos_profile.durability == QoSDurabilityPolicy.TRANSIENT_LOCAL for info in pubs_info)
```

ros2topic/ros2topic/verb/echo.py
```diff
@@ -82,6 +85,61 @@ def add_arguments(self, parser, cli_name):
...
+        # If all endpoints are reliable, ask for reliable
+        if reliability_reliable_endpoints_count == publishers_count:
+            qos_profile.reliability = QoSReliabilityPolicy.RELIABLE
+        else:
+            if reliability_reliable_endpoints_count > 0:
+                print(
+                    'Some, but not all, publisher are offering '
+                    'QoSReliabilityPolicy.RELIABLE. Falling back to '
+                    'QoSReliabilityPolicy.BEST_EFFORT as it will connect '
+                    'to all publishers', end='\n\n'
```
ivanpauno, line 124: is there any benefit of using `end`?
do we need to print this message?
should we use stderr?

```suggestion
                    'to all publishers\n\n'
```

ros2topic/ros2topic/verb/echo.py
```diff
@@ -82,6 +85,61 @@ def add_arguments(self, parser, cli_name):
...
+        # If all endpoints are transient_local, ask for transient_local
+        if durability_transient_local_endpoints_count == publishers_count:
+            qos_profile.durability = QoSDurabilityPolicy.TRANSIENT_LOCAL
+        else:
+            if durability_transient_local_endpoints_count > 0:
+                print(
+                    'Some, but not all, publisher are offering '
+                    'QoSDurabilityPolicy.TRANSIENT_LOCAL. Falling back to '
+                    'QoSDurabilityPolicy.VOLATILE as it will connect '
+                    'to all publishers', end='\n\n'
```
ivanpauno, line 137: same comments here

ivanpauno (commented):
LGTM!

I left some minor comments that can be ignored.
We should wait until a `galactic` branch is created before merging though.

ros2topic/ros2topic/api/__init__.py
```diff
@@ -182,20 +186,20 @@ def qos_profile_from_short_keys(
 
 
 def add_qos_arguments_to_argument_parser(
-    parser: argparse.ArgumentParser, is_publisher: bool = True, default_preset: str = 'sensor_data'
+    parser: argparse.ArgumentParser, is_publisher: bool = True,
+    default_preset: str = default_preset_str
```
ivanpauno, line 190: this argument is unused

ros2topic/ros2topic/verb/echo.py
gonzodepedro: Thanks for pointing this out. I made some changes based on your comments, but not exactly the one you suggested. 

Addressed in: [751f584](https://github.com/gonzodepedro/ros2cli/commit/751f58481e87736d7cf2887b279b900c82b10b85)

ros2topic/ros2topic/verb/echo.py
gonzodepedro: Addressed in: [afb4873
](https://github.com/gonzodepedro/ros2cli/commit/afb4873b11076b8a1a4f80f01a8c724ae97dc197)

ros2topic/ros2topic/verb/echo.py
gonzodepedro: Addressed in: [afb4873
](https://github.com/gonzodepedro/ros2cli/commit/afb4873b11076b8a1a4f80f01a8c724ae97dc197)

ros2topic/ros2topic/api/__init__.py
gonzodepedro: Used [here](https://github.com/gonzodepedro/ros2cli/blob/afb4873b11076b8a1a4f80f01a8c724ae97dc197/ros2topic/ros2topic/api/__init__.py#L199)

ros2topic/ros2topic/verb/echo.py
```diff
@@ -82,6 +85,61 @@ def add_arguments(self, parser, cli_name):
...
+        durability_transient_local_endpoints_count = 0
+
+        pubs_info = node.get_publishers_info_by_topic(args.topic_name)
+        publishers_count = len(pubs_info)
+        if publishers_count == 0:
+            return qos_profile
+
+        for info in pubs_info:
+            if (info.qos_profile.reliability ==
+                    QoSReliabilityPolicy.RELIABLE):
```
jacobperron, line 109: nit: I think this will fit on one line, staying under our 100 character limit.

ros2topic/ros2topic/verb/echo.py
```diff
@@ -82,6 +85,61 @@ def add_arguments(self, parser, cli_name):
...
+                reliability_reliable_endpoints_count += 1
+            if (info.qos_profile.durability ==
+                    QoSDurabilityPolicy.TRANSIENT_LOCAL):
```
jacobperron, line 112: Same here

ros2topic/ros2topic/verb/echo.py
```diff
@@ -82,6 +85,61 @@ def add_arguments(self, parser, cli_name):
...
+                durability_transient_local_endpoints_count += 1
+
+        # If all endpoints are reliable, ask for reliable
+        if reliability_reliable_endpoints_count == publishers_count:
+            qos_profile.reliability = QoSReliabilityPolicy.RELIABLE
+        else:
+            if reliability_reliable_endpoints_count > 0:
+                print(
+                    'Some, but not all, publisher are offering '
```
jacobperron, line 121:
```suggestion
                    'Some, but not all, publishers are offering '
```

ros2topic/ros2topic/verb/echo.py
```diff
@@ -82,6 +85,61 @@ def add_arguments(self, parser, cli_name):
...
+                )
+            qos_profile.reliability = QoSReliabilityPolicy.BEST_EFFORT
+
+        # If all endpoints are transient_local, ask for transient_local
+        if durability_transient_local_endpoints_count == publishers_count:
+            qos_profile.durability = QoSDurabilityPolicy.TRANSIENT_LOCAL
+        else:
+            if durability_transient_local_endpoints_count > 0:
+                print(
+                    'Some, but not all, publisher are offering '
```
jacobperron, line 134:
```suggestion
                    'Some, but not all, publishers are offering '
```

ros2topic/ros2topic/api/__init__.py
```diff
@@ -164,38 +166,39 @@ def qos_profile_from_short_keys(
 ) -> rclpy.qos.QoSProfile:
     """Construct a QoSProfile given the name of a preset, and optional overrides."""
     # Build a QoS profile based on user-supplied arguments
+    if not preset_profile:
+        preset_profile = default_preset_str
```
jacobperron, line 170: Since the default value seems to depend on the verb, I think it would be better to handle this case in `echo.py` where we can pass a default of "sensor_data" if `args.qos_profile` is `None`.

ros2topic/ros2topic/api/__init__.py
```diff
@@ -164,38 +166,39 @@ def qos_profile_from_short_keys(
...
     profile = rclpy.qos.QoSPresetProfiles.get_from_short_key(preset_profile)
     if history:
         profile.history = rclpy.qos.QoSHistoryPolicy.get_from_short_key(history)
     if durability:
         profile.durability = rclpy.qos.QoSDurabilityPolicy.get_from_short_key(durability)
     if reliability:
         profile.reliability = rclpy.qos.QoSReliabilityPolicy.get_from_short_key(reliability)
-    if depth >= 0:
+    if depth and depth >= 0:
```
jacobperron, line 178: Same here. I think we should handle the `None` case outside of this function. Alternatively, we should update the signature of this function:

```python
depth: Optional[int] = None
```

ros2topic/ros2topic/api/__init__.py
```diff
@@ -164,38 +166,39 @@ def qos_profile_from_short_keys(
...
     """Extend an existing ArgumentParser to allow input of QoS policy overrides."""
     verb = 'publish' if is_publisher else 'subscribe'
     parser.add_argument(
         '--qos-profile',
         choices=rclpy.qos.QoSPresetProfiles.short_keys(),
-        default=default_preset,
         help='Quality of service preset profile to {} with (default: {})'
              .format(verb, default_preset))
     default_profile = rclpy.qos.QoSPresetProfiles.get_from_short_key(
         default_preset)
```
jacobperron, line 199: Technically, the default values reported for the `--qos-reliability` and `--qos-durability` options below are incorrect for `ros2 topic echo`. By default, we are automatically matching existing publishers. Maybe we should add separate parser options for `echo` instead of reusing them between verbs.

jacobperron (commented):
Mostly minor comments; I have some concerns about reusing the argparse logic between the `pub` and `echo` verbs. The implementation of `choose_qos` LGTM.

ros2topic/ros2topic/verb/echo.py
```diff
@@ -82,6 +85,61 @@ def add_arguments(self, parser, cli_name):
         parser.add_argument(
             '--raw', action='store_true', help='Echo the raw binary representation')
 
+    def choose_qos(self, node, args):
+
+        if (args.qos_profile or args.qos_reliability or args.qos_durability or
+                args.qos_depth or args.qos_history):
```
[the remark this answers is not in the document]
gonzodepedro: Addressed on [160fce7](https://github.com/ros2/ros2cli/pull/613/commits/160fce7d4411ce53251d12ce92dd55b8e77fc046)

ros2topic/ros2topic/verb/echo.py
```diff
@@ -82,6 +85,61 @@ def add_arguments(self, parser, cli_name):
...
+            return qos_profile_from_short_keys(args.qos_profile,
+                                               reliability=args.qos_reliability,
+                                               durability=args.qos_durability,
+                                               depth=args.qos_depth,
+                                               history=args.qos_history)
```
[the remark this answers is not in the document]
gonzodepedro: Addressed on [160fce7](https://github.com/ros2/ros2cli/pull/613/commits/160fce7d4411ce53251d12ce92dd55b8e77fc046)

ros2topic/ros2topic/verb/echo.py
gonzodepedro: Addressed on [160fce7](https://github.com/ros2/ros2cli/pull/613/commits/160fce7d4411ce53251d12ce92dd55b8e77fc046)

ros2topic/ros2topic/verb/echo.py
gonzodepedro: Addressed on [160fce7](https://github.com/ros2/ros2cli/pull/613/commits/160fce7d4411ce53251d12ce92dd55b8e77fc046)

ros2topic/ros2topic/verb/echo.py
gonzodepedro: Addressed on [160fce7](https://github.com/ros2/ros2cli/pull/613/commits/160fce7d4411ce53251d12ce92dd55b8e77fc046)

ros2topic/ros2topic/verb/echo.py
gonzodepedro: Addressed on [160fce7](https://github.com/ros2/ros2cli/pull/613/commits/160fce7d4411ce53251d12ce92dd55b8e77fc046)

ros2topic/ros2topic/api/__init__.py
gonzodepedro: Addressed on [160fce7](https://github.com/ros2/ros2cli/pull/613/commits/160fce7d4411ce53251d12ce92dd55b8e77fc046)

ros2topic/ros2topic/api/__init__.py
gonzodepedro: Addressed on [160fce7](https://github.com/ros2/ros2cli/pull/613/commits/160fce7d4411ce53251d12ce92dd55b8e77fc046)

ros2topic/ros2topic/api/__init__.py
gonzodepedro: Addressed on [160fce7](https://github.com/ros2/ros2cli/pull/613/commits/160fce7d4411ce53251d12ce92dd55b8e77fc046)

ros2topic/ros2topic/verb/echo.py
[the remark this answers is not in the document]
jacobperron: The style didn't change here.

ros2topic/ros2topic/verb/echo.py
```diff
@@ -82,6 +111,65 @@ def add_arguments(self, parser, cli_name):
...
             '--raw', action='store_true', help='Echo the raw binary representation')
 
+    def choose_qos(self, node, args):
+
+        if (args.qos_profile is not None or
+            args.qos_reliability is not None or
+            args.qos_durability is not None or
+            args.qos_depth is not None or
+            args.qos_history is not None
+            ):
```
jacobperron, line 121:
```suggestion
        ):
```

jacobperron (approved):
There a linter error that my suggestion above should fix. LGTM with green CI.

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14555/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9312/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12220/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14677/)

jacobperron (approved):

ros2topic/ros2topic/api/__init__.py
```diff
@@ -171,49 +170,10 @@ def qos_profile_from_short_keys(
...
         profile.depth = depth
     else:
         if (profile.durability == rclpy.qos.QoSDurabilityPolicy.TRANSIENT_LOCAL
                 and profile.depth == 0):
             profile.depth = 1
-
     return profile
-
-
-def add_qos_arguments_to_argument_parser(
```
jacobperron, line 184: I've found an instance of this function being used in the wild. In the future, we should be mindful of removing API from this file since it appears as part of the public API (unless we otherwise state that users should't rely on it).

PR #616, opened 2021-03-31
Fix Topic Info Test with "Infinite" printing

emersonknapp:
Connected to https://github.com/ros2/rclpy/pull/722

Fixes the output tests given the new printing info.

### DIFF
--- a/ros2topic/test/test_cli.py
+++ b/ros2topic/test/test_cli.py
@@ -30,7 +30,6 @@ import launch_testing.actions
 import launch_testing.asserts
 import launch_testing.markers
 import launch_testing.tools
-import launch_testing.tools.text
 import launch_testing_ros.tools
 
 import pytest
@@ -189,7 +188,6 @@ class TestROS2TopicCLI(unittest.TestCase):
             output_filter=rmw_implementation_filter
         )
 
-    @unittest.skip('poop')
     @launch_testing.markers.retry_on_failure(times=5, delay=1)
     def test_list_topics(self):
         with self.launch_topic_command(arguments=['list']) as topic_command:
@@ -211,7 +209,6 @@ class TestROS2TopicCLI(unittest.TestCase):
             strict=True
         )
 
-    @unittest.skip('poop')
     @launch_testing.markers.retry_on_failure(times=5, delay=1)
     def test_list_all_topics(self):
         with self.launch_topic_command(
@@ -236,7 +233,6 @@ class TestROS2TopicCLI(unittest.TestCase):
             strict=True
         )
 
-    @unittest.skip('poop')
     @launch_testing.markers.retry_on_failure(times=5, delay=1)
     def test_list_with_types(self):
         with self.launch_topic_command(arguments=['list', '-t']) as topic_command:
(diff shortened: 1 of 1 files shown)


emersonknapp:
@ivanpauno One a side note - do you know a better way to debug when `launch_testing.tools.expect_output` is failing? I ended up having to write my own little loop in the test to figure out what line was the problem, it looked like:

```
actual_lines = topic_command.output.splitlines()
for expected_line, actual_line in zip(expected_lines, actual_lines):
    if hasattr(expected_line, 'match'):
        assert expected_line.match(actual_line)
    else:
        assert expected_line == actual_line
```

(
additionally, do you know if it is possible to isolate these tests? the following selects nothing

`colcon test --packages-select ros2topic --event-handlers console_direct+ --pytest-args -k test_topic_endpoint_info_verbose`

The only option I could get to work was `-k test_cli` which only narrows it down to that whole source file, which still takes potentially several minutes to run
)


None of the above is important - this should fix the tests, but if you have insights on the development workflow I would be really happy to hear it.

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14178/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8975/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11859/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14278/)

ivanpauno:
> @ivanpauno One a side note - do you know a better way to debug when launch_testing.tools.expect_output is failing? I ended up having to write my own little loop in the test to figure out what line was the problem, it looked like:

No, I always modify the test to manually capture the output and print it, or sth like that.
It would be good to improve `launch_testing.tools.expect_output` so it provides a more useful error message.

> colcon test --packages-select ros2topic --event-handlers console_direct+ --pytest-args -k test_topic_endpoint_info_verbose

For cmake packages I use `--cmake-args -R <regex_expression_tests_to_run>`.
For python I typically go to the test folder and use `python -m pytest -s <test_name> ...`

Unfortunately for `launch_testing` tests, I don't think there's a way to only run some of the test cases.
Based on the pytest output:

> collected 1 item / 1 deselected                                                                                                                                                                            

it seems that `launch_testing` collects everything in one test case (?).

@hidmic may know better.

ivanpauno (approved):
LGTM with green CI

ivanpauno:
CI results are [here](https://github.com/ros2/rclpy/pull/722#issuecomment-813385290).
Rpr checker is expected to fail until https://github.com/ros2/rclpy/pull/722 gets in (and a new rclpy release is done).

hidmic:
> It would be good to improve launch_testing.tools.expect_output so it provides a more useful error message.

Agreed. It's not the first time someone has complained about this --> https://github.com/ros2/system_tests/issues/404

> Unfortunately for launch_testing tests, I don't think there's a way to only run some of the test cases.

There isn't. `launch_testing` doesn't fully integrate with `pytest`. I cannot: https://github.com/ros2/launch/issues/405

audrow (approved):
Looks good to me.

hidmic (approved):

PR #618, opened 2021-04-02
Depend on rclcpp::rclcpp target

audrow:
Part of ros2/rclcpp#1613, https://github.com/ros-visualization/rqt/pull/246, and ament/googletest#15.

### DIFF
--- a/ros2lifecycle_test_fixtures/CMakeLists.txt
+++ b/ros2lifecycle_test_fixtures/CMakeLists.txt
@@ -23,6 +23,7 @@ add_executable(simple_lifecycle_node
   src/simple_lifecycle_node.cpp)
 target_link_libraries(simple_lifecycle_node
   ${rclcpp_lifecycle_LIBRARIES}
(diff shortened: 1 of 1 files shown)


Karsten1987 (changes requested):
I actually think this should be refactored.
You generally don't have to deal with `include_directories` or even direct cmake targets such as `rclcpp::rclcpp`. 

you can get rid of all these calls by simply leveraging `ament_target_dependencies(simple_lifecycle_node rclpp_lifecycle)`.

Also, I think you should not need to depend on `rclcpp`. `rclcpp_lifecycle` should technically export `rclcpp` and `ament_target_dependencies` ought to take care of the transitive deps.

wjwwood:
I agree with @Karsten1987 let's try the `ament_target_dependencies` route.

Karsten1987 (approved):
This is how I had envisioned it. 👍

wjwwood (approved):

PR #623, opened 2021-04-16
continue to next iteration after exceptions in generate_reports and

alsora:
This PR fixes a bug in the `ros2 doctor` utility.

Both the `run_checks` and the `generate_reports` functions contain a for loop that iterates over all the specified entry points.
If there are issues with one of the entry points, the exception is catched and a warning is printed.

However, after printing the warning, the loop continues executing.
This causes a problem as variables such as `check_class`, `report_class`, `check_instance` and `report_instance` may still hold the value assigned during the previous iteration of the loop.

Let's assume that we have 2 reports: A and B, with B raising an `ImportError` during `report_entry_pt.load()`.
This will result in printing twice the report from A.

This PR adds a `continue` after the warnings.

Signed-off-by: Alberto Soragna <alberto.soragna@gmail.com>

### DIFF
--- a/ros2doctor/ros2doctor/api/__init__.py
+++ b/ros2doctor/ros2doctor/api/__init__.py
@@ -95,10 +95,12 @@ def run_checks(*, include_warnings=False) -> Tuple[Set[str], int, int]:
             check_class = check_entry_pt.load()
         except ImportError:
             doctor_warn(f'Check entry point {check_entry_pt.name} fails to load.')
+            continue
         try:
             check_instance = check_class()
         except Exception:
             doctor_warn(f'Unable to instantiate check object from {check_entry_pt.name}.')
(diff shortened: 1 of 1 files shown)


clalancette (approved):
Your reasoning and the patch make perfect sense to me.  I'll run CI on this next.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14392/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9166/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12067/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14490/)

audrow (approved):

Issue #624, opened 2021-04-16
`ros2 topic pub` doesn't publish until after first period has expired


clalancette:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04, arm64
- Installation type:
  - Debian packages
- Version or commit hash:
  - ROS Rolling, package version 0.13.0-1
- DDS implementation:
  - CycloneDDS
- Client library (if applicable):
  - rclpy

#### Steps to reproduce issue

In terminal 1, run:
```
ros2 topic echo /talker std_msgs/msg/String
```

In terminal 2, run:
```
ros2 topic pub -r 0.1 /talker std_msgs/String "data: Hello World"
```
(that is, publish once every 10 seconds)

#### Expected behavior
The `ros2 topic pub` command would publish, then sleep for 10 seconds, then publish, sleep for 10 seconds, etc.

#### Actual behavior
The `ros2 topic pub` command sleeps for 10 seconds, then publishes, then sleeps for 10 seconds, then publishes, etc.

#### Additional information

You could argue that the current way is not wrong, but I think it is unexpected behavior for the user.

fujitatomoya:
as far as i know, the current behavior for `once` and `rate` is consistent with ROS1 `rostopic` command. besides, if `rate` is specified by user, probably frequency is what user cares, not starting timepoint.

but one use case that i think of is, if frequency is long enough (like 10 seconds), probably waiting that time window until 1st message would be not really useful.

clalancette:
> but one use case that i think of is, if frequency is long enough (like 10 seconds), probably waiting that time window until 1st message would be not really useful.

Yeah, that's exactly where I ran into it.  My example above is a very long time period, and thus the initial message doesn't get published until after the first 10 seconds.

This honestly isn't a huge deal; I doubt too many people are using a 0.1 Hz rate.  I still think it should be fixed, but changing behavior at this point in Galactic is probably a bad idea.  I'll move it to the H-Turtle board instead.

NiklasBurggraaff:
I'm looking to get started with contributing to ROS, and I thought this might be a good task to start with. Should I just get started with working on making this change and submit a PR once I'm done?

clalancette:
> I'm looking to get started with contributing to ROS, and I thought this might be a good task to start with. Should I just get started with working on making this change and submit a PR once I'm done?

There is already a change proposed to fix this in #626.  One thing to do would be to go ahead and review that change and leave any comments you might have.

[...]

PR #626, opened 2021-04-19
`ros2 topic pub` starts publishing right away.

fujitatomoya:
address https://github.com/ros2/ros2cli/issues/624

### DIFF
--- a/ros2topic/ros2topic/verb/pub.py
+++ b/ros2topic/ros2topic/verb/pub.py
@@ -188,11 +188,16 @@ def publisher(
             print('publishing #%d: %r\n' % (count, msg))
         pub.publish(msg)
 
-    timer = node.create_timer(period, timer_callback)
-    while times == 0 or count < times:
-        rclpy.spin_once(node)
-
-    # give some time for the messages to reach the wire before exiting
-    time.sleep(keep_alive)
-
-    node.destroy_timer(timer)
+    # give some time for discovery process
+    time.sleep(0.1)
+    timer_callback()
+    if times != 1:
+        timer = node.create_timer(period, timer_callback)
+        while times == 0 or count < times:
+            rclpy.spin_once(node)
+        # give some time for the messages to reach the wire before exiting
+        time.sleep(keep_alive)
+        node.destroy_timer(timer)
+    else:
+        # give some time for the messages to reach the wire before exiting
+        time.sleep(keep_alive)


ros2topic/ros2topic/verb/pub.py
```diff
@@ -155,11 +155,16 @@ def timer_callback():
...
-
-    # give some time for the messages to reach the wire before exiting
-    time.sleep(keep_alive)
-
-    node.destroy_timer(timer)
+    # give some time for discovery process
+    time.sleep(0.1)
+    timer_callback()
+    if times != 1:
+        timer = node.create_timer(period, timer_callback)
```
fujitatomoya, line 162: Or we could have this option in rclcpp and rclpy `Timer` class? probably not, since user can do the same in application code just like this change.

fujitatomoya:
@clalancette @ivanpauno what do you think? i'd like to hear your opinion.

clalancette:
I think we should punt on this for now; the issue isn't that important and I don't think we should change it for Galactic.  I'll add this to H-Turtle board instead.

ros2topic/ros2topic/verb/pub.py
audrow: If I understand correctly, if we want to make the timer fire immediately, we should make that change in `rcl`. That being said, maybe this change helps `ros2 topic pub` more in line with a user's expectations.

audrow (approved):
The change looks good and I tested that it worked on my machine. 

I think maybe the true solution is to change the timer behavior in `rcl`, but this might be an easier way to make `ros2 topic pub` perform how the user expects.

audrow:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14466/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9234/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12135/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14567/)

ivanpauno:
> I think maybe the true solution is to change the timer behavior in rcl

I don't think that's a good idea.
e.g.: if you want a "one-shot timer", what you currently do is to register a timer and cancel it from the callback.
If the first call happens immediatly that wouldn't work.

fujitatomoya:
> e.g.: if you want a "one-shot timer", what you currently do is to register a timer and cancel it from the callback.

yeah i am with @ivanpauno on this. application can have more control for use cases.

audrow:
Rerunning CI to double check that things are good:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14588/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9343/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12252/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14717/)

fujitatomoya:
I believe that rebase required to include https://github.com/ros2/ros2cli/pull/616, retry CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14590/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9345/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12254/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14720/)

ivanpauno (approved):

ivanpauno:

ivanpauno:
@fujitatomoya could you confirm if the Rpr failure is unrelated or not?

Besides that failure, this seems to be ready to be merged.

fujitatomoya:
@ivanpauno thanks! 

https://build.ros2.org/job/Rpr__ros2cli__ubuntu_focal_amd64/223/testReport/ros2topic.ros2topic.test/test_cli/test_cli/ is not related to the fix. it is failure of `ros2 topic info` not `pub`, i do not see the relationship from the test implementation. besides, i cannot reproduce this failure on my local environment.

(Just FYI, https://build.ros2.org/job/Rpr__ros2cli__ubuntu_focal_amd64/223/testReport/ros2topic.ros2topic.test/test_cli/test_cli/ does have the expected output which is `Type: std_msgs/msg/String\n\nPublisher count: 1\n\nNode name: talker_node\nNode namespace: /\nTopic type: std_msgs/ms...ds\n  Liveliness: AUTOMATIC\n  Liveliness lease duration: 9223372036854775807 nanoseconds\n\nSubscription count: 0\n\n`. i do not really see why this is failing.)

ivanpauno:
Ah, I know the problem

> Liveliness lease duration: 9223372036854775807

https://github.com/ros2/rclpy/pull/722, that change wasn't released into rolling when the Rpr builder last run (and the tests here were already updated).

It should pass now:

@ros-pull-request-builder retest this please.

fujitatomoya:
@ivanpauno thanks for the support!

PR #627, opened 2021-04-19
[ros2pkg] Use underscores in setup.cfg.em instead of dashes

ivanpauno:
Get rid of these warnings:

```
c:\python38\lib\site-packages\setuptools\dist.py:642: UserWarning: Usage of dash-separated 'script-dir' will not be supported in future versions. Please use the underscore name 'script_dir' instead
warnings.warn(
c:\python38\lib\site-packages\setuptools\dist.py:642: UserWarning: Usage of dash-separated 'install-scripts' will not be supported in future versions. Please use the underscore name 'install_scripts' instead
warnings.warn(
--- stderr: my_package
c:\python38\lib\site-packages\setuptools\dist.py:642: UserWarning: Usage of dash-separated 'script-dir' will not be supported in future versions. Please use the underscore name 'script_dir' instead
warnings.warn(
c:\python38\lib\site-packages\setuptools\dist.py:642: UserWarning: Usage of dash-separated 'install-scripts' will not be supported in future versions. Please use the underscore name 'install_scripts' instead
warnings.warn(
```

e.g. when running the [create your first package](https://docs.ros.org/en/rolling/Tutorials/Creating-Your-First-ROS2-Package.html) tutorial.

### DIFF
--- a/ros2pkg/ros2pkg/resource/ament_python/setup.cfg.em
+++ b/ros2pkg/ros2pkg/resource/ament_python/setup.cfg.em
@@ -1,4 +1,4 @@
 [develop]
-script-dir=$base/lib/@(project_name)
+script_dir=$base/lib/@(project_name)
 [install]
-install-scripts=$base/lib/@(project_name)
+install_scripts=$base/lib/@(project_name)


clalancette:
We can't do this right now, as far as I understand it.  The problem is that while newer setuptools has this warning, setuptools 45 (which is what is in Ubuntu Focal) doesn't support the underscore separated keys.  Thus, if we were to merge this, we would break all of the debian packages.

The best we can do is wait until Ubuntu 22.04 and then change them all at once.

ivanpauno:
> We can't do this right now, as far as I understand it. The problem is that while newer setuptools has this warning, setuptools 45 (which is what is in Ubuntu Focal) doesn't support the underscore separated keys. Thus, if we were to merge this, we would break all of the debian packages.
> 
> The best we can do is wait until Ubuntu 22.04 and then change them all at once.

Oh, that's unfortunate.

clalancette:
> Oh, that's unfortunate.

Actually, maybe I'm wrong: https://github.com/pypa/setuptools/blob/v45.2.0/setuptools/dist.py .  I'm not sure.

If we are going to do this, I'll suggest testing it out on a leaf package somehow and making sure it still works as an Ubuntu package.

ivanpauno:
> Actually, maybe I'm wrong: https://github.com/pypa/setuptools/blob/v45.2.0/setuptools/dist.py . I'm not sure.
> 
> If we are going to do this, I'll suggest testing it out on a leaf package somehow and making sure it still works as an Ubuntu package.

Yes, I was double checking that locally.
I uninstalled the pip installed setuptools from my docker development environment, checked that I was using the system one (v45.2.0) and build `demo_nodes_py` with https://github.com/ros2/demos/pull/502 applied. It worked perfectly.

I think that we should get at least this PR in, because it affect a tutorial.
The other ones can wait after the freeze, but I don't think we need to wait until Ubuntu 22.04.

hidmic (approved):
LGTM provided we're sure it works with the `setuptools` version distributed in Focal.

clalancette:
> LGTM provided we're sure it works with the `setuptools` version distributed in Focal.

Yeah, we're sure it does now.  We released demo_nodes_cpp onto the buildfarm with https://github.com/ros2/demos/pull/502 merged, and all seems happy there.  So I think we're good to go here.

clalancette (approved):

ivanpauno:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14420/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9193/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12094/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14519/)

PR #404, opened 2019-11-19
Update network check to fix missing flags on Windows 

claireyywang:
Warn about missing flags on windows `ipconfig` output. Closes #403 - false negative warnings on windows.

PR #631, opened 2021-04-21
Improve ros2 doctor on Windows

ivanpauno:
The output of `ros2 doctor` on Windows was confusing.

* Print an useful error message when failing to import rosdistro.
* Print a clearer error message when failing to find either a loopback or no loopback interface on Windows.
* Print "latest version" instead of "required version".
* Print the corrent name of the python executable according to the platform.

See the comment I wrote in the Galactic testing matric for more context.

`rosdistro` python package is not usually installed on Windows, so having a nicer message when not found is helpful.
I updated "latest version" to "required version" because the warning shows that for that ros distribution there's a newer version of the package available in rosdistro, it's not really a required version.

### DIFF
--- a/ros2doctor/ros2doctor/api/network.py
+++ b/ros2doctor/ros2doctor/api/network.py
@@ -13,6 +13,8 @@
 # limitations under the License.
 
 import os
+from pathlib import Path
+import sys
 from typing import Tuple
 
 from ros2doctor.api import DoctorCheck
@@ -27,7 +29,7 @@ try:
 except ImportError:  # check import error for windows and osx
     doctor_warn(
         'Unable to import ifcfg. '
-        'Use `python3 -m pip install ifcfg` to install needed package.')
+        f'Use `{Path(sys.executable).name} -m pip install ifcfg` to install needed package.')
 
 
 def _is_unix_like_platform() -> bool:
@@ -74,11 +76,10 @@ class NetworkCheck(DoctorCheck):
         if not _is_unix_like_platform():
             if not has_loopback and not has_non_loopback:
                 # no flags found, otherwise one of them should be True.
-                doctor_error(
-                    'No flags found. '
-                    'Run `ipconfig` on Windows or '
-                    'install `ifconfig` on Unix to check network interfaces.')
-                result.add_error()
+                doctor_warn(
+                    'Did not found either a loopback or no loopback interface.'
+                    'Run `ipconfig` on Windows to check what interfaces are available')
+                result.add_warning()
                 return result
         if not has_loopback:
             doctor_error('No loopback IP address is found.')
--- a/ros2doctor/ros2doctor/api/network.py
+++ b/ros2doctor/ros2doctor/api/network.py
@@ -13,9 +13,9 @@
 # limitations under the License.
 
 import os
-from pathlib import Path
-import sys
 from typing import Tuple
+
+import ifcfg
 
 from ros2doctor.api import DoctorCheck
 from ros2doctor.api import DoctorReport
@@ -23,13 +23,6 @@ from ros2doctor.api import Report
 from ros2doctor.api import Result
 from ros2doctor.api.format import doctor_error
 from ros2doctor.api.format import doctor_warn
-
-try:
-    import ifcfg
-except ImportError:  # check import error for windows and osx
-    doctor_warn(
-        'Unable to import ifcfg. '
-        f'Use `{Path(sys.executable).name} -m pip install ifcfg` to install needed package.')
 
 
 def _is_unix_like_platform() -> bool:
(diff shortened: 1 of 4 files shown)


ros2doctor/ros2doctor/api/package.py
```diff
@@ -26,7 +28,12 @@
 from ros2doctor.api.format import doctor_error
 from ros2doctor.api.format import doctor_warn
 
-import rosdistro
+try:
+    import rosdistro
+except ImportError:
+    doctor_warn(
+        'Unable to import rosdistro. '
+        f'Use `{Path(sys.executable).name} -m pip install rosdistro` to install needed package.')
```
clalancette, line 36: I'm not really sure about this bit.  If `rosdistro` is a pip package that can be installed on Windows, then I feel like that should just be part of our installation instructions.  If it can't be installed on Windows, then I think we just want to skip this stuff.  Having this halfway place where we warn if certain things aren't installed seems weird to me.

clalancette (commented):
A couple of comments in here.

ros2doctor/ros2doctor/api/network.py
```diff
@@ -74,11 +76,10 @@ def check(self):
...
             if not has_loopback and not has_non_loopback:
                 # no flags found, otherwise one of them should be True.
-                doctor_error(
-                    'No flags found. '
-                    'Run `ipconfig` on Windows or '
-                    'install `ifconfig` on Unix to check network interfaces.')
-                result.add_error()
+                doctor_warn(
+                    'Did not found either a loopback or no loopback interface.'
+                    'Run `ipconfig` on Windows to check what interfaces are available')
```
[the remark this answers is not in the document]
ivanpauno: > Separately, why is there no similar warning for Unix-like platforms?

I think that on Unix if you get that output here you actually don't have a network interface.
On windows this check sometimes fails when you have a network interface available, because the `ifcfg` package doesn't populate the interface "flags" when listing interfaces.

I applied your suggestion though actually the issue isn't than an interface wasn't found, but that we cannot recognize which one is a loopback interface and which one not.

See also https://github.com/ros2/ros2cli/pull/404.

ros2doctor/ros2doctor/api/network.py
```diff
@@ -74,11 +76,10 @@ def check(self):
         if not _is_unix_like_platform():
             if not has_loopback and not has_non_loopback:
                 # no flags found, otherwise one of them should be True.
-                doctor_error(
-                    'No flags found. '
-                    'Run `ipconfig` on Windows or '
-                    'install `ifconfig` on Unix to check network interfaces.')
-                result.add_error()
+                doctor_warn(
+                    'Could not find any available network interfaces.'
```
christophebedard, line 80: perhaps you were going to change the message again (https://github.com/ros2/ros2cli/pull/631#discussion_r618413015), but just in case :smile: 

```suggestion
                    'Could not find any available network interfaces. '
```

ivanpauno: thanks!

ros2doctor/ros2doctor/api/package.py
hidmic: So, the way `ros2doctor` implements "optional" checks is odd. I'll ignore the fact that all dependencies are listed in the `package.xml`, which we generally assume to be satisfied. Why replicating checks everywhere, instead of failing entry point load with a meaningful message if dependencies are not met?

ivanpauno: > Why replicating checks everywhere, instead of failing entry point load with a meaningful message if dependencies are not met?

Yes, that's a good point.

See https://github.com/ros2/ros2cli/pull/631/commits/cd5e980710c595875bcf00d108b623da45e4d8c4 and https://github.com/ros2/ros2cli/pull/631/commits/fbf7510f248d57be25b07d5ef970d4742a96cccd.

ros2doctor/ros2doctor/api/network.py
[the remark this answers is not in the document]
ivanpauno: Modified message again, see latest changes.

hidmic (approved):
This looks a lot simpler :) LGTM !

ivanpauno:

ivanpauno:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14575/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9331/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12238/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14702/)

ivanpauno:
@Mergifyio backport galactic

PR #635, opened 2021-05-06
Generate changelogs

ivanpauno:
* Used `catking_generate_changelogs --all`.
* Did some automatic formating (start sentences with uppercase, end sentences with a dot, delete `Forthcoming` section).
* Cleaned-up all changelog items after Foxy (0.10.0 onwards).


ivanpauno:
I would recommend only reviewing this commit https://github.com/ros2/ros2cli/pull/635/commits/ed2e57e928e634ee26db3aa9af087d56d5cb1f35 and not looking at the other commits deeply (it's just an automatic generated changelog + formatting).

ivanpauno:
The motivation for generating the changelogs is https://github.com/ros2/ros2_documentation/issues/1465

clalancette (approved):
This looks great, thanks!

Issue #625, opened 2021-04-16
`ros2 param dump` silently overwrites existing file


clalancette:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- Installation type:
  - Debian packages
- Version or commit hash:
  - ros-rolling-ros2param 0.13.0-1
- DDS implementation:
  - CycloneDDS
- Client library (if applicable):
  - rclpy

#### Steps to reproduce issue

In terminal 1:
```
ros2 run turtlesim turtlesim_node
```

In terminal 2:
```
echo "foo" > turtlesim.yaml  # turtlesim.yaml now contains "foo"
ros2 param dump /turtlesim  # this overwrites the existing turtlesim.yaml, which may be surprising
```

#### Expected behavior

When `ros2 param dump` is choosing a filename, it should either error out if a file with the name it chose already exists, or it should keep choosing names until it finds one that has no conflict.

#### Actual behavior

`ros2 param dump` silently ovewrites the existing file.

hidmic:
Sometimes it is practical for it to just overwrite it. Perhaps making opt-in would be best.

clalancette:
> Sometimes it is practical for it to just overwrite it. Perhaps making opt-in would be best.

Yeah, I'd be fine with a `--force` or `--overwrite` flag to cause it to always overwrite.  I just think that the current behavior could lead to data loss, where I was storing something I actually wanted in `turtlesim.yaml`, and running `ros2 param dump /turtlesim` in the same directory just overwrites it.

sloretz:
What if the default were to write to stdout, with an option to write to a named file? Then overwriting is no big deal because the user had to give a filename to write to.

clalancette:
> What if the default were to write to stdout, with an option to write to a named file? Then overwriting is no big deal because the user had to give a filename to write to.

That's a pretty good idea, I like it.  This is also more Unix-y; you could always do `ros2 param dump /turtlesim > turtlesim.yaml`.  

The downside is that this breaks compatibility with earlier releases.  That is, in Foxy, if you do `ros2 param dump /turtlesim`, by default it writes to a file.  If you then pass `--print`, it prints to stdout.  If we were to make this change, then in Galactic `ros2 param dump /turtlesim` would print to stdout, and passing `--write <fname>` (or whatever) would write it to a file.

I don't know, how do people feel about a command-line break like that?

sloretz:
@clalancette ok to close now that #638 is merged?

[...]

PR #638, opened 2021-05-19
add '--write' option to avoid an unintentional data loss.

fujitatomoya:
address https://github.com/ros2/ros2cli/issues/625

### DIFF
--- a/ros2param/ros2param/verb/dump.py
+++ b/ros2param/ros2param/verb/dump.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import os
+import sys
 
 from rcl_interfaces.srv import ListParameters
 
@@ -50,10 +51,10 @@ class DumpVerb(VerbExtension):
         parser.add_argument(
             '--output-dir',
             default='.',
-            help='The absolute path were to save the generated file')
+            help='DEPRECATED: The absolute path were to save the generated file')
         parser.add_argument(
-            '--write', action='store_true',
-            help='Save parameters to a file, otherwise print to stdout in terminal.')
+            '--print', action='store_true',
+            help='DEPRECATED: Does nothing.')
 
     @staticmethod
     def get_parameter_value(node, node_name, param):
@@ -123,7 +124,16 @@ class DumpVerb(VerbExtension):
                     'Exception while calling service of node '
                     f"'{node_name.full_name}': {e}")
 
-            if not args.write:
+            if args.print:
+                print(
+                    "WARNING: '--print' is deprecated; print to stdout in terminal by default",
+                    file=sys.stderr)
+
+            if args.output_dir != '.':
+                print(
+                    "WARNING: '--output-dir' is deprecated; use redirection to save file",
+                    file=sys.stderr)
+            else:
                 print(yaml.dump(yaml_output, default_flow_style=False))
                 return
 
--- a/ros2param/ros2param/verb/dump.py
+++ b/ros2param/ros2param/verb/dump.py
@@ -52,8 +52,8 @@ class DumpVerb(VerbExtension):
             default='.',
             help='The absolute path were to save the generated file')
         parser.add_argument(
-            '--print', action='store_true',
-            help='Print generated file in terminal rather than saving a file.')
+            '--write', action='store_true',
+            help='Save parameters to a file, otherwise print to stdout in terminal.')
 
     @staticmethod
     def get_parameter_value(node, node_name, param):
@@ -123,7 +123,7 @@ class DumpVerb(VerbExtension):
                     'Exception while calling service of node '
                     f"'{node_name.full_name}': {e}")
 
-            if args.print:
+            if not args.write:
                 print(yaml.dump(yaml_output, default_flow_style=False))
                 return
 
--- a/ros2param/ros2param/verb/dump.py
+++ b/ros2param/ros2param/verb/dump.py
@@ -51,7 +51,7 @@ class DumpVerb(VerbExtension):
         parser.add_argument(
             '--output-dir',
             default='.',
-            help='DEPRECATED: The absolute path were to save the generated file')
+            help='DEPRECATED: The absolute path where to save the generated file')
         parser.add_argument(
             '--print', action='store_true',
             help='DEPRECATED: Does nothing.')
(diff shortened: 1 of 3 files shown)


fujitatomoya:
@clalancette @sloretz could you take a look?

ros2param/ros2param/verb/dump.py
```diff
@@ -52,8 +52,8 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
             default='.',
             help='The absolute path were to save the generated file')
         parser.add_argument(
-            '--print', action='store_true',
-            help='Print generated file in terminal rather than saving a file.')
```
clalancette, line 56: From a backwards-compatibility standpoint, I think we may want to leave the `--print` option, but just have it print a warning.  That way if someone has scripting where they call this, it will still work in H-Turtle, but it will throw a warning.

ros2param/ros2param/verb/dump.py
```diff
@@ -52,8 +52,8 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
...
+            '--write', action='store_true',
+            help='Save parameters to a file, otherwise print to stdout in terminal.')
```
clalancette, line 56: I'm wondering if we need the `--write` argument at all.  After all, the user can always redirect the output from stdout to a file (`ros2 param dump /foo > foo.yaml`), and that seems a lot more natural to me.

What do others think?

audrow: I agree. The redirect does seem much more natural.

ros2param/ros2param/verb/dump.py
fujitatomoya: makes sense.

ros2param/ros2param/verb/dump.py
fujitatomoya: that requires user to specify the directory w/o `--output-dir`, i think it is still fine by me. but what would you think? can we deprecate `--output-dir` as well?

artivis: Thanks for the heads up :+1: .
I personally don't mind the change but it is indeed somewhat breaking. If the initial issue is overwritten files, why not simply fail if the file exists (together with an `--force` flag)? It is still changing the current behavior, but less than this proposal.

ros2param/ros2param/verb/dump.py
artivis: Make sure that the warning would not turn the output into invalid yaml ;)

ros2param/ros2param/verb/dump.py
clalancette: > I personally don't mind the change but it is indeed somewhat breaking. If the initial issue is overwritten files, why not simply fail if the file exists (together with an `--force` flag)? It is still changing the current behavior, but less than this proposal.

We could also go that route.  My thinking was that in the long run, just always printing to stdout and letting the user use the shell redirection would be better.  So we can move that direction for H-Turtle, and then completely remove the old flags for I-Turtle.  After all, part of the reason for having distributions is so we can make "larger" changes like this.

artivis: I agree with the reasoning. I'm simply wondering how many people will curse us for this change, hopefully not many :D

fujitatomoya: @clalancette @artivis thanks for the comments.

i am good to go with https://github.com/ros2/ros2cli/pull/638#discussion_r636039660, fix coming up soon.

> Put a release note at https://docs.ros.org/en/rolling on the H-Turtle page about this change in behavior (we'll create that H-Turtle release note page early next week).

i will keep that in mind.

clalancette: > I agree with the reasoning. I'm simply wondering how many people will curse us for this change, hopefully not many :D

Yeah, I do appreciate you bringing up the point.  I think this is a very "leafy" utility, in that I don't expect it to be embedded in too many scripts.  I also think that mildly retraining people on how to use it isn't too onerous.  Finally, we aren't really changing the semantic of what is going on, just where it prints.  So I'm willing to take the risk here and get to where we want to be for H-Turtle.

fujitatomoya:
@clalancette @audrow friendly ping.

ros2param/ros2param/verb/dump.py
```diff
@@ -50,10 +51,10 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
         parser.add_argument(
             '--output-dir',
             default='.',
-            help='The absolute path were to save the generated file')
+            help='DEPRECATED: The absolute path were to save the generated file')
```
clalancette, line 54:
```suggestion
            help='DEPRECATED: The absolute path where to save the generated file')
```

ros2param/ros2param/verb/dump.py
```diff
@@ -124,6 +125,15 @@ def main(self, *, args):  # noqa: D102
                     f"'{node_name.full_name}': {e}")
 
             if args.print:
+                print(
+                    "WARNING: '--print' is deprecated; print to stdout in terminal by default",
```
clalancette, line 129:
```suggestion
                    "WARNING: '--print' is deprecated; this utility prints to stdout by default",
```

ros2param/ros2param/verb/dump.py
```diff
@@ -124,6 +125,15 @@ def main(self, *, args):  # noqa: D102
...
+                    file=sys.stderr)
+
+            if args.output_dir != '.':
+                print(
+                    "WARNING: '--output-dir' is deprecated; use redirection to save file",
```
clalancette, line 134:
```suggestion
                    "WARNING: '--output-dir' is deprecated; use redirection to save to a file",
```

clalancette (changes requested):
A few minor wording changes, and then I'll be happy to approve.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14715/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9467/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12382/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14860/)

clalancette (approved):
This change looks good to me now, thanks!  Can you please also add a note to https://github.com/ros2/ros2_documentation/blob/rolling/source/Releases/Release-Humble-Hawksbill.rst with this change?  Thanks.

audrow (approved):

clalancette:
I jumped the gun and merged the documentation PR before this one.  I'm going to go ahead and merge this one so the docs and the code are in sync.  Thanks @fujitatomoya for picking this up!

Issue #601, opened 2021-02-23
Ros2 run doesn't indicate a segfault when it occurs.


amussell:
## Bug report
ros2 run doesn't print that a segfault occurred when one occurs making it look like the program just exited.
**Required Info:**

- Operating System:
  Ubuntu 20.04
- Installation type:
  apt
- Version or commit hash:
  foxy
- DDS implementation:
  default
- Client library (if applicable):
  rclcpp

#### Steps to reproduce issue
Run a program that segfualts with ros2 run.

#### Expected behavior
It should print a segmentation fault (core dump) when a segfault occurs.

audrow:
I was able to confirm this behavior. 

@amussell, feel free to open a PR for this and ping me as a reviewer.

fujitatomoya:
@audrow @amussell 

https://github.com/ros2/ros2cli/pull/639 can fix the problem and it's been merged. i will close this issue, feel free to open another ticket if we still have the problem.

PR #639, opened 2021-05-20
check subprocess.returncode to print error message.

fujitatomoya:
address https://github.com/ros2/ros2cli/issues/601

### DIFF
--- a/ros2run/ros2run/api/__init__.py
+++ b/ros2run/ros2run/api/__init__.py
@@ -20,6 +20,8 @@ import sys
 from ros2pkg.api import get_executable_paths
 from ros2pkg.api import PackageNotFound
 
+ROS2RUN_MSG_PREFIX = '[ros2run]:'
+
 
 class MultipleExecutables(Exception):
 
@@ -70,10 +72,10 @@ def run_executable(*, path, argv, prefix=None):
     if process.returncode != 0:
         if -process.returncode in signal.valid_signals() and os.name == 'posix':
             # a negative value -N indicates that the child was terminated by signal N.
-            print(signal.strsignal(-process.returncode))
+            print(ROS2RUN_MSG_PREFIX, signal.strsignal(-process.returncode))
         else:
             # print general failure message instead.
-            print('Process exited with failure %d' % (process.returncode))
+            print(ROS2RUN_MSG_PREFIX, 'Process exited with failure %d' % (process.returncode))
     return process.returncode
 
 
--- a/ros2run/ros2run/api/__init__.py
+++ b/ros2run/ros2run/api/__init__.py
@@ -67,8 +67,13 @@ def run_executable(*, path, argv, prefix=None):
             # the subprocess will also receive the signal and should shut down
             # therefore we continue here until the process has finished
             pass
-    if process.returncode < 0:
-        print(signal.strsignal(-process.returncode))
+    if process.returncode != 0:
+        if os.name == 'posix':
+            # a negative value -N indicates that the child was terminated by signal N.
+            print(signal.strsignal(-process.returncode))
+        else:
+            # print general failure message instead.
+            print('Process exits failure')
     return process.returncode
 
 
--- a/ros2run/ros2run/api/__init__.py
+++ b/ros2run/ros2run/api/__init__.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import os
+import signal
 import subprocess
 import sys
 
@@ -66,6 +67,8 @@ def run_executable(*, path, argv, prefix=None):
             # the subprocess will also receive the signal and should shut down
             # therefore we continue here until the process has finished
             pass
+    if process.returncode < 0:
+        print(signal.strsignal(-process.returncode))
     return process.returncode
 
 
--- a/ros2run/ros2run/api/__init__.py
+++ b/ros2run/ros2run/api/__init__.py
@@ -68,12 +68,12 @@ def run_executable(*, path, argv, prefix=None):
             # therefore we continue here until the process has finished
             pass
     if process.returncode != 0:
-        if os.name == 'posix':
+        if -process.returncode in signal.valid_signals() and os.name == 'posix':
             # a negative value -N indicates that the child was terminated by signal N.
             print(signal.strsignal(-process.returncode))
         else:
             # print general failure message instead.
-            print('Process exits failure')
+            print('Process exited with failure %d' % (process.returncode))
     return process.returncode
 
 


ros2run/ros2run/api/__init__.py
```diff
@@ -66,6 +67,8 @@ def run_executable(*, path, argv, prefix=None):
             # the subprocess will also receive the signal and should shut down
             # therefore we continue here until the process has finished
             pass
+    if process.returncode < 0:
+        print(signal.strsignal(-process.returncode))
```
fujitatomoya, line 71: the alternative is we can wrap subprocess with `shell=True` but i just do not like to instantiate shell process only for this.

fujitatomoya:
@audrow could you take a look when you got time?

ros2run/ros2run/api/__init__.py
clalancette: I don't think we should use `shell=True` either.

As far as the code here, can `returncode` ever be `< 0`?  In Linux, at least, return codes are always 0 or greater.

fujitatomoya: yeah, it is python spec which you can see https://docs.python.org/3/library/subprocess.html#subprocess.CompletedProcess.returncode.

`returncode` will be negative if the child process is terminated with signal. (e.d `Segmentation Fault` will be `-11`)

clalancette: > yeah, it is python spec which you can see https://docs.python.org/3/library/subprocess.html#subprocess.CompletedProcess.returncode.

Oh!  I didn't realize that, thanks for pointing it out.

The documentation there says:

```
A negative value -N indicates that the child was terminated by signal N (POSIX only).
```

So I wonder what happens on Windows.

audrow (approved):
Looks like a good change to me. Is there a way to add a test for it?

ros2run/ros2run/api/__init__.py
fujitatomoya: that is actually nice question, i am not really sure. (dont have environment.) does anyone happen to know about this?
CC: @iuhilnehc-ynos @Barry-Xu-2018

Barry-Xu-2018: @fujitatomoya  @clalancette 

I check return code on window.  
Run ` ros2 run demo_nodes_cpp talker` and terminate `talker.exe` in windows task manager.  
The return code is `1`.   

According to wiki [exit status for window](https://en.wikipedia.org/wiki/Exit_status#Windows), it provides a reference to [Windows System Error Codes (exit codes)](http://www.hiteksoftware.com/knowledge/articles/049.htm). In this document, exit codes are  all  greater than 0 if error occurs or terminated by other process.

fujitatomoya: thanks @Barry-Xu-2018 

then we can do the following? any suggestions are welcome.

1. check if returcode is NOT zero?
2. if POSIX, call strsignal. and else print general message that subprocess exited with failure.
3. return `returncode`

clalancette: I also did a bit of testing on Windows.  I wrote this silly code:

```
#include <iostream>

int main()
{
    std::cout << "Hello, world" << std::endl;
    *(char *)0 = 0;
    return 0;
}
```

Which obviously causes a crash.  If I compile and run that program on the command-line, I get a negative error code:

```
C:\dev>cl /EHsc crash.cpp
Microsoft (R) C/C++ Optimizing Compiler Version 19.28.29336 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

crash.cpp
Microsoft (R) Incremental Linker Version 14.28.29336.0
Copyright (C) Microsoft Corporation.  All rights reserved.

/out:crash.exe
crash.obj

C:\dev>crash
Hello, world

C:\dev>echo %errorlevel%
-1073741819
```

However, I get a different result if I run it through subprocess on Python.  Here is my simple Python code:

```
import subprocess

process = subprocess.Popen(['crash'])
process.communicate()
print(process.returncode)
```

When I run that, I get:

```
C:\dev>python subprocess-crash.py
Hello, world
3221225477

```

So it looks like Python is wrapping the negative return code or something like that.

Regardless, I think @fujitatomoya 's proposal sounds like an improvement over the current situation, so :+1: from me.

fujitatomoya:
@clalancette i'd like to request a review.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14703/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9455/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12368/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14847/)

ros2run/ros2run/api/__init__.py
```diff
@@ -66,6 +67,13 @@ def run_executable(*, path, argv, prefix=None):
             # the subprocess will also receive the signal and should shut down
             # therefore we continue here until the process has finished
             pass
+    if process.returncode != 0:
+        if os.name == 'posix':
+            # a negative value -N indicates that the child was terminated by signal N.
+            print(signal.strsignal(-process.returncode))
```
clalancette, line 73: If the process doesn't crash, but returns a non-zero exit code, this ends up with:

```
ValueError: signal number out of range
```

So we'll have to handle that differently.

ros2run/ros2run/api/__init__.py
```diff
@@ -66,6 +67,13 @@ def run_executable(*, path, argv, prefix=None):
...
+        else:
+            # print general failure message instead.
+            print('Process exits failure')
```
clalancette, line 76:
```suggestion
            print('Process exited with failure %d' % (process.returncode))
```

clalancette (changes requested):
Besides the comments inline, I'm thinking we should have some prefix to show that these messages came from `ros2run`, and not from the process itself.  Maybe something like "[ros2run]: Message".

ros2run/ros2run/api/__init__.py
fujitatomoya: Ah i see, nice catch 👍 thanks. i will do check valid signal range before that.

fujitatomoya:
@clalancette all comments are addressed, i will run CI.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14716/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9468/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12383/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14861/)

fujitatomoya:
@clalancette friendly ping.

clalancette:
> Besides the comments inline, I'm thinking we should have some prefix to show that these messages came from `ros2run`, and not from the process itself. Maybe something like "[ros2run]: Message".

This comment is still open.  I do think we should have some kind of prefix on the output so people can tell it comes from `ros2run` and not from their binary.

fujitatomoya:
@clalancette my bad, i missed that. i will address it.

fujitatomoya:
unit test to confirm the message. (using simple example to generate segmentation fault)

https://github.com/fujitatomoya/ros2_test_prover/blob/master/prover_rclcpp/src/ros2cli_601.cpp

```
root@134f29e8f25f:~/ros2_ws/colcon_ws# ros2 run prover_rclcpp ros2cli_601
Doing something terrible!
[ros2run]: Segmentation fault
```

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14738/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9489/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12403/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14884/)

clalancette (approved):

audrow:
Rerunning CI (just in case):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14770/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9519/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12436/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14917/)

audrow:
Thanks for the PR, @fujitatomoya!

PR #641, opened 2021-05-26
[ros2topic pub] Make discovery spin time configurable

ivanpauno:
This is particularly useful when using `ros2 topic pub -1` with reliable publishing.
Sometimes, data will not actually be received because of not long enough discovery time, this PR makes it configurable.

This is not great though, in the future we could:

- Publish messages from the daemon.
- Have an interactive "ros2cli" console. Something like:

```
$ ros2 console enter  # This enters in an interactive program that creates a node and keeps it alive until it's ctrl-c'd
> topic list  # run this a few times until the topic I see the topic I want to publish to
> topic pub ...
```

The interactive `ros2cli` console in an interesting alternative to the daemon approach we're using.

### DIFF
--- a/ros2topic/ros2topic/verb/pub.py
+++ b/ros2topic/ros2topic/verb/pub.py
@@ -19,6 +19,7 @@ from typing import TypeVar
 import rclpy
 from rclpy.node import Node
 from rclpy.qos import QoSProfile
+from ros2cli.node.direct import add_arguments as add_direct_node_arguments
 from ros2cli.node.direct import DirectNode
 from ros2topic.api import qos_profile_from_short_keys
 from ros2topic.api import TopicMessagePrototypeCompleter
@@ -53,6 +54,7 @@ class PubVerb(VerbExtension):
     """Publish a message to a topic."""
 
     def add_arguments(self, parser, cli_name):
+        add_direct_node_arguments(parser)
         arg = parser.add_argument(
             'topic_name',
             help="Name of the ROS topic to publish to (e.g. '/chatter')")
@@ -146,6 +148,7 @@ def main(args):
             1. / args.rate,
             args.print,
             times,
+            args.spin_time,
             qos_profile,
             args.keep_alive)
 
@@ -158,6 +161,7 @@ def publisher(
     period: float,
     print_nth: int,
     times: int,
+    spin_time: float,
     qos_profile: QoSProfile,
     keep_alive: float,
 ) -> Optional[str]:
@@ -189,7 +193,7 @@ def publisher(
         pub.publish(msg)
 
     # give some time for discovery process
-    time.sleep(0.1)
+    time.sleep(spin_time)
     timer_callback()
     if times != 1:
         timer = node.create_timer(period, timer_callback)


ivanpauno:
> This is not great though, in the future we could:

We could also have an `--wait-until-matching-subscriptions N` option, where N=1 would be the default.
`ros2 topic pub` would first loop until finding N matching subscriptions, logging "waiting for N matching subscriptions..." until that happens.

clalancette:
> We could also have an `--wait-until-matching-subscriptions N` option, where N=1 would be the default.

I like this idea, though we would have to have N=0 be the default (there's no guarantee that there are any subscribers out there when you start publishing).

ivanpauno:
> > We could also have an `--wait-until-matching-subscriptions N` option, where N=1 would be the default.
> 
> I like this idea, though we would have to have N=0 be the default (there's no guarantee that there are any subscribers out there when you start publishing).

No, there isn't.
As long as the output is obvious, i.e. `ros2 topic pub` logs periodically that it's waiting for a matching subscription, I don't mind which one is the default.

IMO, `N=1` is nicer, as publishing a message to nobody doesn't make much sense and it's typically really confusing.
It also tends to happen a lot when using connext.

ros2topic/ros2topic/verb/pub.py
```diff
@@ -53,6 +54,7 @@ class PubVerb(VerbExtension):
     """Publish a message to a topic."""
 
     def add_arguments(self, parser, cli_name):
+        add_direct_node_arguments(parser)
```
hidmic, line 57: @ivanpauno nit: `help` string for `--spin-time` is somewhat misleading in this case:

https://github.com/ros2/ros2cli/blob/edbc1600aa161845c58138bc18907bd873670a54/ros2cli/ros2cli/node/direct.py#L91-L94

ivanpauno: why? the goal is exactly to wait for discovery to happen.
It might be less obvious that in the introspection commands, but it seems clear to me.

Maybe: `
'Spin time in seconds to wait for discovery, including endpoint matching if applies (only applies when '           'not using an already running daemon)'`?

ivanpauno: Well, actually you're right, in this case we would be spinning twice which is unexpected.
I will take a look again tomorrow, maybe I will go ahead with the approach commented above https://github.com/ros2/ros2cli/pull/641#issuecomment-848968007.

jacobperron:
An interactive ROS prompt sounds fun :smile:

ros2topic/ros2topic/verb/pub.py
jacobperron: Yeah, it seems like a "wait for N subscriptions" option would be more robust.

hidmic: > Well, actually you're right, in this case we would be spinning twice which is unexpected.

That, and the second spin not affected by the presence of the daemon.

fujitatomoya: > it seems like a "wait for N subscriptions" option would be more robust.

i do think so too :+1:

ivanpauno:
Closing in favor of https://github.com/ros2/ros2cli/pull/642.

ivanpauno:
> An interactive ROS prompt sounds fun

I actually want to have that :smile:, I have some ideas for it

jacobperron:
> I actually want to have that smile, I have some ideas for it

I smell a side project

PR #642, opened 2021-05-27
Add option in ros2 topic pub to wait for N matching subscriptions, use N=1 by default when combined with --times

ivanpauno:
Replaces https://github.com/ros2/ros2cli/pull/641, seems to be a better idea to me.

### DIFF
--- a/ros2topic/ros2topic/verb/pub.py
+++ b/ros2topic/ros2topic/verb/pub.py
@@ -84,8 +84,10 @@ class PubVerb(VerbExtension):
             '-t', '--times', type=nonnegative_int, default=0,
             help='Publish this number of times and then exit')
         parser.add_argument(
-            '-w', '--wait-matching-subscriptions', type=nonnegative_int, default=1,
-            help='Wait until finding the specified number of matching subscriptions')
+            '-w', '--wait-matching-subscriptions', type=nonnegative_int, default=None,
+            help=(
+                'Wait until finding the specified number of matching subscriptions. '
+                'Defaults to 1 when using "-1"/"--once"/"--times", otherwise defaults to 0.'))
         parser.add_argument(
             '--keep-alive', metavar='N', type=positive_float, default=0.1,
             help='Keep publishing node alive for N seconds after the last msg '
@@ -149,7 +151,8 @@ def main(args):
             1. / args.rate,
             args.print,
             times,
-            args.wait_matching_subscriptions,
+            args.wait_matching_subscriptions
+            if args.wait_matching_subscriptions is not None else int(times != 0),
             qos_profile,
             args.keep_alive)
 
--- a/ros2topic/ros2topic/verb/pub.py
+++ b/ros2topic/ros2topic/verb/pub.py
@@ -81,11 +81,11 @@ class PubVerb(VerbExtension):
             '-1', '--once', action='store_true',
             help='Publish one message and exit')
         group.add_argument(
-            '-m', '--wait-matching-subscriptions', type=nonnegative_int, default=1,
-            help='Wait until finding the specified number of matching subscriptions')
-        group.add_argument(
             '-t', '--times', type=nonnegative_int, default=0,
             help='Publish this number of times and then exit')
+        parser.add_argument(
+            '-m', '--wait-matching-subscriptions', type=nonnegative_int, default=1,
+            help='Wait until finding the specified number of matching subscriptions')
         parser.add_argument(
             '--keep-alive', metavar='N', type=positive_float, default=0.1,
             help='Keep publishing node alive for N seconds after the last msg '
--- a/ros2topic/ros2topic/verb/pub.py
+++ b/ros2topic/ros2topic/verb/pub.py
@@ -181,7 +181,8 @@ def publisher(
     while pub.get_subscription_count() < wait_matching_subscriptions:
         # Print a message reporting we're waiting each 1s, check condition each 100ms.
         if not times_since_last_log:
-            print(f'Waiting for {wait_matching_subscriptions} matching subscription(s)...')
+            print(
+                f'Waiting for at least {wait_matching_subscriptions} matching subscription(s)...')
         times_since_last_log = (times_since_last_log + 1) % 10
         time.sleep(0.1)
 
--- a/ros2topic/ros2topic/verb/pub.py
+++ b/ros2topic/ros2topic/verb/pub.py
@@ -81,6 +81,9 @@ class PubVerb(VerbExtension):
             '-1', '--once', action='store_true',
             help='Publish one message and exit')
         group.add_argument(
+            '-m', '--wait-matching-subscriptions', type=nonnegative_int, default=1,
+            help='Wait until finding the specified number of matching subscriptions')
+        group.add_argument(
             '-t', '--times', type=nonnegative_int, default=0,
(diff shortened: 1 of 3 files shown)


audrow (approved):

Also, it would be great to add a test for this.

hidmic (approved):
LGTM, +1 to that test.

jacobperron (commented):
LGTM with a test.

fujitatomoya (approved):

ros2topic/ros2topic/verb/pub.py
```diff
@@ -80,6 +80,9 @@ def add_arguments(self, parser, cli_name):
         group.add_argument(
             '-1', '--once', action='store_true',
             help='Publish one message and exit')
+        group.add_argument(
+            '-m', '--wait-matching-subscriptions', type=nonnegative_int, default=1,
```
clalancette, line 84: I'll still argue that having the default be `1` is a substantial difference in behavior from earlier versions of ROS 2 and ROS 1.  As long as we are OK with that difference in behavior, then we can go forward, but I also think we then need to have a release note about it.

ros2topic/ros2topic/verb/pub.py
```diff
@@ -172,6 +177,14 @@ def publisher(
 
     pub = node.create_publisher(msg_module, topic_name, qos_profile)
 
+    times_since_last_log = 0
+    while pub.get_subscription_count() < wait_matching_subscriptions:
+        # Print a message reporting we're waiting each 1s, check condition each 100ms.
+        if not times_since_last_log:
+            print(f'Waiting for {wait_matching_subscriptions} matching subscription(s)...')
```
clalancette, line 184: I'll suggest:

```suggestion
            print(f'Waiting for at least {wait_matching_subscriptions} matching subscription(s)...')
```

ros2topic/ros2topic/verb/pub.py
ivanpauno: > As long as we are OK with that difference in behavior, then we can go forward, but I also think we then need to have a release note about it.

I can bring this up in the next ROS 2 meeting and ask for opinions, mentioning this in the release notes sounds fine to me.

> I'll still argue that having the default be 1 is a substantial difference in behavior from earlier versions of ROS 2 and ROS 1

This wasn't a problem in ROS 1 as it has an unique graph state that all nodes are aware.
When you created a node and a publisher, it was guaranteed that you were going to see all already existing subscriptions.

This is not the case in ROS 2, and I have got confused with `ros2 topic pub -1 --qos-reliability reliable <topic_name> <topic_type> <message>` not working enough times to think the default behavior isn't intuitive.

Ideally, we would also have https://github.com/ros2/rmw/issues/295 and call it before ending the process, specially when using `--times/--once`.

ros2topic/ros2topic/verb/pub.py
ivanpauno:
```suggestion
            print(
                f'Waiting for at least {wait_matching_subscriptions} matching subscription(s)...')
```

ivanpauno: Fixed in https://github.com/ros2/ros2cli/pull/642/commits/3d59b36971615b8cab6f733578023b182b833864

ivanpauno:

ivanpauno:
I've added a test in https://github.com/ros2/ros2cli/pull/642/commits/19ce63631a8c8c47cdd74175ccb05ff9ccab3211.

ros2topic/test/test_echo_pub.py
```diff
@@ -157,6 +157,8 @@ def message_callback(msg):
                     with launch_testing.tools.launch_process(
                         launch_service, command_action, proc_info, proc_output,
                         output_filter=launch_testing_ros.tools.basic_output_filter(
+                            filtered_patterns=[
+                                r'Waiting for \{[0-7]+\} matching subscription\(s\)\.\.\.'],
```
hidmic, line 161: @ivanpauno why the braces? Those go away after `f-string` interpolation.

hidmic: Also why `[0-7]+`? Wouldn't `[0-9]+` be more comprehensive?

hidmic (approved):

ros2topic/test/test_echo_pub.py
ivanpauno: > Also why [0-7]+? Wouldn't [0-9]+ be more comprehensive?

lol
I've to double-check this, I'm not sure if this filter is doing anything at all.

ivanpauno: I have delted the filter, as we aren't asserting that the output of the command matches something.

ivanpauno:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14706/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9458/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12371/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14850/)

ros2topic/ros2topic/verb/pub.py
```diff
@@ -83,6 +83,9 @@ def add_arguments(self, parser, cli_name):
         group.add_argument(
             '-t', '--times', type=nonnegative_int, default=0,
             help='Publish this number of times and then exit')
+        parser.add_argument(
+            '-m', '--wait-matching-subscriptions', type=nonnegative_int, default=1,
```
clalancette, line 87: This is very subjective, but `-w` feels more natural to me as the short version of the flag.  What do others think?

ivanpauno: Sounds fair to me.

I think I picked `-m` because I also was going to rename the long flag to `--matching-subscriptions`, but I ended up mixing both.
I don't have a strong opinion, either is fine to me.

ivanpauno: Fixed in https://github.com/ros2/ros2cli/pull/642/commits/bf92afecf75376ff2d08c1b39e802d875dafb93c.

ivanpauno:
I have changed the default value according to the offline discussion in https://github.com/ros2/ros2cli/pull/642/commits/93cf8474fb4a4334a8438530adf03897cb803cb1.

clalancette (approved):
This all seems reasonable to me now, so I'm approving.  A few things before we merge this:

1. Green CI.
2. I think a release note in https://github.com/ros2/ros2_documentation/blob/rolling/source/Releases/Release-Humble-Hawksbill.rst is appropriate.
3. I think we should get one more approval before merging this.

hidmic (approved):

ivanpauno:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14718/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9470/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12385/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14864/)

ivanpauno:
CI is green, there're two approvals, PR to document this in the `Humble` release notes is on the way https://github.com/ros2/ros2_documentation/pull/1654.
Going in!!

PR #644, opened 2021-06-01
Add qos autodetect to ros2 topic pub

gonzodepedro:

### DIFF
--- a/ros2topic/ros2topic/verb/pub.py
+++ b/ros2topic/ros2topic/verb/pub.py
@@ -20,7 +20,6 @@ import rclpy
 from rclpy.node import Node
 from rclpy.qos import QoSProfile
 from ros2cli.node.direct import DirectNode
-from ros2topic.api import add_qos_arguments_to_argument_parser
 from ros2topic.api import qos_profile_from_short_keys
 from ros2topic.api import TopicMessagePrototypeCompleter
 from ros2topic.api import TopicNameCompleter
@@ -31,6 +30,7 @@ from rosidl_runtime_py.utilities import get_message
 import yaml
 
 MsgType = TypeVar('MsgType')
+default_profile_str = 'system_default'
 
 
 def nonnegative_int(inval):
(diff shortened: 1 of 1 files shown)


fujitatomoya (commented):
@gonzodepedro i think we need to rebase this on master?

ros2topic/ros2topic/verb/pub.py
```diff
@@ -117,6 +152,66 @@ def main(args):
...
+    subs_info = node.get_subscriptions_info_by_topic(args.topic_name)
+    subscribers_count = len(pubs_info)
+    if subscribers_count == 0:
+        return qos_profile
+
+    for info in subs_info:
+        if (info.qos_profile.reliability == QoSReliabilityPolicy.RELIABLE):
+            reliability_reliable_endpoints_count += 1
+        if (info.qos_profile.durability == QoSDurabilityPolicy.TRANSIENT_LOCAL):
+            durability_transient_local_endpoints_count += 1
```
ivanpauno, line 184: Same here, there isn't any benefit of doing this.
Using TRANSIENT_LOCAL as the default will match both VOLATILE and TRANSIENT_LOCAL subscriptions.

ivanpauno (changes requested):
I don't see the point of the automatching logic in publishers.
The DDS qos compability logic works in the following fashion:

- The publisher **offers** a QoS.
- The subscription **requires** a QoS.

It might make sense in some cases to use this sort of logic in subscriptions, because using the higher quality profile by default might end up in incompatibilities. But I don't see why to do the same in publishers, as using the higher quality profile will never result in incompatibilities.

P.S.: We should also actually extend this reasoning to other places and start using the most flexible qos possible in publishers as the default, there isn't much reason to not do so.

ros2topic/ros2topic/verb/pub.py
ivanpauno: In the case of transient local, the publisher actually might need to keep messages for a longer time than when using volatile.
Using it as the default here for purposes of endpoint compatibility doesn't hurt, particularly considering that we're using a KEEP_LAST history by default.

jacobperron (commented):
I tend to agree with Ivan. I seems like we should make the default reliability *reliable*.

I suppose defaulting durability to *transient local* would be okay. Though I'm not sure if it's what users will expect. E.g. starting a publisher before a subscription will result in a burst of historical messages instead of the most recent message first. If we do the auto matching, then we still have potentially the same burst of messages happening for late-joining subscriptions. ~~TBH, I'm not sure what we should do here. Maybe it's not worth changing the current behavior when it comes to durability.~~

Edit: well actually, the burst of messages should only happen for subscriptions who also set durability to transient local, so it seems more acceptable.

ivanpauno:
> Though I'm not sure if it's what users will expect. E.g. starting a publisher before a subscription will result in a burst of historical messages instead of the most recent message first

That will only happen if the subscription is also using TRANSIENT_LOCAL durability.
i.e.: you're letting the receiving end decide if they want to receive historical data or not.
Though, the amount of historical data that is preserved is decided by the publisher side.

IMHO, using "transient local" by default would be nice.
The default publisher history queue depth doesn't matter much, though it should be "keep last".

gonzodepedro:
Droped in favour of #653

PR #649, opened 2021-06-08
consider not absolute node name when loading parameter

### LINKED ISSUE #648: Can't load non-absolute node name in parameter file
## Bug report

I have a parameter file in the style of 

```
some_node_name:
  ros__parameters:
    a_list:
        - item1
        - item2
```

#### Expected behavior

`ros2 param load` works to load the correct parameters for node `some_node_name` even though the node name is not specified as absolute.

#### Actual behavior

error is raised: `param file does not contain parameters for /some_node_name,  only for nodes: dict_keys(['some_node_name'])`

Note that there's a simple mismatch between the node names given that one is absolute.

#### Additional information

The reproducer here is based on a ros2-control example in which I would like to use the spawner script to spawn a controller with a given type and a parameter file. 
@v-lopez fyi

Karsten1987:
This fixes #648 for me even though it feels pretty dirty. Not sure if there's a better workaround for the described issue.

### DIFF
--- a/ros2param/ros2param/api/__init__.py
+++ b/ros2param/ros2param/api/__init__.py
@@ -146,6 +146,8 @@ def load_parameter_file(*, node, node_name, parameter_file, use_wildcard):
             param_keys.append('/**')
         if node_name in param_file:
             param_keys.append(node_name)
+        if node_name[1:] in param_file:
+            param_keys.append(node_name[1:])
 
         if param_keys == []:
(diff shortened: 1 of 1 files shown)


audrow (commented):
I'm okay with this change. Could you add a test?

fujitatomoya:
closing in favor of https://github.com/ros2/ros2cli/issues/648#issuecomment-1786176686

PR #650, opened 2021-06-15
Generate LICENSE files on ros2 pkg create.

clalancette:
That is, if the '--license' command-line argument is one of
the ones that ament_copyright knows, fetch the license text
out of ament_copyright and place it in 'LICENSE' at the
root of the package.

Here's the current list of strings that will trigger the creation of the LICENSE file:

* Apache License, Version 2.0
* Boost Software License - Version 1.0
* BSD License 2.0
* 2-Clause BSD License
* 3-Clause BSD License
* GNU General Public License 3.0
* GNU Lesser General Public License 3.0
* MIT License
* MIT-0 License

One thing to be aware of here is that this is the first use of `ament_copyright` in an `exec_depend`.  I don't think that should be a problem, but it is something to think about.

An enhancement we could do here is to add the SPDX identifiers to the [license tuple in ament_copyright](https://github.com/ament/ament_lint/blob/100683ca73de2ccf377937e64955d5b79c43721e/ament_copyright/ament_copyright/licenses.py#L21-L22), and then use that instead.  I don't mind strongly which way we go, but we should think about it and be deliberate; changing it later on will be difficult.  Feedback is most welcome.

@nuclearsandwich @tfoote @gbalke @ros2/team FYI.

### DIFF
--- a/ros2pkg/test/test_cli.py
+++ b/ros2pkg/test/test_cli.py
@@ -126,7 +126,7 @@ class TestROS2PkgCLI(unittest.TestCase):
                     'create', 'a_test_package',
                     '--package-format', '3',
                     '--description', 'A test package dummy description',
-                    '--license', 'Apache License, Version 2.0',
+                    '--license', 'Apache-2.0',
                     '--build-type', 'ament_cmake',
                     '--dependencies', 'ros2pkg',
                     '--maintainer-email', 'nobody@nowhere.com',
@@ -146,7 +146,7 @@ class TestROS2PkgCLI(unittest.TestCase):
                     'version: 0.0.0',
                     'description: A test package dummy description',
                     "maintainer: ['Nobody <nobody@nowhere.com>']",
-                    "licenses: ['Apache License, Version 2.0']",
+                    "licenses: ['Apache-2.0']",
                     'build type: ament_cmake',
                     "dependencies: ['ros2pkg']",
                     'node_name: test_node',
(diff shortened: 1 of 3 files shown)


jacobperron:
Supporting the SPDX identifier seems a little more friendly to me, since that might be what some users want to appear in the package.xml.

wjwwood:
I agree about the SPDX stuff. I've already seen two packages use that in their package.xml files during my ros/rosdistro cycle.

tfoote:
+1 for SPDX support it will make our lives much simpler in the future for auditing

clalancette:
All right, I've switched this over to using SPDX identifiers instead.  This PR now requires https://github.com/ament/ament_lint/pull/315 to go in first.

audrow (approved):
Looks good to me.

ros2pkg/test/test_cli.py
```diff
@@ -200,6 +201,6 @@ def test_create_package(self):
             assert root.find('description').text == 'A test package dummy description'
             assert root.find('maintainer').text == 'Nobody'
             assert root.find('maintainer').attrib['email'] == 'nobody@nowhere.com'
-            assert root.find('license').text == 'Apache License 2.0'
+            assert root.find('license').text == 'Apache License, Version 2.0'
```
wjwwood, line 204: What I've seen (and assumed it was the SPDX identifier) was `Apache-2.0`. I guess there are different styles or something?

wjwwood (approved):
lgtm, just a question

ros2pkg/test/test_cli.py
clalancette: No, I'm just dumb.  I never re-ran the tests after I switched to SPDX identifiers, so this test is just going to fail :).  Thanks for pointing it out, I'll update it.

wjwwood (approved):

clalancette:

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14801/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9550/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12467/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14952/)

clalancette:
All right, approved, CI is green.  Going to merge, thanks everyone!

PR #652, opened 2021-07-12
Transfer daemon socket ownership on spawn.

hidmic:
This patch is an alternative to #632, transferring socket ownership on spawn to preclude the TOCTOU races that running daemon checks induce. 

This patch is functional in my local Linux environment. Let's see what CI (and Windows) have to say:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14866/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9616/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12533/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15034/)

### DIFF
--- a/ros2cli/ros2cli/daemon/__init__.py
+++ b/ros2cli/ros2cli/daemon/__init__.py
@@ -1,4 +1,4 @@
-# Copyright 2017-2019 Open Source Robotics Foundation, Inc.
+# Copyright 2017-2021 Open Source Robotics Foundation, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -13,23 +13,115 @@
 # limitations under the License.
 
 import argparse
-from collections import namedtuple
-import functools
-import inspect
 import os
-
-import netifaces
 
 import rclpy
 import rclpy.action
 
-from ros2cli.node.direct import DirectNode
+from ros2cli.helpers import before_invocation
+from ros2cli.helpers import bind
+from ros2cli.helpers import get_ros_domain_id
+from ros2cli.helpers import pretty_print_call
+
+from ros2cli.node.network_aware import NetworkAwareNode
 
 from ros2cli.xmlrpc.local_server import LocalXMLRPCServer
-from ros2cli.xmlrpc.local_server import RequestHandler
+from ros2cli.xmlrpc.local_server import SimpleXMLRPCRequestHandler
 
 
-def main(*, script_name='_ros2_daemon', argv=None):
+def get_port():
+    base_port = 11511
+    base_port += int(os.environ.get('ROS_DOMAIN_ID', 0))
+    return base_port
+
+
+def get_address():
+    return '127.0.0.1', get_port()
+
+
+class RequestHandler(SimpleXMLRPCRequestHandler):
+    rpc_paths = ('/ros2cli/',)
+
+
+def get_xmlrpc_server_url(address=None):
+    if not address:
+        address = get_address()
+    host, port = address
+    path = RequestHandler.rpc_paths[0]
+    return f'http://{host}:{port}{path}'
+
+
+def make_xmlrpc_server():
+    address = get_address()
+
+    return LocalXMLRPCServer(
+        address, logRequests=False,
+        requestHandler=RequestHandler,
+        allow_none=True
+    )
+
+
+def serve_forever(server, *, timeout=2 * 60 * 60):
+    try:
+        ros_domain_id = get_ros_domain_id()
+        node_args = argparse.Namespace(
+            node_name_suffix=f'_daemon_{ros_domain_id}',
+            start_parameter_services=False)
+        with NetworkAwareNode(node_args) as node:
+            functions = [
+                node.get_name,
+                node.get_namespace,
+                node.get_node_names_and_namespaces,
+                node.get_node_names_and_namespaces_with_enclaves,
+                node.get_topic_names_and_types,
+                node.get_service_names_and_types,
+                bind(rclpy.action.get_action_names_and_types, node),
+                node.get_publisher_names_and_types_by_node,
+                node.get_publishers_info_by_topic,
+                node.get_subscriber_names_and_types_by_node,
+                node.get_subscriptions_info_by_topic,
+                node.get_service_names_and_types_by_node,
+                node.get_client_names_and_types_by_node,
+                bind(rclpy.action.get_action_server_names_and_types_by_node, node),
+                bind(rclpy.action.get_action_client_names_and_types_by_node, node),
+                node.count_publishers,
+                node.count_subscribers
+            ]
+
+            server.register_introspection_functions()
+            for func in functions:
+                server.register_function(
+                    before_invocation(
+                        func, pretty_print_call))
+
+            shutdown = False
+
+            # shutdown the daemon in case of a timeout
+            def timeout_handler():
+                nonlocal shutdown
+                print('Shutdown due to timeout')
+                shutdown = True
+            server.handle_timeout = timeout_handler
+            server.timeout = timeout
+
+            # function to shutdown daemon remotely
+            def shutdown_handler():
+                nonlocal shutdown
+                print('Remote shutdown requested')
+                shutdown = True
+            server.register_function(shutdown_handler, 'system.shutdown')
+
+            print('Serving XML-RPC on ' + get_xmlrpc_server_url(server.server_address))
+            try:
+                while not shutdown:
+                    server.handle_request()
+            except KeyboardInterrupt:
+                pass
+    finally:
+        server.server_close()
+
+
+def main(*, argv=None):
     parser = argparse.ArgumentParser(
         formatter_class=argparse.ArgumentDefaultsHelpFormatter)
     parser.add_argument(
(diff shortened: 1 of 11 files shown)


hidmic:
Once CI goes green, I'll document things better.

hidmic:
Alright, it looks like this is working like a charm :smiley:

ros2cli/ros2cli/daemon/daemonize.py
```diff
@@ -0,0 +1,113 @@
...
+class PicklerForProcess(pickle.Pickler):
+
+    def __init__(self, process, *args, **kwargs):
+        super().__init__(process.stdin, *args, **kwargs)
+        self.process = process
+        self.dispatch_table = copyreg.dispatch_table.copy()
+        self.dispatch_table[socket.socket] = self.reduce_socket
+        self.dispatch_table[threading.Event] = self.reduce_event
+
+    def reduce_event(self, obj):
```
ivanpauno, line 36: why is this method needed?

ros2cli/ros2cli/daemon/daemonize.py
```diff
@@ -0,0 +1,113 @@
...
+
+    pickler.dump(func)
+
+    if timeout is not None:
+        def daemon_ready():
+            try:
+                pickler.dump(None)
+                return False
+            except OSError:
+                return True
```
ivanpauno, line 100: how does this work?

ros2cli/ros2cli/daemon/__init__.py
```diff
@@ -120,106 +111,37 @@ def shutdown_handler():
...
-            [f'{k}={v!r}' for k, v in kwargs.items()]
-        )
-        print(f'{name}({arguments})')
-        return func(*args, **kwargs)
-    wrapper.__signature__ = inspect.signature(func)
-    return wrapper
+    serve_forever(make_xmlrpc_server(), timeout=args.timeout)
 
 
 if __name__ == '__main__':
```
ivanpauno, line 147: do we still need this?

ros2cli/ros2cli/daemon/__init__.py
```diff
@@ -120,106 +111,37 @@ def shutdown_handler():
...
-
-def get_daemon_port():
-    base_port = 11511
-    base_port += int(os.environ.get('ROS_DOMAIN_ID', 0))
-    return base_port
+    finally:
+        server.server_close()
+
+
+def main(*, argv=None):
```
ivanpauno, line 124: I'm confused, is the main() function being used or are we calling `serve_forever()` directly?

ros2cli/ros2cli/node/network_aware.py
```diff
@@ -0,0 +1,75 @@
+# Copyright 2021 Open Source Robotics Foundation, Inc.
```
ivanpauno, line 1: I guess you moved the code without editing it, if that's the case that's okay.

ros2cli/ros2cli/xmlrpc/local_server.py
```diff
@@ -14,7 +14,7 @@
 
 import socket
 import struct
-from xmlrpc.server import SimpleXMLRPCRequestHandler
+from xmlrpc.server import SimpleXMLRPCRequestHandler  # noqa
```
ivanpauno, line 17: do we still need this line here?

ivanpauno (commented):
I have some comments and questions.
The refactor sounds good to me, but the code is a bit tricky.

ros2cli/ros2cli/daemon/daemonize.py
hidmic: Good question. When pickling a `LocalXMLRPCServer` instance, eventually we hit a `threading.Event` attribute in [`socketserver.BaseServer`](https://github.com/python/cpython/blob/5fc784e28ab9cd02c30395d1973ff825b787efa2/Lib/socketserver.py#L204). Thus this method, which simply makes sure that another `threading.Event` instance is constructed on the other end.

ros2cli/ros2cli/daemon/__init__.py
hidmic: I kept it for backwards compatibility.

ros2cli/ros2cli/daemon/__init__.py
hidmic: I kept this for backwards compatibility. You can still invoke the daemon separately and it'll use this function. The `ros2 daemon` CLI will daemonize the `serve_forever()` function.

ros2cli/ros2cli/node/network_aware.py
hidmic: I did.

ros2cli/ros2cli/xmlrpc/local_server.py
hidmic: This one's convenience, so downstream code can get the `SimpleXMLRPCRequestHandler` class from the same module the `LocalXMLRPCServer` is coming from.

ros2cli/ros2cli/daemon/daemonize.py
hidmic: Also a good question. 

A daemon that is ready is a daemon that has deserialized the callable sent. When passing a socket, like this patch does when binding the server instance, that means it has been successfully reconstructed on the other end. By waiting on the daemon to be ready, we make sure it is safe to destroy the server and release the socket in the parent process. But only the child process knows when it's ready, so we need a form of communication. To not interfere with `stdout` and `stderr` redirection nor create more IPCs, I chose to use the act of closing the `stdin` pipe as the signal.

So, to summarize, the time flow of `daemonize()` goes like this: 

1. Parent process spawns child process, piping `stdin`.
2. Parent process pickles callable and sends it through `stdin`. 
4. Child process unpickles callable from `stdin`.
5. Child process closes `stdin`.
6. Parent process pickles `None` and sends it through `stdin`.
7. Parent process gets an `stdin` pipe closed exception.
8. Parent process checks child process is still alive.

ros2cli/ros2cli/daemon/daemonize.py
ivanpauno: If in the future `socketserver.BaseServer` implementation change, do we also need to change this class?
That sounds a bit fragile.
Could we send something simpler?
Like only the socket object, and then creating a new XMLRPC server based on that socket (if that's possible).

ros2cli/ros2cli/daemon/daemonize.py
ivanpauno: I see, that sounds like a fair way of solving the issue.
I would also leave a minimal comment here, the logic doesn't seem to be immediately obvious.

ros2cli/ros2cli/daemon/__init__.py
ivanpauno: sounds good to me

ros2cli/ros2cli/xmlrpc/local_server.py
ivanpauno: okay, I would leave a comment saying that we're doing this to re-export SimpleXMLRPCRequestHandler.

ros2cli/ros2cli/daemon/daemonize.py
hidmic: > If in the future `socketserver.BaseServer` implementation change, do we also need to change this class?
> That sounds a bit fragile.

That is correct. That piece of code's been there unchanged for 14 years though. I wouldn't expect it to change anytime soon, and if it does, it doesn't necessarily mean we have to change this code -- only if it changes in a way that requires the `Pickler` subclass to be extended somehow.

> Could we send something simpler?
> Like only the socket object, and then creating a new XMLRPC server based on that socket (if that's possible).

I thought about it. It is possible by re-implementing [`socketserver.TCPServer`](https://github.com/python/cpython/blob/5fc784e28ab9cd02c30395d1973ff825b787efa2/Lib/socketserver.py#L390) and [`xmlrpc.server.SimpleXMLRPCServer`](https://github.com/python/cpython/blob/5fc784e28ab9cd02c30395d1973ff825b787efa2/Lib/xmlrpc/server.py#L580) classes here. I don't love duplicating but that'd do.

ivanpauno: >  only if it changes in a way that requires the Pickler subclass to be extended somehow.

Yes, that's true.
I guess that would be unusual.

> I thought about it. It is possible by re-implementing socketserver.TCPServer and xmlrpc.server.SimpleXMLRPCServer classes here. I don't love duplicating but that'd do.

Okay, if the existing objects don't support that I guess it's fair to serialize everything.

---

I wonder if there's any race when serializing `LocalXMLRPCServer`.
e.g. might the `LocalXMLRPCServer` instance be processing a request (and thus, modifying its internal state) while being serialized?
I haven't checked the python library code, but that would be a problem.

hidmic: > I wonder if there's any race when serializing `LocalXMLRPCServer` e.g. might the `LocalXMLRPCServer` instance be processing a request (and thus, modifying its internal state) while being serialized?

Good question. It cannot process a request until [`handle_request()`](https://github.com/python/cpython/blob/5fc784e28ab9cd02c30395d1973ff825b787efa2/Lib/socketserver.py#L273) is called so that's not a problem BUT if a connection request arrives during the brief period of time in which the socket exists in both processes, I don't know where that connection request will end up. 

That's likely OS specific. I haven't read what's up with file descriptor inheritance on Unix and handle duplication on Windows in that specific case. But in the worst case scenario I guess some client will get their connection refused when the parent process destroys the server.

ros2cli/ros2cli/daemon/daemonize.py
hidmic: Comments and documentation added in https://github.com/ros2/ros2cli/pull/652/commits/4ceb6fef7d400f7f3db94dfdd6c9f71f929fcab4.

ros2cli/ros2cli/xmlrpc/local_server.py
hidmic: Comments and documentation added in https://github.com/ros2/ros2cli/pull/652/commits/4ceb6fef7d400f7f3db94dfdd6c9f71f929fcab4.

hidmic:
CI up to `ros2cli`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14892/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9635/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12557/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15045/)

ivanpauno (approved):

hidmic:
Alright, all green. Going in !

PR #653, opened 2021-07-13
Change default QoSProfile for pub

gonzodepedro:

### DIFF
--- a/ros2topic/ros2topic/verb/pub.py
+++ b/ros2topic/ros2topic/verb/pub.py
@@ -142,7 +142,7 @@ def main(args):
     if qos_profile_name:
         qos_profile = rclpy.qos.QoSPresetProfiles.get_from_short_key(qos_profile_name)
     profile_configure_short_keys(qos_profile, args.qos_reliability, args.qos_durability,
-      args.qos_depth, args.qos_history)
+                                 args.qos_depth, args.qos_history)
 
     times = args.times
     if args.once:
--- a/ros2topic/ros2topic/verb/pub.py
+++ b/ros2topic/ros2topic/verb/pub.py
@@ -141,8 +141,9 @@ def main(args):
     qos_profile_name = args.qos_profile
     if qos_profile_name:
         qos_profile = rclpy.qos.QoSPresetProfiles.get_from_short_key(qos_profile_name)
-    profile_configure_short_keys(qos_profile, args.qos_reliability, args.qos_durability,
-                                 args.qos_depth, args.qos_history)
+    profile_configure_short_keys(
+        qos_profile, args.qos_reliability, args.qos_durability,
+        args.qos_depth, args.qos_history)
 
     times = args.times
     if args.once:
--- a/ros2topic/ros2topic/verb/pub.py
+++ b/ros2topic/ros2topic/verb/pub.py
@@ -19,10 +19,10 @@ from typing import TypeVar
 import rclpy
 from rclpy.node import Node
 from rclpy.qos import QoSDurabilityPolicy
-from rclpy.qos import QoSHistoryPolicy
 from rclpy.qos import QoSProfile
 from rclpy.qos import QoSReliabilityPolicy
 from ros2cli.node.direct import DirectNode
+from ros2topic.api import profile_configure_short_keys
 from ros2topic.api import TopicMessagePrototypeCompleter
 from ros2topic.api import TopicNameCompleter
 from ros2topic.api import TopicTypeCompleter
@@ -32,7 +32,6 @@ from rosidl_runtime_py.utilities import get_message
 import yaml
 
 MsgType = TypeVar('MsgType')
-default_profile_str = 'ACTION_STATUS_DEFAULT'
 
 
 def nonnegative_int(inval):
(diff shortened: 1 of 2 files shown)


gonzodepedro:
This PR uses ACTION_STATUS_DEFAULT which sets "Reliable" reliability and "Transient-local" durability.

jacobperron (commented):
It seems a bit awkward to choose the "action status" preset. How about we don't set a default QoS profile for the `--qos-profile` CLI arg? We can instead use some custom QoS profile, unless the user explicitly passes the `--qos-profile` option.

fujitatomoya (commented):
just a quick question, is there any requirement or actual problems to reset the default?

jacobperron:
> just a quick question, is there any requirement or actual problems to reset the default?

While addressing https://github.com/ros2/ros2cli/issues/594 for `ros2 topic echo`, we also reviewed what QoS was being used for `ros2 topic pub`. Most users won't notice QoS matching issues with the "system default" QoS profile, though it does not guarantee that QoS will always match since it is system dependent. Also, chances are the default durability setting is _volatile_, which won't match _transient\_local_ subscriptions. By making `ros2 topic pub` _reliable_ and _transient\_local_ by default, we provide the most flexible QoS.

As an aside, "system default" was chosen arbitrarily: https://github.com/ros2/ros2cli/pull/245

@fujitatomoya Does that answer your question?

fujitatomoya:
@jacobperron 

it does. thanks for the explanation.

ros2topic/ros2topic/verb/pub.py
```diff
@@ -30,7 +32,7 @@
 import yaml
 
 MsgType = TypeVar('MsgType')
-default_profile_str = 'system_default'
+default_profile_str = 'ACTION_STATUS_DEFAULT'
```
ivanpauno, line 35: I think we can delete this line

ros2topic/ros2topic/verb/pub.py
```diff
@@ -131,17 +137,26 @@ def main(self, *, args):
...
+    if args.qos_durability:
+        qos_profile.durability = QoSDurabilityPolicy.get_from_short_key(args.qos_durability)
+    if args.qos_reliability:
+        qos_profile.reliability = QoSReliabilityPolicy.get_from_short_key(args.qos_reliability)
+    if args.qos_depth and args.qos_depth >= 0:
+        qos_profile.depth = args.qos_depth
+    else:
+        if (qos_profile.durability == rclpy.qos.QoSDurabilityPolicy.TRANSIENT_LOCAL
+                and qos_profile.depth == 0):
+            qos_profile.depth = 1
```
ivanpauno, line 155: instead of duplicating most of the logic from `qos_profile_from_short_keys()`, we could modify it so the first argument can either be a `QoSProfile` object or a profile name (e.g. checking isinstance(qos_profile, str) at the beggining).

ivanpauno (approved):
LGTM, I have a few minor comments

ros2topic/ros2topic/verb/pub.py
gonzodepedro: Thanks for your review. I agree on the inconvenience of duplicating logic from qos_profile_from_short_keys(), but I'm not so sure about using isinstance. So I took a slightly different approach. PTAL

ivanpauno: > I'm not so sure about using isinstance

duck typing is common practice in python.

> So I took a slightly different approach. PTAL

LGTM!

ros2topic/ros2topic/verb/pub.py
```diff
@@ -131,17 +136,18 @@ def main(self, *, args):
...
 def main(args):
+    qos_profile = get_pub_qos_profile()
 
     qos_profile_name = args.qos_profile
-    if not qos_profile_name:
-        qos_profile_name = default_profile_str
+    if qos_profile_name:
+        qos_profile = rclpy.qos.QoSPresetProfiles.get_from_short_key(qos_profile_name)
+    profile_configure_short_keys(qos_profile, args.qos_reliability, args.qos_durability,
+                                 args.qos_depth, args.qos_history)
```
ivanpauno, line 145: nit:

```suggestion
    profile_configure_short_keys(
        qos_profile, args.qos_reliability, args.qos_durability,
        args.qos_depth, args.qos_history)
```

linters accept both styles, but we usually prefer this one.

ivanpauno (approved):
LGTM with green CI.

I would recommend using `--packages-up-to ros2topic` for building and `--packages-select ros2topic` for testing.

gonzodepedro:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15145/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9875/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12834/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15322/)

ivanpauno (approved):

jacobperron (approved):
LGTM

Issue #595, opened 2021-02-02
Disable output buffering by default


jacobperron:
#### Feature description

The current default of Python buffering output can lead to undesirable behavior when working with `ros2cli` tools. For example,

    ros2 echo /chatter | grep "Hello" 

won't output messages right away unless I disable buffering with `PYTHONUNBUFFERED=1`.

I've created this ticket to discuss if we should default to disabling output buffering. Though we've had this discussion related to other features/tools in ROS 2, I couldn't find much about the CLI tools (just this: https://github.com/ros2/ros2cli/pull/588).

#### Implementation considerations

Cons:

Disabling output buffering can lead to performance issues. But, I'd argue that it would be acceptable presuming these tools are primarily used for debugging, and they are not very performant as-is.

Pros:

Users won't get confused when they rely on output of `ros2cli` tools not immediately being fed into another process.
E.g. tests won't have to explicitly set PYTHONUNBUFFERED.

fujitatomoya:
> if we should default to disabling output buffering.

+1 for disabling output buffer default. and we can guide user how to enable buffer it if user wants. i think that could be more friendly for user.

clalancette:
I'm OK with changing the default behavior of the CLI to be unbuffered by default.  I just want to make sure that:

1. All of the CLI tools are consistent, and
2. That there is a way to go back to buffered mode if the user wants

ivanpauno:
We could use stderr for logging which is always unbuffered, though that might be confusing.
I think for commands that return "immediately", using buffered output is fine (e.g. ros2 topic list), but for things like echo it makes sense to have unbuffered ouput.

We can also use the `flush` kwarg of print() in those cases: https://docs.python.org/3/library/functions.html#print.
I think I'm going to use that approach.

clalancette:
> We can also use the `flush` kwarg of print() in those cases: https://docs.python.org/3/library/functions.html#print.
> I think I'm going to use that approach.

I'm kind of against doing it this way.  I think that for any new code, it is going to be *really* easy to forget to do it (and for the reviewer to notice that it is missing).

What if we instead set `PYTHONUNBUFFERED=1` at startup by default, and then let the user disable it using the same mechanism if they want?

ivanpauno:
> What if we instead set PYTHONUNBUFFERED=1 at startup by default, and then let the user disable it using the same mechanism if they want?

[...]

PR #659, opened 2021-08-11
Make the ros2cli output always line buffered

ivanpauno:
Fixes https://github.com/ros2/ros2cli/issues/595.

This forces all calls to print() to also flush the output buffer, the result is similar to having line buffering.

Alternative: We could also do something like what it's suggested here https://stackoverflow.com/questions/107705/disable-output-buffering.

Another alternative: I don't think we need unbuffered output for all commands though, e.g. it's not needed for `ros2 topic list`.
The ones that need unbuffered (or line buffered) output are the ones that are periodically reporting progress, e.g. `ros2 topic pub`, `ros2 topic echo`, etc.
Some of that output could arguably be going to stderr, which in python is always line buffered (even when non interactive), e.g. the output of `ros2 topic pub`.
The output of `ros2 topic echo` should probably go to `stdout` though, and if we don't want global line buffering we would need to pass the flush argument manually.

See also:

- https://docs.python.org/3/library/sys.html#sys.stderr
- https://stackoverflow.com/questions/230751/how-can-i-flush-the-output-of-the-print-function
- https://stackoverflow.com/questions/881696/unbuffered-stdout-in-python-as-in-python-u-from-within-the-program

### DIFF
--- a/ros2cli/ros2cli/cli.py
+++ b/ros2cli/ros2cli/cli.py
@@ -14,12 +14,20 @@
 # limitations under the License.
 
 import argparse
+import builtins
+import functools
 import signal
 
 from ros2cli.command import add_subparsers_on_demand
 
 
 def main(*, script_name='ros2', argv=None, description=None, extension=None):
+    # Make the output always line buffered, even when piping the output to another process.
+    # If you explicitly passed `flush=false`, that's not going to be the case.
+    # This only modifies the behavior of print(), if you write to stdout in another way line
+    # buffering is not guaranteed.
+    builtins.print = functools.partial(print, flush=True)
+
     if description is None:
         description = f'{script_name} is an extensible command-line tool ' \
             'for ROS 2.'
--- a/ros2cli/ros2cli/cli.py
+++ b/ros2cli/ros2cli/cli.py
@@ -14,6 +14,8 @@
 # limitations under the License.
 
 import argparse
+import builtins
+import functools
 import signal
 import sys
 
@@ -68,7 +70,9 @@ def main(*, script_name='ros2', argv=None, description=None, extension=None):
         try:
             sys.stdout.reconfigure(line_buffering=True)
         except AttributeError:
-            # if stdout is not a TextIoWrapper instance, we don't do anything
+            # if stdout is not a TextIoWrapper instance, or we're using python older than 3.7,
+            # force line buffering by patching print
+            builtins.print = functools.partial(print, flush=True)
             pass
 
     if extension is None:
--- a/ros2cli/ros2cli/cli.py
+++ b/ros2cli/ros2cli/cli.py
@@ -14,20 +14,13 @@
 # limitations under the License.
 
 import argparse
-import builtins
-import functools
 import signal
+import sys
 
 from ros2cli.command import add_subparsers_on_demand
 
 
 def main(*, script_name='ros2', argv=None, description=None, extension=None):
-    # Make the output always line buffered, even when piping the output to another process.
-    # If you explicitly passed `flush=false`, that's not going to be the case.
-    # This only modifies the behavior of print(), if you write to stdout in another way line
-    # buffering is not guaranteed.
-    builtins.print = functools.partial(print, flush=True)
-
     if description is None:
         description = f'{script_name} is an extensible command-line tool ' \
             'for ROS 2.'
@@ -37,6 +30,13 @@ def main(*, script_name='ros2', argv=None, description=None, extension=None):
         description=description,
         formatter_class=argparse.RawDescriptionHelpFormatter
     )
+    parser.add_argument(
+        '--use-python-default-buffering',
+        action='store_true',
+        help=(
+            'Do not force line buffering in stdout and instead use the python default buffering, '
+            'which might be affected by PYTHONUNBUFFERED/-u and depends on whatever stdout is '
+            'interactive or not'))
 
     # add arguments for command extension(s)
     if extension:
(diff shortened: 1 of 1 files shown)


hidmic (approved):
I like the approach, but it'd be nice to have a way to toggle that flag (e.g. an envvar,  a CLI argument, take your pick).

fujitatomoya (approved):

clalancette (approved):
I like the approach.  The only real reason to allow disabling of flushing is for performance, but I actually don't think that is super important here.

I think we should run full CI on this, just to make sure something above this package doesn't break.

Finally, when we do merge this I'll request a note on https://github.com/ros2/ros2_documentation/blob/rolling/source/Releases/Release-Humble-Hawksbill.rst , since it is a change in behavior.

wjwwood:
After talking offline about it, the use case I think is steaming relatively high bandwidth data from echo into a file for processing later. In that case performance could be an issue. So I'm with @hidmic and there should be a way to "undo" our always-unbuffered magic.

ivanpauno:
I have made two changes:

- Added an argument to disable line buffering.
- Use [TextIoWrapper.reconfigure()](https://docs.python.org/3/library/io.html#io.TextIOWrapper.reconfigure) instead of patching print().
  The advantage is that only affects `stdout`, and not `print()` to other files.
  `stdout` is [documented](https://docs.python.org/3/library/sys.html#sys.stdout) to be a `FileObject` as returned by `open()`, and [open() docs say](https://docs.python.org/3/library/functions.html#open) that they return a `TextIoWrapper` instance for files in text mode (which stdout always is).
   So I would assume that means that `stdout` is always a `TextIoWrapper` instance with a `reconfigure()` method (except if someone patched `sys.stdout` with a `StringIO` stream before, which cannot happen here).

ivanpauno:
> Use TextIoWrapper.reconfigure() instead of patching print().

I noticed while writing this that that method was added in python 3.7, and we still support python 3.6.
So, I will add a fallback mechanism to support line buffering (or use the previous print() patch).

ivanpauno:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15032/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9769/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12722/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15196/)

ivanpauno:
Windows failure is unrelated

ros2cli/ros2cli/cli.py
```diff
@@ -29,6 +32,13 @@ def main(*, script_name='ros2', argv=None, description=None, extension=None):
         description=description,
         formatter_class=argparse.RawDescriptionHelpFormatter
     )
+    parser.add_argument(
```
clalancette, line 35: Nit: I prefer setting a `default=False` here.  I know that it is implied by `action='store_true'`, but it is just easier for the reader to know what is intended without reading the argparse documentation.

ros2cli/ros2cli/cli.py
```diff
@@ -53,6 +63,18 @@ def main(*, script_name='ros2', argv=None, description=None, extension=None):
...
+        # Make the output always line buffered.
+        # TextIoWrapper has a reconfigure() method, call that if available.
+        # https://docs.python.org/3/library/io.html#io.TextIOWrapper.reconfigure
+        try:
+            sys.stdout.reconfigure(line_buffering=True)
+        except AttributeError:
+            # if stdout is not a TextIoWrapper instance, or we're using python older than 3.7,
+            # force line buffering by patching print
+            builtins.print = functools.partial(print, flush=True)
+            pass
```
clalancette, line 76: Nit: This `pass` is unnecessary.

clalancette (approved):
Two nits inline, but I'm going to approve anyway.

jacobperron (approved):
lgtm with Chris's comments addressed.

ros2cli/ros2cli/cli.py
ivanpauno: Fixed in https://github.com/ros2/ros2cli/pull/659/commits/6c5a30e0ecc9ba90bfe27e1b924a236683318d8d.

ros2cli/ros2cli/cli.py
ivanpauno: Nice catch, fixed in https://github.com/ros2/ros2cli/pull/659/commits/6c5a30e0ecc9ba90bfe27e1b924a236683318d8d.

ros2cli/ros2cli/cli.py
```diff
@@ -53,6 +64,17 @@ def main(*, script_name='ros2', argv=None, description=None, extension=None):
...
+    if not args.use_python_default_buffering:
+        # Make the output always line buffered.
+        # TextIoWrapper has a reconfigure() method, call that if available.
+        # https://docs.python.org/3/library/io.html#io.TextIOWrapper.reconfigure
+        try:
+            sys.stdout.reconfigure(line_buffering=True)
+        except AttributeError:
+            # if stdout is not a TextIoWrapper instance, or we're using python older than 3.7,
+            # force line buffering by patching print
+            builtins.print = functools.partial(print, flush=True)
```
hidmic, line 76: @ivanpauno FYI if the decorator monitors the `file` keyword argument of `print`, you can limit its action to `sys.stdout`.

ivanpauno: true, I didn't feel that was really needed though.
do you think we need a patch for this?

hidmic: Nope, it's unlikely anyone will ever hit it before Python 3.6 support is gone.

fujitatomoya:
This breaks the CI for ros2param (see https://ci.ros2.org/job/ci_linux/15050/testReport/junit/ros2param.ros2param.test/test_verb_dump/test_verb_dump/), we need to revert this now.

PR #681, opened 2021-12-16
WIP: Add enclave information

danthony06:
Adds information about the security enclave associated with a node to the information. This helps us debug complex security configurations in large systems, and verify that nodes are associated with appropriate enclaves. I'd like to get feedback on the approach before I attempt to fix the unit tests.

 Distro A, OPSEC #4584
Signed-off-by: David Anthony david.anthony@swri.org


ros2node/ros2node/api/__init__.py
```diff
@@ -71,6 +72,102 @@ def get_node_names(*, node, include_hidden_nodes=False):
...
+
+    def __call__(self, prefix, parsed_args, **kwargs):
+        include_hidden_nodes = getattr(
+            parsed_args, self.include_hidden_nodes_key) \
+            if self.include_hidden_nodes_key else False
+        with NodeStrategy(parsed_args) as node:
+            return [
+                n.full_name for n in get_node_names(
+                    node=node, include_hidden_nodes=include_hidden_nodes)]
+
```
fujitatomoya, line 170: why we see this difference?

ros2node/ros2node/verb/info.py
```diff
@@ -80,5 +86,9 @@ def main(self, *, args):
                     node=node, remote_node_name=args.node_name, include_hidden=args.include_hidden)
                 print('  Action Clients:')
                 print_names_and_types(actions_clients)
+                print('  Enclaves:')
+                print_enclaves([n.enclave for n in node_names_with_enclaves
+                    if n.name.full_name == args.node_name])
+
```
fujitatomoya, line 92: we can delete this line?

ros2node/ros2node/verb/info.py
```diff
@@ -80,5 +86,9 @@ def main(self, *, args):
                     node=node, remote_node_name=args.node_name, include_hidden=args.include_hidden)
                 print('  Action Clients:')
                 print_names_and_types(actions_clients)
+                print('  Enclaves:')
+                print_enclaves([n.enclave for n in node_names_with_enclaves
+                    if n.name.full_name == args.node_name])
```
fujitatomoya, line 91: i am okay with either way, but we could consider this is one of the option for `ros2 node info`.

ros2node/ros2node/api/__init__.py
danthony06: Sorry, I'm not sure how all that was duplicated. I'll remove it.

ros2node/ros2node/verb/info.py
danthony06: I wasn't sure which is the best option. My rationale for leaving it in the default output is:

1. Most people are running unsecured systems, and in this case it does not add a lot to the information output
2. If someone is running a secured system, this can be important information, and users might want to frequently see this when they are examining a system.

danthony06: Is there any interest in this, or should I withdraw the PR? Internally, we've found that this information would be very useful for our development, and I'd like to help the community if possible.

fujitatomoya: I think this is helpful. can you open this PR as `Non-Draft` for review?

> Most people are running unsecured systems, and in this case it does not add a lot to the information output

IMO, this is why i came up with option such as `--include-enclaves`.

PR #685, opened 2021-12-23
Depend on launch packages instead of ros_testing to avoid circular dependency

sloretz:
ros_testing depends on ros2test which depends on ros2cli. This causes a circular dependency in the repository that means `ros-rolling-ros2cli` gets installed in the PR job for this repo.

### DIFF
--- a/ros2action/package.xml
+++ b/ros2action/package.xml
@@ -27,8 +27,11 @@
   <test_depend>ament_flake8</test_depend>
   <test_depend>ament_pep257</test_depend>
   <test_depend>ament_xmllint</test_depend>
+  <test_depend>launch</test_depend>
+  <test_depend>launch_ros</test_depend>
+  <test_depend>launch_testing</test_depend>
+  <test_depend>launch_testing_ros</test_depend>
   <test_depend>python3-pytest</test_depend>
-  <test_depend>ros_testing</test_depend>
   <test_depend>test_msgs</test_depend>
 
   <export>
--- a/ros2action/package.xml
+++ b/ros2action/package.xml
@@ -28,7 +28,6 @@
   <test_depend>ament_pep257</test_depend>
   <test_depend>ament_xmllint</test_depend>
   <test_depend>launch</test_depend>
-  <test_depend>launch_ros</test_depend>
   <test_depend>launch_testing</test_depend>
   <test_depend>launch_testing_ros</test_depend>
   <test_depend>python3-pytest</test_depend>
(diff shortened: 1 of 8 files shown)


ros2action/package.xml
```diff
@@ -27,8 +27,11 @@
   <test_depend>ament_flake8</test_depend>
   <test_depend>ament_pep257</test_depend>
   <test_depend>ament_xmllint</test_depend>
+  <test_depend>launch</test_depend>
+  <test_depend>launch_ros</test_depend>
```
clalancette, line 31: Technically, I don't think this package directly depends on `launch_ros`.

ros2interface/package.xml
```diff
@@ -26,8 +26,11 @@
   <test_depend>ament_flake8</test_depend>
   <test_depend>ament_pep257</test_depend>
   <test_depend>ament_xmllint</test_depend>
+  <test_depend>launch</test_depend>
+  <test_depend>launch_ros</test_depend>
```
clalancette, line 30: Technically, I don't think this package directly depends on launch_ros.

ros2pkg/package.xml
```diff
@@ -28,8 +28,11 @@
   <test_depend>ament_flake8</test_depend>
   <test_depend>ament_pep257</test_depend>
   <test_depend>ament_xmllint</test_depend>
+  <test_depend>launch</test_depend>
+  <test_depend>launch_ros</test_depend>
```
clalancette, line 32: Technically, I don't think this package directly depends on launch_ros.

clalancette (changes requested):
There are 3 packages here that don't actually depend on `launch_ros`.  It's a bit nit-picky, but since we are reducing dependencies we may as well only include the things that we need.

ros2action/package.xml
sloretz: Fixed in 6f5a060

ros2interface/package.xml
sloretz: 6f5a060

ros2pkg/package.xml
sloretz: 6f5a060

clalancette (approved):

sloretz:
CI (build: ` --packages-up-to ros2action ros2doctor ros2interface ros2lifecycle ros2node ros2pkg ros2service ros2topic` test: `--packages-select ros2action ros2doctor ros2interface ros2lifecycle ros2node ros2pkg ros2service ros2topic`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15851/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10528/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13516/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16143/)

Issue #637, opened 2021-05-14
Daemon slows down CLI responsiveness on Windows


hidmic:
## Bug report

**Required Info:**

- Operating System:
  - Windows 10 
- Installation type:
  - From source
- Version or commit hash:
  - Current [master](edbc1600aa161845c58138bc18907bd873670a54) 
- DDS implementation:
  - All 
- Client library (if applicable):
  - rclpy 

#### Steps to reproduce issue

On a Windows box, do the following with and without a daemon node running:

- Start an action server 
  ```sh
  ros2 run action_tutorials_cpp fibonacci_action_server
  ```
- Ensure daemon is not running
  ```sh
  ros2 daemon stop
  ```
- Query action server info
  ```sh
  ros2 action info /fibonacci_action_server
  ```

#### Expected behavior

Querying with and without a daemon running takes a comparable amount of time.

#### Actual behavior

Querying with a daemon running takes a way longer.

PR #686, opened 2022-01-09
Improve performance of commands for Foxy on Windows by 4x

Ace314159:
This improves the performance of commands like `ros2 topic list`, `ros2 service list`, `ros2 node list`, and I assume anything else using the daemon proxy. I was able to find this change and an explanation from [this Stack Overflow post](https://stackoverflow.com/questions/14504450/pythons-xmlrpc-extremely-slow-one-second-per-call).

When having no nodes running, the execution time of the above commands go from 5.3 seconds to 1.2 seconds on my machine. This is still much slower than running these commands in WSL, but now the majority of the delay seems to be from the imports in `ros2-script.py` and `list.py`

This also addresses #637

### DIFF
--- a/ros2cli/ros2cli/node/daemon.py
+++ b/ros2cli/ros2cli/node/daemon.py
@@ -101,7 +101,7 @@ class DaemonNode:
     def __init__(self, args):
         self._args = args
         self._proxy = ServerProxy(
-            'http://localhost:%d/ros2cli/' % get_daemon_port(),
+            'http://127.0.0.1:%d/ros2cli/' % get_daemon_port(),
             allow_none=True)
         self._methods = []
 
--- a/ros2cli/ros2cli/node/daemon.py
+++ b/ros2cli/ros2cli/node/daemon.py
@@ -100,8 +100,9 @@ class DaemonNode:
 
     def __init__(self, args):
         self._args = args
+        host = '127.0.0.1' if platform.system() == 'Windows' else 'localhost'
         self._proxy = ServerProxy(
-            'http://127.0.0.1:%d/ros2cli/' % get_daemon_port(),
+            'http://%s:%d/ros2cli/' % (host, get_daemon_port()),
             allow_none=True)
         self._methods = []
 
--- a/ros2cli/ros2cli/node/daemon.py
+++ b/ros2cli/ros2cli/node/daemon.py
@@ -100,9 +100,8 @@ class DaemonNode:
 
     def __init__(self, args):
         self._args = args
-        host = '127.0.0.1' if platform.system() == 'Windows' else 'localhost'
         self._proxy = ServerProxy(
-            'http://%s:%d/ros2cli/' % (host, get_daemon_port()),
+            'http://127.0.0.1:%d/ros2cli/' % get_daemon_port(),
             allow_none=True)
         self._methods = []
 
--- a/ros2action/test/test_cli.py
+++ b/ros2action/test/test_cli.py
@@ -35,6 +35,14 @@ from rclpy.utilities import get_available_rmw_implementations
 import yaml
 
 
+# Skip cli tests on Windows while they exhibit pathological behavior
+# https://github.com/ros2/build_farmer/issues/248
+if sys.platform.startswith('win'):
+    pytest.skip(
+            'CLI tests can block for a pathological amount of time on Windows.',
+            allow_module_level=True)
+
+
 @pytest.mark.rostest
 @launch_testing.parametrize('rmw_implementation', get_available_rmw_implementations())
 def generate_test_description(rmw_implementation):
--- a/ros2action/test/test_cli.py
+++ b/ros2action/test/test_cli.py
@@ -35,14 +35,6 @@ from rclpy.utilities import get_available_rmw_implementations
 import yaml
 
 
-# Skip cli tests on Windows while they exhibit pathological behavior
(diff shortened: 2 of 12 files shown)


audrow (commented):
This seems like a good change to me. Here are a few thoughts:

- It seems that a few tests are failing in the PR job - maybe we can use `127.0.0.1` only for Windows platforms? Perhaps that would make the tests pass.
- Should this be retargeted to rolling and then maybe backported to Foxy (and Galactic) as a bug fix?
- Is there a good way to test this? I believe some tests are disabled on Windows for speed. Maybe some of those can be enabled?

clalancette:
> * Should this be retargeted to rolling and then maybe backported to Foxy (and Galactic) as a bug fix?

Yes, we should definitely retarget it to rolling, as that is our standard development process.  If we find we can make it work on Rolling, then we can consider backporting.

Ace314159:
I'm not able to reproduce these the test failures on WSL, and this change already seems to be made in rolling [here](https://github.com/ros2/ros2cli/blob/master/ros2cli/ros2cli/daemon/__init__.py#L40), which seems to be fine. I can make the same change to galactic once the test failures on Foxy are figured out.

Are these test known to be flaky or is it actually the change itself causing the failures?

I can try enabling the tests on Windows to test this. The commands are still about 4x slower than in WSL, so it still might not be fast enough to pass all the tests.

Ace314159:
In [Build 67](https://build.ros2.org/job/Fpr__ros2cli__ubuntu_focal_amd64/67/#showFailuresLink), service and doctor which originally failed don't fail, but topic and node still do. In [Build 68](https://build.ros2.org/job/Fpr__ros2cli__ubuntu_focal_amd64/68/#showFailuresLink), doctor doesn't fail, but action does now, and topic and node still fail.

It appears that only topic and node consistently fail, but I still can't reproduce either of them locally on WSL. Any ideas why it's failing here even though the same change is already in rolling?

Also, can you point me to where I can see the Windows PR job?

Ace314159:
Even when using `127.0.0.1` only on Windows, the tests still don't pass. Any ideas for what else I could try?

clalancette:
Especially for Foxy, let's keep the `test_cli` tests disabled for Windows.  We had massive problems with them, and I'm pretty sure fixing them isn't going to be as simple as changing to 127.0.0.1 here.

Ace314159:
@clalancette Is there anything else you'd like me to do? Although this change alone might not be enough to fix the tests, this still results in a substantial speed increase, making the ros2 CLI a bit more usable on Windows.

Ace314159:
@audrow @clalancette Friendly ping

Ace314159:
I was able to generate flamegraphs and inverted flamegraphs using [flameprof](https://github.com/baverman/flameprof/) in order to better visualize the impacts of this seemingly minimal change. I was also able to find identify another bottleneck involvinng `os.stat` that might be able to explain any remaining speed difference between Windows and Linux.

I ran `ros2 topic list` without any nodes running, before and after the changes on Windows. I also ran it on WSL to get an estimate of the performance on Linux to identify any remaining bottlenecks.

Before the change, the call took ~5.3 seconds, and the flame graph and inverted flame graph look like below:
image
image

The majority of the time is spent in the `create_connection` method of socket, which I assume is called when using the `ServerProxy`.

After the change, the call took ~1.2 seconds, and the flame graph and inverted flame graph look like below:
image
image

The massive bottleneck involving the `ServerProxy` is removed, and the remaining bottleneck is due to `nt.stat`. Reading from the inverted flame graph, the total time of all calls of `nt.stat` calls is ~0.29 seconds.

The WSL call took ~0.9 seconds, and the flame graph and inverted flame graph look like below:
image
image

In WSL, `posix.stat`, which is the largest (but still pretty small) dark purple bar on the right in the inverted flame graph, is much smaller and barely visible. Reading from the inverted flame graph, the total time of all calls of `posix.stat` is ~0.059 seconds.

This probably means that the ~0.3 second difference between WSL and Windows is due to `os.stat` being much slower on Windows.

Ace314159:
@clalancette @audrow Just to note, since I seemed to have forgotten to mention this, this change is already in rolling, so it would just be a matter of backporting to foxy and galactic:
https://github.com/ros2/ros2cli/blob/0479d5d17ee0f7ccad2b30835d464d5aaa19bb65/ros2cli/ros2cli/daemon/__init__.py#L40

ros2cli/ros2cli/node/daemon.py
```diff
@@ -100,8 +100,9 @@ class DaemonNode:
 
     def __init__(self, args):
         self._args = args
+        host = '127.0.0.1' if platform.system() == 'Windows' else 'localhost'
```
clalancette, line 103: So if we are going to follow in the same idea as what Rolling has, I'll suggest we just unconditionally use 127.0.0.1 here.

clalancette (changes requested):
One small change to make, then I think this is good to go.  We'll also have to figure out why CI is unhappy.

clalancette (approved):
This looks like a reasonable enough backport for me.  I'll run CI on it next.

clalancette:
OK, CI is happy (the failures on Windows are well-known, and not related to this PR).  I'm going to go ahead and merge this one in, thanks for all of the work here.

jacobperron:
I noticed this was never ported to Galactic. I've opened up a PR here: https://github.com/ros2/ros2cli/pull/740

PR #687, opened 2022-01-19
Switch ros2 doctor to using psutil for network checks.

clalancette:
ifcfg has a couple of different problems.  The first is that
the code literally calls 'ifconfig' and parses the output.  While
this generally works, it is pretty ugly.  For those reasons,
Debian has specifically excluded it.

Instead, switch to psutil to get information about the network
interfaces on the system (at least, on POSIX).  This gets us
all of the information we had before, but using ioctl's instead.

Note that we can't actually get any flags from psutil at the
moment.  There is a PR open upstream to get this information,
but since it is not going to be available soon we implement
the equivalent thing in Python by using fcntl.ioctl.

This will also require a change to CI to install psutil on Windows; PR for that will be upcoming.  @nuclearsandwich @j-rivero FYI

### DIFF
--- a/ros2doctor/ros2doctor/api/network.py
+++ b/ros2doctor/ros2doctor/api/network.py
@@ -13,9 +13,11 @@
 # limitations under the License.
 
 import os
-from typing import Tuple
+import socket
+import struct
+import sys
 
-import ifcfg
+import psutil
 
 from ros2doctor.api import DoctorCheck
 from ros2doctor.api import DoctorReport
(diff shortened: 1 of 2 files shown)


ros2doctor/ros2doctor/api/network.py
```diff
@@ -87,12 +176,44 @@ def category(self):
...
+            if_info = {
+                'inet': None,
+                'inet4': [],
+                'ether': None,
+                'inet6': [],
+                'netmask': None,
+                'device': interface,
+                'flags': None,
+                'mtu': None,
+                'broadcast': None,
```
nuclearsandwich, line 193: This dict seems like a prime use case for [dataclasses](https://docs.python.org/3/library/dataclasses.html) would you accept a patch that creates and uses an InterfaceInfo dataclass instead of a dict or would that be gilding the lily?

nuclearsandwich: On further reading since these values are added to the report in raw key-value pairs we'd have to tease the dataclass apart anyway https://github.com/ros2/ros2cli/pull/687/files/e4b540a3f1aa10215a7f7064b26281b7331a914d#diff-9fcdc075292a69769ee336433dd05af791b6a8363dc138ac06fe852ac5aed05aR216-R218

nuclearsandwich (approved):

ros2doctor/ros2doctor/api/network.py
clalancette: > On further reading since these values are added to the report in raw key-value pairs we'd have to tease the dataclass apart anyway

Right.  This way seems pretty straightforward to me (though not the shortest of code), so I'm inclined to leave it unless there is another way that is clearer.

clalancette:
CI for this is in https://github.com/ros2/ci/pull/609#issuecomment-1018507224 .

ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-january-20th-2022/23986/1

Issue #477, opened 2020-04-08
Default email when creating a package should use local user.email value, not global


comkieffer:
## Bug report

**Required Info:**

- Operating System: linux
- Installation type: binaries
- Version or commit hash: eloquent
- DDS implementation: N/A
- Client library (if applicable): N/A

#### Steps to reproduce issue

Inside the workspace override the global git config:
```
git config user.email test@example.com
```

Once you're done, clear the value with 
```
git config --unset user.email
```

#### Expected behavior

`ros2 pkg create` should use the local value of `user.email` if it is set. 

#### Actual behavior

`ros2 pkg create` specifically asks for the `global` value: [ros2pkg/ros2pkg/verb/create.py:97](https://github.com/ros2/ros2cli/blob/master/ros2pkg/ros2pkg/verb/create.py#L97)

Just removing `--global` would solve the issue. If the value is not set in the repository, git will fall back to the global value in any case. 

If this makes sense to you, I can create a pull request for this.

dirk-thomas:
Sounds good. Please go ahead and create a pull request for this enhancement.

PR #692, opened 2022-02-11
User git local config instead of global

aalbaali:
Using local git config instead of global config.

This issue is reported in #477 and is fixed for ROS2 Eloquent in #479 but not for Foxy.

Signed-off-by: aalbaali <albaalia@live.com>


clalancette:
Can you please retarget this to the `master` branch?  That way we will have it in all distributions going forward.  Once we have that approved and in, we can backport it to `foxy` and `galactic`.  Thanks.

aalbaali:
@clalancette I closed this PR and issued another one targeting `master` in #693.

PR #711, opened 2022-05-09
Remove unused arguments from ros2 component types.

### LINKED ISSUE #710: Remove options that do not do anything from ros2component
This is not really a bug or a feature request. I just noticed that when we run `ros2 component types --help` we see the following options that have no affect on the operation of listing items in the ament index:

```
  --spin-time SPIN_TIME
```

IMO, it would be better to remove those from the output to avoid any potential confusion.

clalancette:
It never uses a node, so it should never have node-style
arguments.

Should fix #710

### DIFF
--- a/ros2component/ros2component/verb/types.py
+++ b/ros2component/ros2component/verb/types.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from ros2cli.node.strategy import add_arguments
 from ros2component.api import get_package_component_types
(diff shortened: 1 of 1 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16668/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11254/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17063/)

fujitatomoya (approved):

PR #716, opened 2022-05-25
refactor: make ros2param use rclpy.parameter_client

ihasdapie:
Use `rclpy.parameter_client` from instead of internal implementation (Related PR: https://github.com/ros2/rclpy/pull/945)  to avoid duplicate logic. Certain functions (e.g. `get_parameter_value`) are moved over to `rclpy` and `ros2param`. Behavior of `ros2param.api` is left unchanged.

I'd like some review on the proposed architecture & implementation before I clean things up and write tests etc.

### DIFF
--- a/ros2param/ros2param/api/__init__.py
+++ b/ros2param/ros2param/api/__init__.py
@@ -114,16 +114,13 @@ def call_set_parameters(*, node, node_name, parameters):
 def call_list_parameters(*, node, node_name, prefixes=None):
     client = AsyncParameterClient(node, node_name)
     ready = client.wait_for_services(timeout_sec=5.0)
+    # return None on exception cases so that ros2 param list still
+    # prints parameters for nodes that are working
     if not ready:
-        raise RuntimeError('Wait for service timed out waiting for '
-                           f'parameter services for node {node_name}')
+        return None
     future = client.list_parameters(prefixes=prefixes)
     rclpy.spin_until_future_complete(node, future)
-    response = future.result()
-    if response is None:
-        raise RuntimeError('Exception while calling service of node '
-                           f'{node_name}: {future.exception()}')
-    return response.result.names
+    return future
 
 
 def get_parameter_type_string(parameter_type):
--- a/ros2param/ros2param/api/__init__.py
+++ b/ros2param/ros2param/api/__init__.py
@@ -43,7 +43,7 @@ def get_value(*, parameter_value):
 def load_parameter_file(*, node, node_name, parameter_file, use_wildcard):
     # Remove leading slash and namespaces
     client = AsyncParameterClient(node, node_name)
-    client.wait_for_services(30)
+    client.wait_for_services(timeout_sec=5.0)
     if not client.services_are_ready():
         raise RuntimeError('Could not reach parameter services')
 
@@ -69,7 +69,7 @@ def load_parameter_file(*, node, node_name, parameter_file, use_wildcard):
 
 def call_describe_parameters(*, node, node_name, parameter_names=None):
     client = AsyncParameterClient(node, node_name)
-    client.wait_for_services(30)
+    client.wait_for_services(5.0)
     if not client.services_are_ready():
         raise RuntimeError('Could not reach parameter services')
     future = client.describe_parameters(parameter_names)
@@ -80,7 +80,7 @@ def call_describe_parameters(*, node, node_name, parameter_names=None):
 
 def call_get_parameters(*, node, node_name, parameter_names):
     client = AsyncParameterClient(node, node_name)
-    client.wait_for_services(30)
+    client.wait_for_services(timeout_sec=5.0)
     if not client.services_are_ready():
         raise RuntimeError('Could not reach parameter services')
     future = client.get_parameters(parameter_names)
@@ -91,7 +91,7 @@ def call_get_parameters(*, node, node_name, parameter_names):
 
 def call_set_parameters(*, node, node_name, parameters):
     client = AsyncParameterClient(node, node_name)
-    client.wait_for_services(30)
+    client.wait_for_services(timeout_sec=5.0)
     if not client.services_are_ready():
         raise RuntimeError('Could not reach parameter services')
     future = client.set_parameters(parameters)
@@ -102,7 +102,7 @@ def call_set_parameters(*, node, node_name, parameters):
 
 def call_list_parameters(*, node, node_name):
     client = AsyncParameterClient(node, node_name)
-    client.wait_for_services(30)
+    client.wait_for_services(timeout_sec=5.0)
     if not client.services_are_ready():
         raise RuntimeError('Could not reach parameter services')
     future = client.list_parameters()
--- a/ros2param/ros2param/api/__init__.py
+++ b/ros2param/ros2param/api/__init__.py
@@ -69,7 +69,7 @@ def load_parameter_file(*, node, node_name, parameter_file, use_wildcard):
 
 def call_describe_parameters(*, node, node_name, parameter_names=None):
     client = AsyncParameterClient(node, node_name)
-    client.wait_for_services(5.0)
+    client.wait_for_services(timeout_sec=5.0)
     if not client.services_are_ready():
         raise RuntimeError('Could not reach parameter services')
     future = client.describe_parameters(parameter_names)
@@ -106,7 +106,7 @@ def call_list_parameters(*, node, node_name):
     if not client.services_are_ready():
         raise RuntimeError('Could not reach parameter services')
     future = client.list_parameters()
-    rclpy.spin_until_future_complete(node, future)
+    rclpy.spin_until_future_complete(node, future)  # This is hanging..
     response = future.result()
     return response.result.names
 
--- a/ros2param/ros2param/api/__init__.py
+++ b/ros2param/ros2param/api/__init__.py
@@ -42,7 +42,9 @@ def get_value(*, parameter_value):
 
 def load_parameter_file(*, node, node_name, parameter_file, use_wildcard):
     client = AsyncParameterClient(node, node_name)
-    client.wait_for_services(timeout_sec=5.0)
+    ready = client.wait_for_services(timeout_sec=5.0)
+    if not ready:
+        raise RuntimeError('Wait for service timed out')
     future = client.load_parameter_file(parameter_file, use_wildcard)
     parameters = list(parameter_dict_from_yaml_file(parameter_file, use_wildcard).values())
     rclpy.spin_until_future_complete(node, future)
@@ -65,7 +67,9 @@ def load_parameter_file(*, node, node_name, parameter_file, use_wildcard):
 
 def call_describe_parameters(*, node, node_name, parameter_names=None):
     client = AsyncParameterClient(node, node_name)
-    client.wait_for_services(timeout_sec=5.0)
+    ready = client.wait_for_services(timeout_sec=5.0)
+    if not ready:
+        raise RuntimeError('Wait for service timed out')
     future = client.describe_parameters(parameter_names)
     rclpy.spin_until_future_complete(node, future)
     response = future.result()
@@ -77,7 +81,9 @@ def call_describe_parameters(*, node, node_name, parameter_names=None):
 
 def call_get_parameters(*, node, node_name, parameter_names):
     client = AsyncParameterClient(node, node_name)
-    client.wait_for_services(timeout_sec=5.0)
+    ready = client.wait_for_services(timeout_sec=5.0)
+    if not ready:
+        raise RuntimeError('Wait for service timed out')
     future = client.get_parameters(parameter_names)
     rclpy.spin_until_future_complete(node, future)
     response = future.result()
@@ -89,7 +95,9 @@ def call_get_parameters(*, node, node_name, parameter_names):
 
 def call_set_parameters(*, node, node_name, parameters):
     client = AsyncParameterClient(node, node_name)
-    client.wait_for_services(timeout_sec=5.0)
+    ready = client.wait_for_services(timeout_sec=5.0)
+    if not ready:
+        raise RuntimeError('Wait for service timed out')
     future = client.set_parameters(parameters)
     rclpy.spin_until_future_complete(node, future)
     response = future.result()
@@ -101,7 +109,9 @@ def call_set_parameters(*, node, node_name, parameters):
 
 def call_list_parameters(*, node, node_name, prefixes=None):
     client = AsyncParameterClient(node, node_name)
-    client.wait_for_services(timeout_sec=5.0)
+    ready = client.wait_for_services(timeout_sec=5.0)
+    if not ready:
+        raise RuntimeError('Wait for service timed out')
     future = client.list_parameters(prefixes=prefixes)
     rclpy.spin_until_future_complete(node, future)
     response = future.result()
--- a/ros2param/ros2param/api/__init__.py
+++ b/ros2param/ros2param/api/__init__.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 import sys
-import time
 import warnings
 
 from rcl_interfaces.msg import ParameterType
@@ -41,20 +40,9 @@ def get_value(*, parameter_value):
     return value
 
 
-def busy_wait_client(client, node):
-    timeout_sec = 5
-    prev = time.time()
-    while not (client.services_are_ready() or timeout_sec < 0):
-        client.wait_for_services(timeout_sec=0.1)
-        rclpy.spin_once(node)
-        timeout_sec -= prev - time.time()
-    if not client.services_are_ready():
-        raise RuntimeError('Could not reach parameter services')
-
-
 def load_parameter_file(*, node, node_name, parameter_file, use_wildcard):
     client = AsyncParameterClient(node, node_name)
-    busy_wait_client(client, node)
+    client.wait_for_services(timeout_sec=5.0)
     future = client.load_parameter_file(parameter_file, use_wildcard)
     parameters = list(parameter_dict_from_yaml_file(parameter_file, use_wildcard).values())
     rclpy.spin_until_future_complete(node, future)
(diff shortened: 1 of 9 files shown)


ihasdapie:
Just a sanity check:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16751/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11334/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17151/)

ros2param/ros2param/api/__init__.py
```diff
@@ -14,112 +14,38 @@
...
+from rcl_interfaces.msg import ParameterType
+from rclpy.parameter import parameter_value_to_python
+from rclpy.parameter import get_parameter_value as rclpy_get_parameter_value
+from rclpy.parameter_client import AsyncParameterClient
+
 from ros2cli.node.direct import DirectNode
 
-import yaml
+
+def get_parameter_value(*, string_value):
```
jacobperron, line 26: If we've replaced all calls to this function in ros2cli, then I guess we add a deprecation warning here. You can `import warnings` and then call `warnings.warn('get_parameter_value() is deprecated. Use rclpy.parameter.get_parameter_value instead')` in this function.

jacobperron (commented):
The refactor makes sense to me :+1:

ros2param/ros2param/api/__init__.py
```diff
@@ -13,113 +13,41 @@
...
-            value.string_array_value = yaml_value
-        else:
-            value.type = ParameterType.PARAMETER_STRING
-            value.string_value = string_value
-    else:
-        value.type = ParameterType.PARAMETER_STRING
-        value.string_value = yaml_value
+        value = parameter_value_to_python(parameter_value)
+    except RuntimeError as e:
+        print(f'Runtime error {e} raised')
```
ihasdapie, line 39: In the original implementation they ignored the error whereas in the `rclpy` implementation an error is raised. I think we should allow for the error to be raised but this is here to preserve the old behaviour

ihasdapie:
There are still large swathes of code scattered throughout which look like they should be replaced with `get_parameter_value` but are all *slightly* different. [like this](https://github.com/ros2/ros2cli/blob/faf3d052fc5bdc1cf981b5b1f49e5b9dc7b10148/ros2param/ros2param/verb/get.py#L97-L68)

ihasdapie:
build fail due to gh actions not using related branch. See:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16772/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11353/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17174/)

ros2param/ros2param/api/__init__.py
```diff
@@ -135,103 +63,34 @@ def load_parameter_dict(*, node, node_name, parameter_dict):
...
+    client = AsyncParameterClient(node, node_name)
+    future = client.set_parameters(parameters)
     rclpy.spin_until_future_complete(node, future)
-
-    # handle response
     response = future.result()
     return response
 
 
 def call_list_parameters(*, node, node_name, prefix=None):
```
jacobperron, line 90: I noticed we're not using (nor were we previously using) the `prefix` argument. I suppose we could just remove it:

```suggestion
def call_list_parameters(*, node, node_name):
```

ros2param/ros2param/api/__init__.py
```diff
@@ -135,103 +63,34 @@ def load_parameter_dict(*, node, node_name, parameter_dict):
...
-
-    # call as soon as ready
-    ready = client.wait_for_service(timeout_sec=5.0)
-    if not ready:
-        raise RuntimeError('Wait for service timed out')
-
-    request = ListParameters.Request()
-    future = client.call_async(request)
+    client = AsyncParameterClient(node, node_name)
+    future = client.list_parameters()
```
jacobperron, line 92: Before the refactor, we were waiting for the service to be ready. Maybe we should wait for the services to be ready before calling `list_parameters()`?

Same question applies to all other service calls above.

ros2param/ros2param/api/__init__.py
jacobperron: Under what conditions do we expect to see this runtime error? I'm just wondering if we should quietly print nothing instead.

ros2param/ros2param/verb/dump.py
```diff
@@ -97,6 +94,7 @@ def main(self, *, args):  # noqa: D102
             # create client
             service_name = f'{absolute_node_name}/list_parameters'
             client = node.create_client(ListParameters, service_name)
+            # TODO(ihasdapie): use api call
```
jacobperron, line 97: I think this TODO still needs to be resolved.

Looks like we're calling the list parameters service twice now.

ros2param/ros2param/api/__init__.py
ihasdapie: Runtime error would occur if we hit an unexpected type. The only place where `get_value` is called in [`dump`](https://github.com/ros2/ros2cli/blob/faf3d052fc5bdc1cf981b5b1f49e5b9dc7b10148/ros2param/ros2param/verb/dump.py#L67) where I don't expect an unexpected type to ever appear.


I did it this way so that it matches the behaviour of the function before my refactor

jacobperron: Makes sense to me!

ros2param/ros2param/verb/list.py
```diff
@@ -69,53 +72,20 @@ def main(self, *, args):  # noqa: D102
...
-
-                if len(futures) == len(clients):
-                    break
-                rclpy.spin_once(node, timeout_sec=1.0)
-
-            # wait for all responses
-            for future in futures.values():
-                rclpy.spin_until_future_complete(node, future, timeout_sec=1.0)
-
+                responses[node_name] = call_list_parameters(node=node, node_name=node_name.name)
```
jacobperron, line 78: I think we should use the fully-qualified node name here.
```suggestion
                responses[node_name] = call_list_parameters(node=node, node_name=node_name.full_name)
```

ros2param/ros2param/api/__init__.py
```diff
@@ -13,113 +13,45 @@
...
-def load_parameter_dict(*, node, node_name, parameter_dict):
-
-    parameters = parse_parameter_dict(namespace='', parameter_dict=parameter_dict)
-    response = call_set_parameters(
-        node=node, node_name=node_name, parameters=parameters)
+def load_parameter_file(*, node, node_name, parameter_file, use_wildcard):
+    # Remove leading slash and namespaces
+    client = AsyncParameterClient(node, node_name)
+    client.wait_for_services(timeout_sec=5.0)
+    if not client.services_are_ready():
```
jacobperron, line 47: I think this is equivalent:
```suggestion
    if not client.wait_for_services(timeout_sec=5.0):
```

ros2param/ros2param/api/__init__.py
```diff
@@ -135,103 +67,46 @@ def load_parameter_dict(*, node, node_name, parameter_dict):
...
-    if not ready:
-        raise RuntimeError('Wait for service timed out')
-
-    request = DescribeParameters.Request()
-    if parameter_names:
-        request.names = parameter_names
-    future = client.call_async(request)
+    client = AsyncParameterClient(node, node_name)
+    client.wait_for_services(timeout_sec=5.0)
+    if not client.services_are_ready():
```
jacobperron, line 73:
```suggestion
    if not client.wait_for_services(timeout_sec=5.0):
```

ros2param/ros2param/api/__init__.py
```diff
@@ -135,103 +67,46 @@ def load_parameter_dict(*, node, node_name, parameter_dict):
...
-    ready = client.wait_for_service(timeout_sec=5.0)
-    if not ready:
-        raise RuntimeError('Wait for service timed out')
-
-    request = GetParameters.Request()
-    request.names = parameter_names
-    future = client.call_async(request)
+    client = AsyncParameterClient(node, node_name)
+    client.wait_for_services(timeout_sec=5.0)
+    if not client.services_are_ready():
```
jacobperron, line 84:
```suggestion
    if not client.wait_for_services(timeout_sec=5.0):
```

ros2param/ros2param/api/__init__.py
```diff
@@ -135,103 +67,46 @@ def load_parameter_dict(*, node, node_name, parameter_dict):
...
-    ready = client.wait_for_service(timeout_sec=5.0)
-    if not ready:
-        raise RuntimeError('Wait for service timed out')
-
-    request = SetParameters.Request()
-    request.parameters = parameters
-    future = client.call_async(request)
+    client = AsyncParameterClient(node, node_name)
+    client.wait_for_services(timeout_sec=5.0)
+    if not client.services_are_ready():
```
jacobperron, line 95:
```suggestion
    if not client.wait_for_services(timeout_sec=5.0):
```

ros2param/ros2param/api/__init__.py
```diff
@@ -135,103 +67,46 @@ def load_parameter_dict(*, node, node_name, parameter_dict):
...
-
-    request = ListParameters.Request()
-    future = client.call_async(request)
-    rclpy.spin_until_future_complete(node, future)
-
-    # handle response
+def call_list_parameters(*, node, node_name):
+    client = AsyncParameterClient(node, node_name)
+    client.wait_for_services(timeout_sec=5.0)
+    if not client.services_are_ready():
```
jacobperron, line 106:
```suggestion
    if not client.wait_for_services(timeout_sec=5.0):
```

ros2param/ros2param/api/__init__.py
ihasdapie: It's used in https://github.com/ros2/ros2cli/blob/4d6463b7442d8f5b7945c92182826d12acd7bdc3/ros2param/ros2param/verb/list.py#L76

ihasdapie:
CI: 
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16849/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11427/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17257/)

ros2param/ros2param/api/__init__.py
jacobperron: Oops :upside_down_face:

ihasdapie:
See: https://github.com/ros2/rclpy/pull/959#issuecomment-1157084353

Running again with increased timeout and more `--retest-until-pass`

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16851/)

ihasdapie:
Hopefully removing the busy wait and streamlining the `dump` verb helps those tests pass...

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16860/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11435/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17265/)

ihasdapie:
CI is passing for this PR, failures unrelated

clalancette:
> CI is passing for this PR, failures unrelated

Just to be clear, that does not seem to be the case.  All of the nightly builds on Linux and Windows are green, and the [last CI](https://github.com/ros2/ros2cli/pull/716#issuecomment-1158266401) was not, so it does look like there is breakage here.

ihasdapie:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16862/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11437/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17267/)

@clalancette Windows error seemed to be a weird flaky test. Yellow because of downstream linting error on one of my related PRs I'm still working on.

ros2param/ros2param/api/__init__.py
```diff
@@ -135,104 +63,51 @@ def load_parameter_dict(*, node, node_name, parameter_dict):
...
-    # call as soon as ready
-    ready = client.wait_for_service(timeout_sec=5.0)
-    if not ready:
-        raise RuntimeError('Wait for service timed out')
-
-    request = ListParameters.Request()
-    future = client.call_async(request)
+def call_list_parameters(*, node, node_name, prefixes=None):
+    client = AsyncParameterClient(node, node_name)
+    client.wait_for_services(timeout_sec=5.0)
```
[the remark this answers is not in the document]
jacobperron: Note, the same issue applies to all other cases where we're waiting for services in this file.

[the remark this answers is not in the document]
jacobperron: You can reproduce this scenario by created a node without parameter services, e.g.

```py
rclpy.init()
node = rclpy.node.Node('foo', start_parameter_services=False)
rclpy.spin(node)
```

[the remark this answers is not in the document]
ihasdapie: The behaviour before varied depending on whether if the parameter service call was being made in the verb (no error handling in `load`, blocking forever in `list`, and raising a `RuntimeError` for those in `api/__init__.py`). I've unifed the behaviour in https://github.com/ros2/ros2cli/pull/716/commits/68b982046a34b41a57c4272a93bf73201324282b to raise a `RuntimeError` whenever it times out.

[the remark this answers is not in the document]
jacobperron: Thanks for making it consistent. I think it's fine to raise a runtime error, but I think the behavior from the user-perspective is not quite right with respect to `ros2 param list`. If we have multiple nodes and one of them does not provide parameter services, then we don't see any parameters listed for the nodes that _do_ offer parameter services.

ros2param/ros2param/verb/list.py
```diff
@@ -69,53 +68,22 @@ def main(self, *, args):  # noqa: D102
...
-                    break
-                rclpy.spin_once(node, timeout_sec=1.0)
-
-            # wait for all responses
-            for future in futures.values():
-                rclpy.spin_until_future_complete(node, future, timeout_sec=1.0)
-
+                responses[node_name] = call_list_parameters(node=node,
+                                                            node_name=node_name.full_name,
+                                                            prefixes=args.param_prefixes)
```
jacobperron, line 76: style nit:
```suggestion
                responses[node_name] = call_list_parameters(
                    node=node,
                    node_name=node_name.full_name,
                    prefixes=args.param_prefixes)
```

ros2param/ros2param/api/__init__.py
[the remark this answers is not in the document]
ihasdapie: That's a good point. I've pushed changes to make `ros2 param list` handle nodes without parameter services gracefully

ros2param/ros2param/verb/dump.py
```diff
@@ -94,36 +89,28 @@ def main(self, *, args):  # noqa: D102
...
-                    pval = self.get_parameter_value(node, absolute_node_name, param_name)
-                    self.insert_dict(
-                        yaml_output[node_name.full_name]['ros__parameters'], param_name, pval)
-            else:
-                e = future.exception()
+            response = call_list_parameters(node=node, node_name=absolute_node_name)
+            if response is None:
+                raise RuntimeError(
+                    'Wait for service timed out waiting for '
+                    f'parameter services for node {node_name}')
```
jacobperron, line 99:
```suggestion
                    f'parameter services for node {node_name.full_name}')
```

jacobperron (commented):
The behavior for `ros2 param list` is much better now, thanks!

jacobperron (approved):
LGTM with green CI

ihasdapie:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17008/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11569/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17423/)

ihasdapie:
This internal api was deprecated in favor of rclpy's `AsyncParameterClient`'s `set_parameter` method: 
```python
from rclpy.parameter_client import AsyncParameterClient

client = AsyncParameterClient(node, node_name)
future = client.set_parameters(list(param_dict.values()))
rclpy.spin_until_future_complete(node, future)
response = future.result()
# ... and so forth
```



Reference:
- https://github.com/ros2/rclpy/blob/1ee5dff998d0e34cfbe0897b66952d5b270ada79/rclpy/rclpy/parameter_client.py#L169
- https://github.com/ros2/rclpy/blob/1ee5dff998d0e34cfbe0897b66952d5b270ada79/rclpy/rclpy/parameter_client.py#L289
- https://github.com/ros2/rclpy/blob/1ee5dff998d0e34cfbe0897b66952d5b270ada79/rclpy/rclpy/parameter.py#L265-L314

Kaju-Bubanja:
@ihasdapie Any reason for that? Did something not work with the function or cause problems? Otherwise I don't see why a perfectly fine function should be removed.

ihasdapie:
The methods in ros2param duplicated code and had inconsistent implementations which this refactor addressed. Plus, we didn't anticipate external users using the internal ros2param methods. If you think that this method should be put back into ros2param feel free to open a PR and we can discuss there, but in my opinion the current design consolidating the parameter management logic in `rclpy` and calling out to it in `ros2param` makes more sense.

Kaju-Bubanja:
Hmmm, I found the `parse_parameter_dict` function very convenient and copy pasted it into my file, in case somebody in the future searches for this issue here you go, important is to also allow_undeclared_parameters:

```
def parse_parameter_dict(*, namespace, parameter_dict):
    parameters = []
    for param_name, param_value in parameter_dict.items():
        full_param_name = namespace + param_name
        # Unroll nested parameters
        if type(param_value) == dict:
            parameters += parse_parameter_dict(
                    namespace=full_param_name + PARAMETER_SEPARATOR_STRING,
                    parameter_dict=param_value)
        else:
            parameter = Parameter(full_param_name, value=param_value)
            parameters.append(parameter)
    return parameters

class Foo(Node):
    def __init__(self):
        self.name = "foo"
        # allow_undeclared_parameters implicitly declares parameters when setting them, basically syntax sugar
        super().__init__(self.name, allow_undeclared_parameters=True)
    def load_parameter_dict(self, d):
        call_set_parameters(node=self, node_name=self.name, parameters=parse_parameter_dict(namespace="", parameter_dict=d))
```

PR #720, opened 2022-06-08
Add netperf, an iperf like tool

ivanpauno:
Example use:

```bash
# terminal 1, server
ros2 netperf server
```

```bash
# terminal 2, client
# 10seconds experiment, 1024 bytes message size, 1Mbps target bandwidth
ros2 netperf client -d 10 -s 1024 -b 1
```

This could use some unit testing, that I'm planning to add before merging.
But, it's already ready to be reviewed.


netperf_tool/CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
+cmake_minimum_required(VERSION 3.5)
+project(netperf_tool)
+
+# Default to C++14
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 17)
+endif()
```
wjwwood, line 7: typo?

```suggestion
# Default to C++17
if(NOT CMAKE_CXX_STANDARD)
  set(CMAKE_CXX_STANDARD 17)
endif()
```

netperf_tool/CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
...
+
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  # About -Wno-sign-conversion: With Clang, -Wconversion implies -Wsign-conversion. There are a number of
+  # implicit sign conversions in rclcpp and gtest.cc, see https://ci.ros2.org/job/ci_osx/9265/.
+  # Hence disabling -Wsign-conversion for now until all those have eventually been fixed.
+  # (from https://github.com/ros2/rclcpp/pull/1188#issuecomment-650229140)
+  add_compile_options(-Wall -Wextra -Wconversion -Wno-sign-conversion -Wpedantic -Wnon-virtual-dtor -Woverloaded-virtual)
```
wjwwood, line 14: Maybe this can be removed for this package. Seems rclcpp specific.

netperf_tool/CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
...
+_ament_cmake_python_get_python_install_dir()
+find_package(ament_cmake_ros REQUIRED)
+find_package(rclcpp REQUIRED)
+find_package(netperf_tool_interfaces REQUIRED)
+
+# Find python before pybind11
+find_package(Python3 REQUIRED COMPONENTS Interpreter Development)
+
+find_package(pybind11_vendor REQUIRED)
+find_package(pybind11 REQUIRED)
```
wjwwood, line 37: It seems odd to find Python two different ways. I'm not sure if this is the best practice or not?

netperf_tool/netperf_tool/__init__.py
```diff
@@ -0,0 +1,88 @@
...
+from rpyutils import add_dll_directories_from_env
+
+# Since Python 3.8, on Windows we should ensure DLL directories are explicitly added
+# to the search path.
+# See https://docs.python.org/3/whatsnew/3.8.html#bpo-36085-whatsnew
+with add_dll_directories_from_env('PATH'):
+    from netperf_tool.netperf_tool_impl import (
+        ClientRunner as _ClientRunner,
+        ServerRunner as _ServerRunner,
+    )
```
wjwwood, line 27: nitpick: import one thing per line

```suggestion
    from netperf_tool.netperf_tool_impl import ClientRunner as _ClientRunner
    from netperf_tool.netperf_tool_impl import ServerRunner as _ServerRunner
```

netperf_tool/netperf_tool/__init__.py
```diff
@@ -0,0 +1,88 @@
...
+
+class ServerRunner(RunnerImpl):
+    _impl_cls = _ServerRunner
+
+    def __init__(
+        self,
+        *,
+        qos: rclpy.qos.QoSProfile
+    ):
+        super().__init__(-1e-9, qos)
```
wjwwood, line 82: This is an odd magic number, maybe it needs a comment or should be a constant. I can't really tell what this is supposed to mean (just looking here).

netperf_tool/package.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="2">
```
wjwwood, line 3: Maybe do format 3?

netperf_tool/package.xml
```diff
@@ -0,0 +1,33 @@
...
+  <name>netperf_tool</name>
+  <version>0.19.0</version>
```
wjwwood, line 5: Why 0.19.0?

netperf_tool/package.xml
```diff
@@ -0,0 +1,33 @@
...
+  <description>
+    Simple iperf like tool for ros2.
+  </description>
+
+  <maintainer email="aditya.pande@openrobotics.org">Aditya Pande</maintainer>
+  <maintainer email="audrow@openrobotics.org">Audrow Nash</maintainer>
+  <maintainer email="michael.jeronimo@openrobotics.org">Michael Jeronimo</maintainer>
```
wjwwood, line 12: Is this intentional due to it being in `ros2cli`?

netperf_tool/package.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="2">
+  <name>netperf_tool</name>
```
wjwwood, line 4: Why not call it `ros2netperf` like the other cli tools?

netperf_tool/src/client_node.cpp
```diff
@@ -0,0 +1,160 @@
...
+#include <functional>
+#include <memory>
+#include <mutex>
+#include <string>
+#include <unordered_map>
+#include <utility>
+#include <vector>
+
+
+#include "netperf_tool_interfaces/msg/bytes.hpp"
```
wjwwood, line 27:
```suggestion
#include <vector>

#include "netperf_tool_interfaces/msg/bytes.hpp"
```

nitpick: minimize vertical whitespace (https://google.github.io/styleguide/cppguide.html#Vertical_Whitespace)

netperf_tool/src/client_node.cpp
```diff
@@ -0,0 +1,160 @@
...
+    }
+    RCLCPP_INFO(this->get_logger(), "perf server not available yet, waiting...");
+    std::this_thread::sleep_for(500ms);
+  }
+  RCLCPP_INFO_STREAM(
+    this->get_logger(),
+    "perf server not available after waiting ["
+      << std::chrono::duration_cast<std::chrono::seconds>(timeout).count() << "s]" << std::endl);
+  return false;
+}
```
wjwwood, line 73: Did you consider using the `wait_for_service()` method on the client class?

https://github.com/ros2/rclcpp/blob/38581cc860fc98509c8c6654ab89a79c7dc045c0/rclcpp/include/rclcpp/client.hpp#L188-L201

netperf_tool/src/client_node.cpp
```diff
@@ -0,0 +1,160 @@
...
+  if (timer_) {
+    timer_->cancel();
+  }
+}
+
+void
+ClientNode::sync_with_server(rclcpp::Executor & exec)
+{
+  if (!pub_->wait_for_all_acked(5s)) {
+    RCLCPP_WARN(this->get_logger(), "Some messages were not acked by the perf server ...");
```
wjwwood, line 95: I think you should tidy up the terminology in these messages, "perf server" might be confusing to someone, instead consider using "netperf server", as that is what the users actually run (`ros2 netperf server`).

wjwwood: nitpick:

```suggestion
    RCLCPP_WARN(this->get_logger(), "Some messages were not acknowledged by the perf server...");
```

netperf_tool/src/client_node.cpp
```diff
@@ -0,0 +1,160 @@
...
+{
+  if (timer_) {
+    timer_->cancel();
+  }
+}
+
+void
+ClientNode::sync_with_server(rclcpp::Executor & exec)
+{
+  if (!pub_->wait_for_all_acked(5s)) {
```
wjwwood, line 94: `5s` seems arbitrary, should it be configurable? Should it be part of the error message so it is observable?

netperf_tool/src/client_node.cpp
```diff
@@ -0,0 +1,160 @@
...
+std::string
+ClientNode::get_stringified_pub_gid()
+{
+  return stringify_gid(pub_->get_gid());
+}
+
+void
+ClientNode::pub_next_msg()
+{
+  netperf_tool_interfaces::msg::Bytes msg;
```
wjwwood, line 141: Could you not reuse this message loop-to-loop? You'd be able to reuse the serialized size too, since the id and timestamp shouldn't change size (afaik).

netperf_tool/src/client_node.hpp
```diff
@@ -0,0 +1,95 @@
...
+#include "rclcpp/serialized_message.hpp"
+
+namespace netperf_tool
+{
+struct ClientCollectedInfo
+{
+  std::vector<size_t> message_ids;
+  std::vector<std::chrono::time_point<std::chrono::system_clock>> message_published_times;
+  std::vector<size_t> message_sizes;
+};
```
wjwwood, line 40: Instead of a struct with three vectors, why not a vector of structs with three entries? Do you ever expect these vectors to have different lengths?

netperf_tool/src/client_node.hpp
```diff
@@ -0,0 +1,95 @@
...
+  uint64_t next_id{};
+  std::shared_ptr<rclcpp::TimerBase> timer_;
+  rclcpp::Publisher<netperf_tool_interfaces::msg::Bytes>::SharedPtr pub_;
+  rclcpp::Client<netperf_tool_interfaces::srv::GetResults>::SharedPtr client_;
+  ClientCollectedInfo collected_info_;
+  mutable std::mutex collected_info_mutex_;
+  rclcpp::Serialization<netperf_tool_interfaces::msg::Bytes> serializer_;
+  netperf_tool_interfaces::srv::GetResults::Response::SharedPtr statistics_;
+  std::chrono::nanoseconds target_pub_period_;
+};
```
wjwwood, line 93: docstrings would be good, even if they are trivial

wjwwood (commented):
partial review, will continue later

netperf_tool/CMakeLists.txt
ivanpauno: Yes, I copy pasted incorrectly.

```suggestion
  add_compile_options(-Wall -Wextra -Wconversion -Wsign-conversion -Wpedantic -Wnon-virtual-dtor -Woverloaded-virtual)
```

netperf_tool/CMakeLists.txt
ivanpauno: I think that we're doing this in other places as well, if not it breaks or Windows debug builds.
If all our platforms would use cmake 3.16 or later, we could use `Python3_FIND_ABI` from [FindPython3](https://cmake.org/cmake/help/latest/module/FindPython3.html) which should work.
Before that version, I think there wasn't a way to make `FindPython3` find the python debug version when both ABIs were available, except for already specifying `Python3_EXECUTABLE` beforehand.

According to rep2000, all our platforms already support cmake 3.16, except for macOS (which uses 3.14.4).
If we don't mind breaking that, I will try using the `Python3_FIND_ABI` hint.

netperf_tool/netperf_tool/__init__.py
ivanpauno: Sorry about that, I only need a negative number really.
The python value is a floating point number in seconds, which is converted to `std::chrono::nanoseconds`.
In this case, `-1e9` is `-1ns` in python.
I could add a comment, or instead use `-1.`, which is less confusing.

See https://pybind11.readthedocs.io/en/stable/advanced/cast/chrono.html#provided-conversions.

ivanpauno: I could maybe use `datetime.timedelta` instead of floating point numbers to avoid confusion as well.

netperf_tool/package.xml
ivanpauno:
```suggestion
<package format="3">
```

netperf_tool/package.xml
ivanpauno: IIUC, we have to use the same version number for all packages in the same repo.
And currently we're on [0.19.0](https://github.com/ros2/ros2cli/blob/master/ros2doctor/package.xml).

Maybe I can just leave it as `0.0.0` and it will be fixed in the next `catkin_prepare_release`, I'm not sure how it works.

netperf_tool/package.xml
ivanpauno: Yes, I always copy the maintainers of the repo.
We could create a new repo for this if that's preferred though.
None of the packages I'm adding need to live in this repo.

netperf_tool/package.xml
clalancette: The version number has to match all others in this repository, otherwise we won't be able to do a release.

Obviously this doesn't apply if we decide to put it in a different repository.

netperf_tool/src/client_node.cpp
ivanpauno: I forgot about that, I think it should work.

netperf_tool/src/client_node.cpp
ivanpauno: > I think you should tidy up the terminology in these messages, "perf server" might be confusing to someone, instead consider using "netperf server", as that is what the users actually run (ros2 netperf server).

Yes, I started calling the packages `perf_tool/ros2perf` and later renamed them to `netperf_tool/ros2netperf`.
It seems that I forgot to replace the old name for the new one in some places.

```suggestion
    RCLCPP_WARN(this->get_logger(), "Some messages were not acknowledged by the netperf server ...");
```

netperf_tool/src/client_node.cpp
ivanpauno: Sounds good, I will add that.

netperf_tool/src/client_node.cpp
ivanpauno: > Could you not reuse this message loop-to-loop?

Yes, I will try that.

> You'd be able to reuse the serialized size too, since the id and timestamp shouldn't change size (afaik).

Agreed.
I think I can try to calculate the serialized message size once in the constructor, and then I will not need to deal with serialization when publishing.
I will try that.

netperf_tool/src/client_node.hpp
ivanpauno: > Do you ever expect these vectors to have different lengths?

No, they are always the same length.

> Instead of a struct with three vectors, why not a vector of structs with three entries?

Both are fine IMO.
In `ServerCollectedInfo`, it was nice to have an struct of arrays to calculate the statistics easily, though I can also do that with the array of structs
Here, I imitated what `ServerCollectedInfo` uses.

Actually, I think I'm not using `ClientCollectedInfo` for anything.
The idea of collecting everything is that I could generate a csv with the raw data from it, which I think it's nice, but I haven't added that.

Same for `ServerCollectedInfo`, if we don't want to have the raw data available, I can just calculate the running statistics instead of collection all the data.

But I think that having the possibility to export the raw data to a csv in the future is nice.

netperf_tool/src/client_node.hpp
ivanpauno: Yes, sorry about that.
I will add docs.

netperf_tool/package.xml
clalancette: Personally, I think it would be better to have this in an external repository.  There are a few reasons for that:

1. As it stands, the `ros2cli` repository is purely Python.  While we don't *have* to maintain that, it does make it easier to for maintainers to only have to deal with a single language in a repository.
2. If we put it in an external repository, we can still release it into the distributions, but it doesn't necessarily need to be part of the core.  That allows it to evolve independently of the core, and also makes it so that it isn't release-critical.

That said, if others feel strongly that it should be here, I can be persuaded otherwise.

ivanpauno: I don't mind stronly either way.
I opened a PR targetting this repo only as a first step.

If we agree to move it to another repo, I think it would still be good to finish addressing all the comments here and then move the code to another repo.

netperf_tool/src/client_node.cpp
ivanpauno: Done in https://github.com/ros2/ros2cli/pull/720/commits/625cb82afd0c9f306bcab93f2471b71e26dc9c9f

netperf_tool/src/client_node.cpp
ivanpauno: See https://github.com/ros2/ros2cli/pull/720/commits/34cf852d446ccdf94a29ec0cb2c08d2f68e477a8 and https://github.com/ros2/ros2cli/pull/720/commits/265d4801f24fb2bdd0abc3842be294a6f669ae96

netperf_tool/src/client_node.cpp
ivanpauno: Done in https://github.com/ros2/ros2cli/pull/720/commits/4cb873ac81d49b34edb5034efc32911fd20bc9ab.

netperf_tool/src/client_node.hpp
ivanpauno: After applying https://github.com/ros2/ros2cli/pull/720/commits/4cb873ac81d49b34edb5034efc32911fd20bc9ab, I think that this could maybe be:

```cpp
{
  size_t n_messages_published;
  size_t serialized_message_size;
  std::vector<std::chrono::time_point<std::chrono::system_clock>> message_published_times;
};
```

as the messages id are `[0, n_messages_published - 1]` and `serialized_message_size` is a constant.

In the server side, I think it's nice to collect everything:
- It's worth to collect the ids as messages can be lost, to have the complete raw data available.
- It's worth to collect the message size, to be ready for clients doing strange things (though it shouldn't actually change).

I don't mind between array of structs or struct of arrays in that case.
Let me know what you think.

netperf_tool/netperf_tool/__init__.py
wjwwood: Whatever you do, it's not clear from this file why it's a negative number.

What not just `-1.0` or even better a constant like `BLOCK_FOREVER`?

netperf_tool/package.xml
wjwwood: Oh right, my bad.

netperf_tool/package.xml
wjwwood: I'm in favor of putting it in a separate repository for the same reasons as @clalancette, but it does add some support overhead. Thought I suppose that is reduced slightly now that we have rolling.

wjwwood (commented):
Another partial review, syncing because things have changed since I started this review.

netperf_tool/src/client_node.hpp
wjwwood: Just thinking from the perspective of data processing, i.e. "DataFrames" which a pervasive idea in things like `pandas`, the array of structs is more natural to me. With each struct being a single record in time. Especially since the data within the struct is correlated, i.e. `message_ids[x]` is associated with `message_published_times[x]`.

wjwwood:
Sorry I haven't finished my review, I'll try to get back to it tomorrow.

netperf_tool/CMakeLists.txt
jacobperron: Are we still supporting macOS?

netperf_tool/src/client_node.cpp
```diff
@@ -0,0 +1,146 @@
...
+{
+  ClientResults results;
+  if (!statistics_) {
+    throw std::runtime_error{"you must first completely run the client before getting results"};
+  }
+  results.statistics = *statistics_;
+  statistics_.reset();
+  {
+    std::lock_guard guard{collected_info_mutex_};
+    results.collected_info = std::move(collected_info_);
```
jacobperron, line 114: I'm wondering if the mutex is necessary. In what scenario can we run into a race? IIUC, the timer should already be canceled so we shouldn't be accessing this in another thread by this point. In the event that the user prematurely terminates the program, I think `statistics_` would not be set anyways and we get an exception before this line. If `statistics_` is set, then that means we successfully ran `sync_with_server` and the publisher has stopped. If there _is_ a race, then I would expect there to be a race with the access to `collected_info_` in `sync_with_server`.

ros2netperf/ros2netperf/verb/server.py
```diff
@@ -0,0 +1,75 @@
...
+        self.gid_id_map_ = {}
+
+    def add_arguments(self, parser, cli_name):
+        add_qos_arguments_to_parser(parser)
+
+
+    def main(self, *, args):
+        qos_profile = get_qos_profile_from_args(args)
+
+        runner = netperf_tool.ServerRunner(qos=qos_profile)
```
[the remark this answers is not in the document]
ivanpauno: https://github.com/ros2/ros2cli/blob/4cb873ac81d49b34edb5034efc32911fd20bc9ab/netperf_tool/src/netperf_tool.cpp#L59
https://github.com/ros2/ros2cli/blob/4cb873ac81d49b34edb5034efc32911fd20bc9ab/netperf_tool/src/netperf_tool.cpp#L85-L91

> I can't find where this is defined, I see ClientRunner, which is confusingly part node_runner.hpp, but not ServerRunner (I'm just grepping the source).

ServerRunner is just a template instantation of `NodeRunner`.
I agree how it's currently organized is a bit confusing.
I could move `ClientRunner` to `netperf_tool.cpp`, or create both `client_runner.hpp/cpp` and `server_runner.hpp`.
The second will only have a using statement.

netperf_tool/src/client_node.cpp
ivanpauno: > IIUC, the timer should already be canceled so we shouldn't be accessing this in another thread by this point

Yes, the mutex seems to be unnecessary in `ClientNode`.
There's a similar mutex in `ServerNode` though, and in that case it's necessary because `extract_results()` is called concurrently while an executor spins.
I could remove it from one and keep it in the other.

netperf_tool/netperf_tool/__init__.py
ivanpauno: Done, see https://github.com/ros2/ros2cli/pull/720/commits/c14657702684f996e694ee718896272250e409f8

netperf_tool/src/client_node.hpp
ivanpauno: Done in https://github.com/ros2/ros2cli/pull/720/commits/d9d34245e333372716f3200555710f482484c588.

ros2netperf/ros2netperf/verb/server.py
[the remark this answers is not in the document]
ivanpauno: Done, see https://github.com/ros2/ros2cli/pull/720/commits/6cfd521fe29fe943ef57873957a172dfb9fc9477.

netperf_tool/src/node_runner.cpp
```diff
@@ -0,0 +1,68 @@
...
+void
+ClientRunner::join()
+{
+  bool sync_needed = running_thread_.joinable();
+  NodeRunner<ClientNode>::join();
+  if (sync_needed) {
+    node_->stop_publishing();
+    node_->sync_with_server(exec_);
+  }
+}
```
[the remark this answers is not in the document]
ivanpauno: Moved, see https://github.com/ros2/ros2cli/pull/720/commits/6cfd521fe29fe943ef57873957a172dfb9fc9477.

netperf_tool/CMakeLists.txt
ivanpauno: >  Either way, the cmake version isn't an issue for macOS.

I didn't know that.
I guess that rep 2000 is indicating that because that was the version of cmake we used to use in macOS CI.
But we're not running that anymore, so I think that requiring a newer minimal cmake version it's worth it.
I think that I'm going to do that in a different PR though, as if not I will block this on getting windows debug builds working (which is always not completely straightforward).

netperf_tool/netperf_tool/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+
+# Since Python 3.8, on Windows we should ensure DLL directories are explicitly added
+# to the search path.
+# See https://docs.python.org/3/whatsnew/3.8.html#bpo-36085-whatsnew
+with add_dll_directories_from_env('PATH'):
+    from netperf_tool.netperf_tool_impl import ClientRunner as _ClientRunner
+    from netperf_tool.netperf_tool_impl import ServerRunner as _ServerRunner
+
+
+class RunnerImpl:
```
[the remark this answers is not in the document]
ivanpauno: Done, see https://github.com/ros2/ros2cli/pull/720/commits/c14657702684f996e694ee718896272250e409f8

netperf_tool/src/client_node.hpp
ivanpauno: Done, see https://github.com/ros2/ros2cli/pull/720/commits/53a16a0d960a83c98284dee94d6e51c0a8756e5a and https://github.com/ros2/ros2cli/pull/720/commits/e9212d6fd5cf30b4cf9d7daa111ab29f8be8d27a.

ivanpauno:
@jacobperron @wjwwood PTAL.
I think I've addressed or replied to all your previous comments.

netperf_tool/CMakeLists.txt
wjwwood: > I guess that rep 2000 is indicating that because that was the version of cmake we used to use in macOS CI.

I think that's right.

Can you bring this up at the next ROS 2 meeting? It sounds like it would be an improvement for all of our packages that use Python.

On VIPER we run into the issue that finding python in two ways caused us to find two different python versions in a single build, which broke it.

ivanpauno: > On VIPER we run into the issue that finding python in two ways caused us to find two different python versions in a single build, which broke it.

Oh, that's interesting ...
Sure, I can bring that up

ivanpauno:
@jacobperron @wjwwood friendly ping

netperf_tool_interfaces/msg/Bytes.msg
```diff
@@ -0,0 +1,3 @@
+uint64  timestamp
```
wjwwood, line 1: I'm curious why you don't use the `builtin_interfaces/msg/Time` type?

ros2netperf/ros2netperf/verb/client.py
```diff
@@ -0,0 +1,90 @@
...
+            message_size_bytes=args.message_size,
+            target_pub_period_s=pub_period,
+            server_timeout_s=args.server_timeout)
+        try:
+            with runner as node:
+                print('---------------------------------------------------------')
+                print('Client running')
+                print(f'\ttopic: {node.get_topic_name()}')
+                print(f'\tpublisher gid: {node.get_stringified_pub_gid()}')
+                print(f'\tqos: {qos_profile}')
```
wjwwood, line 77: For leading whitespace, just use spaces, I think:

```suggestion
                print(f'    topic: {node.get_topic_name()}')
                print(f'    publisher gid: {node.get_stringified_pub_gid()}')
                print(f'    qos: {qos_profile}')
```

ros2netperf/ros2netperf/verb/server.py
```diff
@@ -0,0 +1,75 @@
...
+        try:
+            with runner as node:
+                print('---------------------------------------------------------')
+                print('Server running')
+                print(f'\ttopic: {node.get_topic_name()}')
+                print(f'\tqos: {qos_profile}')
```
wjwwood, line 44:
```suggestion
                print(f'    topic: {node.get_topic_name()}')
                print(f'    qos: {qos_profile}')
```

ros2netperf/ros2netperf/verb/server.py
```diff
@@ -0,0 +1,75 @@
...
+        for client_gid in clients_gids:
+            self.gid_id_map_[client_gid] = self.next_client_id_
+            print(f'[ {self.next_client_id_}] Publisher with gid [{client_gid}] connected')
+            self.next_client_id_ = self.next_client_id_ + 1
+        results_map = node.extract_results()
+        self.print_results(results_map)
+
+
+    def print_results(self, results_map):
+        # TODO(ivanpauno): Add some processing to be able to show better statistics
```
wjwwood, line 69: Is this TODO still pending? What could you do differently here? Why not do it now?

netperf_tool_interfaces/msg/Bytes.msg
ivanpauno: It was easier, so I don't need to convert the timestamps in its `sec` and `nsec` components.

ros2netperf/ros2netperf/verb/server.py
ivanpauno: I think I forgot to delete that comment.

```suggestion
```

ivanpauno: Solved in https://github.com/ros2/ros2cli/pull/720/commits/1efbc0fffd83fa23d1a6bf46e01d7ad6fe927317

ros2netperf/ros2netperf/verb/client.py
ivanpauno: Done in https://github.com/ros2/ros2cli/pull/720/commits/4240ee13173a7cc94f7d1036913c74efd3b61823

ros2netperf/ros2netperf/verb/server.py
ivanpauno: Done in https://github.com/ros2/ros2cli/pull/720/commits/0d98122212fed5ca199b81f1d472888873f9ada3

netperf_tool_interfaces/package.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>netperf_tool_interfaces</name>
+  <version>0.19.0</version>
+  <description>
+    Interface definitions used by the ROS 2 iperf like tool.
```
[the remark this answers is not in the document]
ivanpauno:
```suggestion
    Interface definitions used by the ROS 2 netperf tool.
```

netperf_tool/netperf_tool/__init__.py
```diff
@@ -0,0 +1,138 @@
...
+        """
+        Construct.
+
+        :param experiment_duration: Total time to be spinning the node.
+            A negative number can be used to spin for ever, e.g. `cls._BLOCK_FOREVER`.
+        :param qos: Qos profile used to create the publisher or subscription.
+        :param *args: Other arguments required to construct the node.
+        """
+        if self._impl_cls is None or not isclass(self._impl_cls):
+            raise RuntimeError('internal error')
```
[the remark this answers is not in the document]
ivanpauno: Fixed in https://github.com/ros2/ros2cli/pull/720/commits/b8224875d579d9af4f646d417dd6b3c01f1dbcd0

netperf_tool/netperf_tool/__init__.py
```diff
@@ -0,0 +1,138 @@
...
+
+    _BLOCK_FOREVER = -1.0
+    """
+    Duration used to indicate the runner to block forever.
+    Any other negative number work as well.
+    """
+
+    def __init__(
+        self,
+        experiment_duration: float,
```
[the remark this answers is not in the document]
ivanpauno: Done https://github.com/ros2/ros2cli/pull/720/commits/1dedb0f9192ab5f668ea2beccf2f6c6d55353c4c

netperf_tool/src/client_node.cpp
```diff
@@ -0,0 +1,148 @@
...
+ClientNode::get_stringified_pub_gid()
+{
+  return stringify_gid(pub_->get_gid());
+}
+
+void
+ClientNode::pub_next_msg()
+{
+  auto now = std::chrono::system_clock::now();
+  msg_to_publish_.timestamp = static_cast<size_t>(now.time_since_epoch().count());
```
wjwwood, line 138: For example, if you used `builtin_interfaces/msg/Time` here you could just do:

```c++
msg_to_publish_.timestamp = this->now();
```

And if you wanted to ensure System time (rather than ROS Time which could be sim time), you can do that too:

```c++
auto now = rclcpp::Clock().now();  // default is system time, but could pass `RCL_SYSTEM_TIME`
msg_to_publish_.timestamp = now;
```

netperf_tool/src/server_node.cpp
```diff
@@ -0,0 +1,225 @@
...
+      "source and received timestamp supported, using them to calculate latency!");
+    auto pub_timestamp = std::chrono::time_point<std::chrono::system_clock>{
+      std::chrono::nanoseconds{rmw_msg_info.source_timestamp}};
+    rec_timestamp =
+      std::chrono::time_point<std::chrono::system_clock>{
+      std::chrono::nanoseconds{rmw_msg_info.received_timestamp}};
+    latency = rec_timestamp - pub_timestamp;
+  } else {
+    // TODO(ivanpauno): Maybe it's worth to always calculate this latency (?)
+    // as it takes in account the executor overhead.
```
wjwwood, line 76: Is this todo still active? Should we address this question before merging?

netperf_tool/src/server_node.cpp
```diff
@@ -0,0 +1,225 @@
...
+    RCLCPP_INFO_ONCE(
+      this->get_logger(),
+      "either source or received timestamp are not supported,"
+      "using timestamps added to the message to calculate latency");
+    auto pub_timestamp =
+      std::chrono::time_point<std::chrono::system_clock>{
+      std::chrono::nanoseconds{msg.timestamp}};
```
wjwwood, line 83: This could become:

```c++
    auto pub_timestamp =
      std::chrono::time_point<std::chrono::system_clock>{
      std::chrono::nanoseconds{rclcpp::Time(msg.timestamp).nanoseconds}};
```

netperf_tool_interfaces/msg/Bytes.msg
wjwwood: Is this really the right reason to avoid the Time type? We have facilities to handle the conversions for you. I left a few comments pointing out how.

netperf_tool/src/server_node.cpp
```diff
@@ -0,0 +1,225 @@
...
+    std::bind(
+      &ServerNode::handle_get_results_request,
+      this,
+      std::placeholders::_1,
+      std::placeholders::_2));
+}
+
+void
+ServerNode::handle_msg(
+  const rclcpp::SerializedMessage & serialized_msg,
```
wjwwood, line 53: Why are you using the serialized version of the callback? To try and avoid that playing into the latency calculations?

netperf_tool/src/server_node.cpp
ivanpauno: I think that calculating both when available might be useful, but considering the tool is about network performance (and not executor performance), I think it's out of scope.
Anyway, this can be addressed in a follow up, so I will delete the TODO.

netperf_tool/src/server_node.cpp
ivanpauno: To be able to get the serialized size easily, without having to serialize again.

> To try and avoid that playing into the latency calculations?

That's also a benefit.

netperf_tool/src/server_node.cpp
ivanpauno: Done in https://github.com/ros2/ros2cli/pull/720/commits/084b304892a29231a93a5a0f46fa8b9ad349b639

netperf_tool/src/client_node.cpp
ivanpauno: Fixed in https://github.com/ros2/ros2cli/pull/720/commits/084b304892a29231a93a5a0f46fa8b9ad349b639.

netperf_tool_interfaces/msg/Bytes.msg
ivanpauno: Done, see https://github.com/ros2/ros2cli/pull/720/commits/084b304892a29231a93a5a0f46fa8b9ad349b639

netperf_tool/src/server_node.cpp
ivanpauno: Removed the TODO in https://github.com/ros2/ros2cli/pull/720/commits/0669be244e8a9385a8a72bce1151702fae525f64.

ivanpauno:
@jacobperron @wjwwood friendly ping, could you take another look?

netperf_tool/src/client_node.cpp
wjwwood: What was the result of this comment chain?

netperf_tool_interfaces/msg/Bytes.msg
```diff
@@ -0,0 +1,3 @@
+builtin_interfaces/Time  timestamp
+uint64  id
+byte[]  data
```
wjwwood, line 3: Nitpick: Why are there two spaces between items here? Seems weird to me. Code linters from other languages would complain I think.

wjwwood (approved):
Generally lgtm, with one style question and one follow up question.

netperf_tool_interfaces/msg/Bytes.msg
ivanpauno: I don't know why I was using two spaces :joy:.
Fixed in https://github.com/ros2/ros2cli/pull/720/commits/923e5af693109d3772e234b85dd5fa88a4b15f72

ivanpauno:
@wjwwood @jacobperron let me know if there're any more comments.

@clalancette the idea is to add this tool to a new repo, I don't have permissions to do that. Could you create a `ros2/netperf` repo and give me access? Thanks!

clalancette:
> @clalancette the idea is to add this tool to a new repo, I don't have permissions to do that. Could you create a ros2/netperf repo and give me access? Thanks!

Done; see https://github.com/ros2/netperf .  You should have write access, but if you don't have enough permissions let me know.

ivanpauno:
@jacobperron @wjwwood do you have any other comment before I move this to the new repo?

jacobperron:
I may have some minor comments, but I can leave on the PR in the new repository.

wjwwood:
Fine with me to move it now.

ivanpauno:
Sounds good, I opened https://github.com/ros2/netperf/pull/1

Issue #721, opened 2022-06-09
Support setting parameters atomically using ros2 service command line


deepanshubansal01:

clalancette:
I'm also going to go ahead and close #727, which was an earlier version of this.

PR #727, opened 2022-06-23
Support setting parameters atomically

ihasdapie:
This PR implements a `--atomic` flag for `ros2 param set` for setting parameters atomically from the command line.

- Closes https://github.com/ros2/ros2cli/issues/721
- Depends on https://github.com/ros2/rclpy/pull/959
- This PR branches off https://github.com/ros2/ros2cli/pull/716 since it makes use of the code there. See https://github.com/ros2/ros2cli/compare/brianc/parameter_client...brianc/atomic_parameter_set for a more reasonable diff


clalancette:
> We should expose the `--atomic` flag for setting params atomically using `ros2 service call` #721 since we can set multiple params using service.

The problem with that, though, is that `--atomic` there only makes sense for setting parameters, not other service calls.  So I don't think the flag belongs on `ros2 service call`.

> Additionally, we can expose the `--atomic` flag for setting params using `ros2 param load` as well.

This we should definitely do.  The other thing we could do is to fix up `ros2 param set` so that it can take multiple parameters; at that point, it makes sense to add the `--atomic` flag.

ihasdapie:
Right, I definitely wasn't thinking when I put this PR together :sweat_smile:. Will update this PR to support multiple parameters & `param load --atomic`

clalancette:
> Will update this PR to support multiple parameters & `param load --atomic`

I'll suggest you do two separate PRs; the first one to add multiple param support to `set`, and then a second one (maybe this one) to add the `--atomic` flag.  They are logically independent things, so they should have different PRs.

ihasdapie:
`ros2 param set` with multiple parameters PR opened: https://github.com/ros2/ros2cli/pull/728

This PR is based off that PR (https://github.com/ros2/ros2cli/pull/728) and now implements `--atomic` for `set` and `load` verbs.
 
Diff: https://github.com/ros2/ros2cli/compare/brianc/multiple_param_set...brianc/atomic_parameter_set

ros2param/ros2param/api/__init__.py
```diff
@@ -66,6 +70,26 @@ def load_parameter_file(*, node, node_name, parameter_file, use_wildcard):
...
+        raise RuntimeError('Wait for service timed out')
+    future = client.load_parameter_file_atomically(parameter_file, use_wildcard)
+    parameters = list(parameter_dict_from_yaml_file(parameter_file, use_wildcard).values())
+    rclpy.spin_until_future_complete(node, future)
+    response = future.result()
+    if response is None:
+        raise RuntimeError('Exception while calling service of node '
+                           f'{node_name}: {future.exception()}')
+
+    if response.result.successful:
```
fujitatomoya, line 86: if `response.result.successful` is false, user can know that loading parameter operation has been failed?

ros2param/ros2param/api/__init__.py
```diff
@@ -66,6 +70,26 @@ def load_parameter_file(*, node, node_name, parameter_file, use_wildcard):
...
+        msg = 'Set parameters {} successful'.format(' '.join([i.name for i in parameters]))
```
fujitatomoya, line 87: this whole operation is atomic, and user knows parameters as input, so i am not sure if we need print all parameters again here. probably this command prints the result only?

ros2param/ros2param/api/__init__.py
```diff
@@ -96,6 +120,18 @@ def call_get_parameters(*, node, node_name, parameter_names):
     return response
 
 
+def call_set_parameters_atomically(*, node, node_name, parameters):
+    client = AsyncParameterClient(node, node_name)
+    client.wait_for_services(timeout_sec=5.0)
+    future = client.set_parameters_atomically(parameters)
+    rclpy.spin_until_future_complete(node, future)
+    response = future.result()
```
fujitatomoya, line 128: same here.

clalancette:
This was eventually done in #1204 , so I'm closing this one out.

PR #728, opened 2022-06-23
Set multiple parameters

ihasdapie:
Support setting multiple parameters at once using `ros2 param set`

See: https://github.com/ros2/ros2cli/pull/727#issuecomment-1164515330

This PR uses features from https://github.com/ros2/ros2cli/pull/716 (Diff: https://github.com/ros2/ros2cli/compare/brianc/parameter_client...brianc/multiple_param_set)

```
$ ros2 param set -h

usage: ros2 param set [-h] [--spin-time SPIN_TIME] [-s] [--no-daemon] [--include-hidden-nodes] node_name parameter_name value

Set parameter

positional arguments:
  node_name             Name of the ROS node
  parameters            List of parameter name and value pairs i.e. "int_param 1 str_param hello_world"

options:
  --spin-time SPIN_TIME
  --include-hidden-nodes
                        Consider hidden nodes as well
```
Example usage:

`ros2 param set /test_node int_param 12 str_param hello_world`

### DIFF
--- a/ros2param/ros2param/verb/set.py
+++ b/ros2param/ros2param/verb/set.py
@@ -83,13 +83,14 @@ class SetVerb(VerbExtension):
             results = response.results
 
             for i, result in enumerate(results):
+                msg = "Set parameter '%s' " % (parameters[i].name)
                 if result.successful:
-                    msg = f'Set parameter {parameters[i].name} successful'
+                    msg += 'successful'
                     if result.reason:
                         msg += ': ' + result.reason
                     print(msg)
                 else:
-                    msg = 'Set parameter failed'
+                    msg += 'failed'
                     if result.reason:
                         msg += ': ' + result.reason
                     print(msg, file=sys.stderr)
--- a/ros2param/test/test_verb_set.py
+++ b/ros2param/test/test_verb_set.py
@@ -167,7 +167,7 @@ class TestVerbSet(unittest.TestCase):
             assert param_load_command.wait_for_shutdown(timeout=TEST_TIMEOUT)
         assert param_load_command.exit_code == launch_testing.asserts.EXIT_OK
         assert launch_testing.tools.expect_output(
-            expected_lines=['Set parameter int_param successful'],
+            expected_lines=["Set parameter 'int_param' successful"],
             text=param_load_command.output,
             strict=True)
 
@@ -184,9 +184,9 @@ class TestVerbSet(unittest.TestCase):
         assert param_load_command.exit_code == launch_testing.asserts.EXIT_OK
         assert launch_testing.tools.expect_output(
             expected_lines=[
-                'Set parameter int_param successful',
-                'Set parameter str_param successful',
-                'Set parameter bool_param successful'],
+                "Set parameter 'int_param' successful",
+                "Set parameter 'str_param' successful",
+                "Set parameter 'bool_param' successful"],
             text=param_load_command.output,
             strict=True)
 
@@ -220,7 +220,3 @@ class TestVerbSet(unittest.TestCase):
             expected_lines=['Node not found'],
             text=param_load_command.output,
             strict=True)
-
-
-if __name__ == '__main__':
-    unittest.main()


ros2param/ros2param/verb/set.py
```diff
@@ -37,14 +37,29 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
...
         arg = parser.add_argument(
-            'parameter_name', help='Name of the parameter')
+            'parameters', nargs='*',
+            help='List of parameter name and value pairs i.e. "int_param 1 str_param hello_world"')
         arg.completer = ParameterNameCompleter()
+
         parser.add_argument(
-            'value', help='Value of the parameter')
+            '--include-hidden-nodes', action='store_true',
+            help='Consider hidden nodes as well')
```
clalancette, line 48: Can you move this code back above?  I think it is more logical to have all of the flags first, and then the "rest of" the arguments as that last item here.  This has no practical effect on how this operates, it is just a bit easier to read (and should also reduce the diff a bit).

ros2param/ros2param/verb/set.py
```diff
@@ -37,14 +37,29 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
...
+        parameters = []
+        if len(params) % 2:
+            raise RuntimeError('Must pass list of parameter name and value pairs')
+
+        params = [(params[i], params[i+1]) for i in range(0, len(params), 2)]
+        for param_str in params:
+            parameter = Parameter()
+            parameter.name = param_str[0]
+            parameter.value = get_parameter_value(string_value=param_str[1])
+            parameters.append(parameter)
```
clalancette, line 60: This seems like we are iterating over the list twice for no really good reason.  This can be simplified to:

```suggestion
        for i in range(0, len(params), 2):
            parameter = Parameter()
            parameter.name = params[i]
            parameter.value = get_parameter_value(string_value=params[i+1])
            parameters.append(parameter)
```

ros2param/ros2param/verb/set.py
```diff
@@ -37,14 +37,29 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
...
+            'parameters', nargs='*',
+            help='List of parameter name and value pairs i.e. "int_param 1 str_param hello_world"')
         arg.completer = ParameterNameCompleter()
+
         parser.add_argument(
-            'value', help='Value of the parameter')
+            '--include-hidden-nodes', action='store_true',
+            help='Consider hidden nodes as well')
+
+    def build_parameters(self, params):
```
clalancette, line 50: What happens here if the user runs the command like:

```
ros2 param set /node
```

(that is, they fail to provide *any* parameter names or values)?  Previously, that was an error (because the user didn't provide enough arguments), but now it will just try to set an empty set of parameters.  I think we should make that an error again.

ros2param/ros2param/verb/set.py
```diff
@@ -37,14 +37,29 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
...
+        parameters = []
+        if len(params) % 2:
+            raise RuntimeError('Must pass list of parameter name and value pairs')
```
clalancette, line 53: I feel like we shouldn't raise an exception here; previously, this would result in printing out a usage message and returning with shell result 2.  I think we should try to emulate that.

ros2param/ros2param/verb/set.py
```diff
@@ -56,23 +71,19 @@ def main(self, *, args):  # noqa: D102
...
+                if result.successful:
+                    msg = 'Set parameter successful'
+                    if result.reason:
+                        msg += ': ' + result.reason
+                    print(msg)
+                else:
+                    msg = 'Setting parameter failed'
+                    if result.reason:
+                        msg += ': ' + result.reason
+                    print(msg, file=sys.stderr)
```
clalancette, line 89: These two blocks are very similar, with the exception of `successful` vs `failed` in the output string, and also using `stderr` or `stdout` for printing.  It feels like they can be combined, though I'm not sure of that so just take this as a suggestion.

clalancette (changes requested):

ros2param/ros2param/verb/set.py
ihasdapie: I've combined them [here](https://github.com/ros2/ros2cli/blob/2e900095aa87f9581c8dda0b9fa7d619034e6756/ros2param/ros2param/verb/set.py#L92-L88) but I feel like it's a little unnecessairly terse.

```python3
           for i, result in enumerate(results):
                print('Set parameter ' + parameters[i].name + ' ' +
                      'successful' if result.successful else 'failed' +
                      str(result.reason) if result.reason else '',
                      file=sys.stderr if result.successful else sys.stdout)
```

clalancette: Yeah, agreed.  This is kind of hard to read.  I'd say let's go back to the previous version (sorry).

ros2param/ros2param/verb/set.py
ihasdapie: That's true -- printing out the whole backtrace and all is rather ugly. I've added an action to the new parameter argument list to maintain the same error behavior as before

https://github.com/ros2/ros2cli/blob/2e900095aa87f9581c8dda0b9fa7d619034e6756/ros2param/ros2param/verb/set.py#L32-L41

```
λ ros2 param set /add_two_ints_server
usage: ros2 param set [-h] [--spin-time SPIN_TIME] [-s]
                      [--no-daemon] [--include-hidden-nodes]
                      node_name [parameters ...]
ros2 param set: error: No parameter pairs specified
λ echo $?
2
λ ros2 param set /add_two_ints_server true
usage: ros2 param set [-h] [--spin-time SPIN_TIME] [-s]
                      [--no-daemon] [--include-hidden-nodes]
                      node_name [parameters ...]
ros2 param set: error: Must provide parameter name and value pairs
```

ihasdapie:
@clalancette Friendly ping -- I've made the requested changes.

ros2param/ros2param/verb/set.py
```diff
@@ -41,38 +54,45 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
...
+
+        return parameters
 
     def main(self, *, args):  # noqa: D102
         with NodeStrategy(args) as node:
             node_names = get_node_names(
                 node=node, include_hidden_nodes=args.include_hidden_nodes)
-
         node_name = get_absolute_node_name(args.node_name)
+
```
fujitatomoya, line 78: unnecessary changes here.

ros2param/ros2param/verb/set.py
```diff
@@ -28,6 +29,18 @@
 from ros2param.verb import VerbExtension
 
 
+class RequireParameterPairAction(argparse.Action):
+    """Argparse action to validate parameter argument pairs."""
+
+    def __call__(self, parser, args, values, option_string=None):
+        if len(values) == 0:
+            parser.error('No parameters specified')
+            SystemExit(2)
```
fujitatomoya, line 38: parser.error is able to handle system error 2?

ros2param/ros2param/verb/set.py
fujitatomoya: I think this will fail with system error 2 via parser.error in `RequireParameterPairAction`, @ihasdapie could you confirm?

fujitatomoya (commented):
I think test needs to be added in `ros2cli/ros2param/test` to set and get the parameters w/o error.

ros2param/ros2param/verb/set.py
ihasdapie: @fujitatomoya  You are correct, I forgot to delete that line after I realized parser.error will handle it. I'll fix your review comments & add tests soon

ihasdapie:
> I think test needs to be added in `ros2cli/ros2param/test` to set and get the parameters w/o error.

Done in https://github.com/ros2/ros2cli/pull/728/commits/5547f87a6efc1a3b7894a07bee021e343977cf74. There weren't any tests for the set verb before.

ros2param/ros2param/verb/set.py
ihasdapie: Yup!
https://github.com/ros2/ros2cli/blob/5547f87a6efc1a3b7894a07bee021e343977cf74/ros2param/ros2param/verb/set.py#L36-L37

clalancette:
I just rebased this and made some small fixes to it.  With this, it is looking good to me but I'd like another opinion before I merge it.  I'll go ahead and run CI, though.

clalancette (approved):

scpeters:
For backwards compatibility with `ros2 param set node param_name value`, I think the sequence of `ros2 param set node param_name1 value1 param_name2 value2` makes sense, but it gets tricky to follow visually. I think the launch syntax of `param:=value` is a little easier to read when there are multiple parameters being set, but it's not clear how to do that in backwards compatible way.

Just an observation, not blocking

clalancette:
> For backwards compatibility with `ros2 param set node param_name value`, I think the sequence of `ros2 param set node param_name1 value1 param_name2 value2` makes sense, but it gets tricky to follow visually. I think the launch syntax of `param:=value` is a little easier to read when there are multiple parameters being set, but it's not clear how to do that in backwards compatible way.

Yeah, I agree completely.  The thing is I think the backwards compatibility is important.

We could do a thing where the first argument is `name val`, and all subsequent ones are `name:=val`, but I think that inconsistency is worse.  We could also do a thing where we accept `name val` for a single argument with a warning, and then accept `name:=val` for arbitrary numbers of values without a warning.  That would be more work, but would be more like launch and `ros2 run` as well.

I'm honestly not sure.  @ros2/team thoughts?

nuclearsandwich:
> `@ros2/team` thoughts?




Perhaps it's possible to maintain the old behavior or at least emit a warning if a sequence of arguments is given and none of the arguments contain the := delimiter, thus indicating that it's a `param value param value` sequence rather than a `param:=value param:=value` sequence.

> The thing is I think the backwards compatibility is important.

Breaking compatibility to improve usability is worthwhile.
I'm not really sure that tick-tock cycles help for command-line tools so I would be in favor of:

* Target Iron for this change (i.e. don't backport it)
* Update the Iron release docs with this as a Notable Change since last release
* Make an extra announcement on Discourse about this when it lands in Rolling

Issue #702, opened 2022-03-18
ROS2 Daemon fails to respond


civerachb-cpr:
## Bug report

**Summary**

Running `ros2 topic list` once results in all subsequent `ros2 *` commands to fail with the following error:
```
Traceback (most recent call last):
  File "/opt/ros/foxy/bin/ros2", line 11, in <module>
    load_entry_point('ros2cli==0.9.11', 'console_scripts', 'ros2')()
  File "/opt/ros/foxy/lib/python3.8/site-packages/ros2cli/cli.py", line 67, in main
    rc = extension.main(parser=parser, args=args)
  File "/opt/ros/foxy/lib/python3.8/site-packages/ros2topic/command/topic.py", line 41, in main
    return extension.main(args=args)
  File "/opt/ros/foxy/lib/python3.8/site-packages/ros2topic/verb/list.py", line 38, in main
    with NodeStrategy(args) as node:
  File "/opt/ros/foxy/lib/python3.8/site-packages/ros2cli/node/strategy.py", line 52, in __enter__
    self._daemon_node.__enter__()
  File "/opt/ros/foxy/lib/python3.8/site-packages/ros2cli/node/daemon.py", line 116, in __enter__
    methods = self._proxy.system.listMethods()
  File "/usr/lib/python3.8/xmlrpc/client.py", line 1109, in __call__
    return self.__send(self.__name, args)
  File "/usr/lib/python3.8/xmlrpc/client.py", line 1450, in __request
    response = self.__transport.request(
  File "/usr/lib/python3.8/xmlrpc/client.py", line 1153, in request
    return self.single_request(host, handler, request_body, verbose)
  File "/usr/lib/python3.8/xmlrpc/client.py", line 1166, in single_request
    resp = http_conn.getresponse()
  File "/usr/lib/python3.8/http/client.py", line 1348, in getresponse
    response.begin()
  File "/usr/lib/python3.8/http/client.py", line 316, in begin
    version, status, reason = self._read_status()
  File "/usr/lib/python3.8/http/client.py", line 285, in _read_status
    raise RemoteDisconnected("Remote end closed connection without"
http.client.RemoteDisconnected: Remote end closed connection without response
```

Running `ros2 daemon stop` fails with the same error; the only way to stop the daemon is to run `killall _ros2_daemon`.

[...]

PR #729, opened 2022-06-23
XMLRPC server accepts request from all local IP addresses.

fujitatomoya:
address https://github.com/ros2/ros2cli/issues/702

### DIFF
--- a/ros2cli/ros2cli/xmlrpc/local_server.py
+++ b/ros2cli/ros2cli/xmlrpc/local_server.py
@@ -18,6 +18,19 @@ import struct
 from xmlrpc.server import SimpleXMLRPCRequestHandler  # noqa
 from xmlrpc.server import SimpleXMLRPCServer
 
+import netifaces
+
+
+def get_local_ipaddrs():
+    iplist = []
+    interfaces = netifaces.interfaces()
+    for interface in interfaces:
+        addrs = netifaces.ifaddresses(interface)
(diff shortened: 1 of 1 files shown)


fujitatomoya:
@clalancette @sloretz do you think this could work?

clalancette:
> @clalancette @sloretz do you think this could work?

In short, I think this will work.  We use `netifaces` elsewhere in ros2cli, so it is not a new dependency, and presumably it works on Windows (though I have not yet tested that).

I also tested this out with the original problematic scenario in #702 , and it seems to fix that problem for me.  All we need is confirmation that it works on Windows and CI, then I think we'll be good to go with this.

clalancette:
I tested this briefly on Windows, and it seems to do what I expect.  So with that in mind, I'm going to approve.  We just need to run CI on this and we'll be good to go.  Thanks @fujitatomoya !

clalancette (approved):

fujitatomoya (commented):
@clalancette thanks for the verification, i will run CI.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16920/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11486/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17319/)

clalancette:
@Mergifyio backport humble galactic foxy

PR #732, opened 2022-07-05
[Service Introspection] Support echo verb for ros2 service cli

deepanshubansal01:
Related to the proposed service introspection REP (https://github.com/ros-infrastructure/rep/pull/360) 

This PR implements the cli for echoing service events. 
`ros2 service echo /name_of_service`

Depends on https://github.com/ros2/rosidl_runtime_py/pull/17 for deserializing bytes type data and echoing contents to the user.

Connects to: https://github.com/ros2/ros2/issues/1285

Signed-off-by: deepanshu <deepanshubansal01@gmail.com>

### DIFF
--- a/ros2service/ros2service/api/__init__.py
+++ b/ros2service/ros2service/api/__init__.py
@@ -33,6 +33,9 @@ def get_service_names(*, node, include_hidden_services=False):
         node=node, include_hidden_services=include_hidden_services)
     return [n for (n, t) in service_names_and_types]
 
+def get_service_type(*, node, service, include_hidden_services=False):
+    pass
+
 
 def service_type_completer(**kwargs):
     """Callable returning a list of service types."""
--- a/ros2service/ros2service/api/__init__.py
+++ b/ros2service/ros2service/api/__init__.py
@@ -12,11 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from time import sleep
 from rclpy.topic_or_service_is_hidden import topic_or_service_is_hidden
 from ros2cli.node.strategy import NodeStrategy
 from rosidl_runtime_py import get_service_interfaces
 from rosidl_runtime_py import message_to_yaml
 from rosidl_runtime_py.utilities import get_service
+
+import rclpy
 
 
 def get_service_names_and_types(*, node, include_hidden_services=False):
@@ -33,8 +36,47 @@ def get_service_names(*, node, include_hidden_services=False):
         node=node, include_hidden_services=include_hidden_services)
     return [n for (n, t) in service_names_and_types]
 
-def get_service_type(*, node, service, include_hidden_services=False):
-    pass
+
+def get_service_class(node, service, blocking=False, include_hidden_services=False):
+    srv_class = _get_service_class(node, service, include_hidden_services)
+    if srv_class:
+        return srv_class
+    elif blocking:
+        print('WARNING: service [%s] does not appear to be started yet' % service)
+        while rclpy.ok():
+            srv_class = _get_service_class(node, service, include_hidden_services)
+            if srv_class:
+                return srv_class
+            else:
+                sleep(0.1)
+    else:
+        print('WARNING: service [%s] does not appear to be started yet' % service)
+    return None
+
+
+def _get_service_class(node, service, include_hidden_services):
+    service_names_and_types = get_service_names_and_types(
+        node=node,
+        include_hidden_services=include_hidden_services)
+
+    service_type = None
+    for (service_name, service_types) in service_names_and_types:
+        if service == service_name:
+            if len(service_types) > 1:
+                raise RuntimeError(
+                    "Cannot echo service '%s', as it contains more than one type: [%s]" %
+                    (service, ', '.join(service_types))
+                )
+        service_type = service_types[0]
+        break
+
+    if service_type is None:
+        return None
+
+    try:
+        return get_service(service_type)
+    except (AttributeError, ModuleNotFoundError, ValueError):
+        raise RuntimeError("The service type '%s' is invalid" % service_type)
 
 
 def service_type_completer(**kwargs):
--- a/ros2service/ros2service/verb/echo.py
+++ b/ros2service/ros2service/verb/echo.py
@@ -1,0 +1,142 @@
+# Copyright 2022 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import importlib
+from typing import Optional
+from typing import TypeVar
+
+import rclpy
+from rclpy.node import Node
+from rclpy.qos import QoSDurabilityPolicy
+from rclpy.qos import QoSPresetProfiles
+from rclpy.qos import QoSProfile
+from rclpy.serialization import deserialize_message
+
+from rcl_interfaces.msg import ServiceEventType
+
+from ros2topic.api import unsigned_int
+
+from ros2service.api import ServiceNameCompleter
+from ros2service.api import ServiceTypeCompleter
+from ros2service.verb import VerbExtension
+from rosidl_runtime_py.utilities import get_message
+from ros2cli.node.strategy import NodeStrategy
+
+from rosidl_runtime_py import message_to_yaml
+import yaml
+
+DEFAULT_TRUNCATE_LENGTH = 128
+MsgType = TypeVar('MsgType')
+
+
+class EchoVerb(VerbExtension):
+    """Echo a service."""
+
+    def __init__(self):
+        super().__init__()
+        self.srv_module = None
+        self.topic_name = None
+        self.no_str = None
+        self.no_arr = None
+        self.truncate_length = None
+        self.flow_style = None
+        self.hidden_topic_suffix = "/_service_event"
+        self.message_type = get_message("rcl_interfaces/msg/ServiceEvent")
+        self.qos_profile = QoSPresetProfiles.get_from_short_key("services_default")
+
+    def add_arguments(self, parser, cli_name):
+        arg = parser.add_argument(
+            'service_name',
+            help="Name of the ROS service to echo from (e.g. '/add_two_ints')")
+        arg.completer = ServiceNameCompleter(
+            include_hidden_services_key='include_hidden_services')
+        arg = parser.add_argument(
+            'service_type', nargs='?',
+            help="Type of the ROS service (e.g. 'std_srvs/srv/Empty')")
+        arg.completer = ServiceTypeCompleter(
+            service_name_key='service_name')
+        parser.add_argument(
+            '--full-length', '-f', action='store_true',
+            help='Output all elements for arrays, bytes, and string with a '
+                 "length > '--truncate-length', by default they are truncated "
+                 "after '--truncate-length' elements with '...''")
+        parser.add_argument(
+            '--truncate-length', '-l', type=unsigned_int, default=DEFAULT_TRUNCATE_LENGTH,
+            help='The length to truncate arrays, bytes, and string to '
+                 '(default: %d)' % DEFAULT_TRUNCATE_LENGTH)
+        parser.add_argument(
+            '--no-arr', action='store_true', help="Don't print array fields of messages")
+        parser.add_argument(
+            '--no-str', action='store_true', help="Don't print string fields of messages")
+        parser.add_argument(
+            '--include-message-info', '-i', action='store_true',
+            help='Shows the associated message info.')
+
+    def main(self, *, args):
+        self.topic_name = args.service_name + self.hidden_topic_suffix
+        self.truncate_length = args.truncate_length if not args.full_length else None
+        self.no_arr = args.no_arr
+        self.no_str = args.no_str
+
+        try:
+            parts = args.service_type.split('/')
+            if len(parts) == 2:
+                parts = [parts[0], 'srv', parts[1]]
+            package_name = parts[0]
+            print(parts[:-1])
+            module = importlib.import_module('.'.join(parts[:-1]))
+            srv_name = parts[-1]
+            self.srv_module = getattr(module, srv_name)
+        except (AttributeError, ModuleNotFoundError, ValueError):
+            raise RuntimeError('The passed service type is invalid')
+        try:
+            var = self.srv_module.Request
+            var = self.srv_module.Response
+        except AttributeError:
+            raise RuntimeError('The passed type is not a service')
+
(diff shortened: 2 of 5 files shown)


ihasdapie:
I gave this a run and it's great to see a working prototype going. I would like to see this tool behave in a way consistent to `ros2 topic echo`, i.e. wait until the service is avaliable instead of exiting if unavaliable. This way we can keep a terminal with `ros2 service echo` open instead of having to scramble to run it once the service starts.

deepanshubansal01:
> I gave this a run and it's great to see a working prototype going. I would like to see this tool behave in a way consistent to `ros2 topic echo`, i.e. wait until the service is avaliable instead of exiting if unavaliable. This way we can keep a terminal with `ros2 service echo` open instead of having to scramble to run it once the service starts.

I see, makes sense. I guess right now the `ros2 service echo` behaves similar to the way` ros2 topic echo` works. The` ros2 topic echo` also exits if the topic name does not exist. For instance if you do` ros2 topic echo /test` it will exit if `/test `topic is not their. The reason is it needs the topic to be present to determine the topic type. However, you can provide the topic type along with topic name, making the echo to not exit even if the topic is not present.

This being said we can also wait for the service to be present, I just made it similar to ros2 topic for now. In order to make it wait for service we can set `blocking` to True here https://github.com/ros2/ros2cli/blob/02c9e93e264bbadaffe8a61b6ce8167e9d751f34/ros2service/ros2service/api/__init__.py#L40

ros2service/ros2service/verb/echo.py
```diff
@@ -0,0 +1,236 @@
...
+
+def represent_ordereddict(dumper, data):
+    items = []
+    for k, v in data.items():
+        items.append((dumper.represent_data(k), dumper.represent_data(v)))
+    return yaml.nodes.MappingNode(u'tag:yaml.org,2002:map', items)
+
+
+class EchoVerb(VerbExtension):
+    """Echo a service."""
```
adityapande-1995, line 49: Could you expand this docstring to provide a short explanation and an example ? It would be nice if we could write a simple test / self contained example for this feature.

ros2service/setup.py
```diff
@@ -44,6 +44,7 @@
             'find = ros2service.verb.find:FindVerb',
             'list = ros2service.verb.list:ListVerb',
             'type = ros2service.verb.type:TypeVerb',
+            'echo = ros2service.verb.echo:EchoVerb',
```
jacobperron, line 47: nitpick: alphabetize this list.

ros2service/ros2service/api/__init__.py
```diff
@@ -34,6 +37,48 @@ def get_service_names(*, node, include_hidden_services=False):
     return [n for (n, t) in service_names_and_types]
 
 
+def get_service_class(node, service, blocking=False, include_hidden_services=False):
```
jacobperron, line 40: Please document this function and its arguments.

ros2service/ros2service/api/__init__.py
```diff
@@ -34,6 +37,48 @@ def get_service_names(*, node, include_hidden_services=False):
...
+        node=node,
+        include_hidden_services=include_hidden_services)
+
+    service_type = None
+    for (service_name, service_types) in service_names_and_types:
+        if service == service_name:
+            if len(service_types) > 1:
+                raise RuntimeError(
+                    "Cannot echo service '%s', as it contains more than one type: [%s]" %
+                    (service, ', '.join(service_types))
```
jacobperron, line 68: IMO, f-strings are more readable
```suggestion
                    f"Cannot echo service '{service}', as it contains more than one type: [{', '.join(service_types)}]")
```

ros2service/ros2service/api/__init__.py
```diff
@@ -34,6 +37,48 @@ def get_service_names(*, node, include_hidden_services=False):
...
+        service_type = service_types[0]
+        break
+
+    if service_type is None:
+        return None
+
+    try:
+        return get_service(service_type)
+    except (AttributeError, ModuleNotFoundError, ValueError):
+        raise RuntimeError("The service type '%s' is invalid" % service_type)
```
jacobperron, line 79:
```suggestion
        raise RuntimeError(f"The service type '{service_type}' is invalid")
```

ros2service/ros2service/api/__init__.py
```diff
@@ -34,6 +37,48 @@ def get_service_names(*, node, include_hidden_services=False):
     return [n for (n, t) in service_names_and_types]
 
 
+def get_service_class(node, service, blocking=False, include_hidden_services=False):
+    srv_class = _get_service_class(node, service, include_hidden_services)
+    if srv_class:
```
jacobperron, line 42:
```suggestion
    if srv_class is not None:
```

ros2service/ros2service/api/__init__.py
```diff
@@ -34,6 +37,48 @@ def get_service_names(*, node, include_hidden_services=False):
...
+        return srv_class
+    elif blocking:
+        print('WARNING: service [%s] does not appear to be started yet' % service)
```
jacobperron, line 45:
```suggestion
        print(f'WARNING: service [{service}] does not appear to be started yet')
```

ros2service/ros2service/api/__init__.py
```diff
@@ -34,6 +37,48 @@ def get_service_names(*, node, include_hidden_services=False):
...
+        while rclpy.ok():
+            srv_class = _get_service_class(node, service, include_hidden_services)
+            if srv_class:
```
jacobperron, line 48:
```suggestion
            if srv_class is not None:
```

ros2service/ros2service/api/__init__.py
```diff
@@ -34,6 +37,48 @@ def get_service_names(*, node, include_hidden_services=False):
...
+                return srv_class
+            else:
+                sleep(0.1)
```
jacobperron, line 51:
```suggestion
            sleep(0.1)
```

ros2service/ros2service/api/__init__.py
```diff
@@ -34,6 +37,48 @@ def get_service_names(*, node, include_hidden_services=False):
...
+    else:
+        print('WARNING: service [%s] does not appear to be started yet' % service)
```
jacobperron, line 53:
```suggestion
        print(f'WARNING: service [{service}] does not appear to be started yet')
```

ros2service/ros2service/verb/echo.py
jacobperron: Yes, we should add a test to [test_cli.py](https://github.com/ros2/ros2cli/blob/rolling/ros2service/test/test_cli.py).

ros2service/ros2service/verb/echo.py
```diff
@@ -0,0 +1,236 @@
...
+        self.csv = None
+        self.flow_style = None
+        self.client = None
+        self.server = None
+        self.truncate_length = None
+        self.include_message_info = None
+        self.srv_module = None
+        self.event_enum = None
+        self.event_msg_type = get_message("rcl_interfaces/msg/ServiceEvent")
+        self.qos_profile = QoSPresetProfiles.get_from_short_key("services_default")
```
jacobperron, line 64: style:
```suggestion
        self.event_msg_type = get_message('rcl_interfaces/msg/ServiceEvent')
        self.qos_profile = QoSPresetProfiles.get_from_short_key('services_default')
```

ros2service/ros2service/verb/echo.py
```diff
@@ -0,0 +1,236 @@
...
+        self.__yaml_representer_registered = False
+
+    def add_arguments(self, parser, cli_name):
+        arg = parser.add_argument(
+            'service_name',
+            help="Name of the ROS service to echo from (e.g. '/add_two_ints')")
```
jacobperron, line 70:
```suggestion
            help="Name of the ROS service to echo (e.g. '/add_two_ints')")
```

ros2service/ros2service/verb/echo.py
```diff
@@ -0,0 +1,236 @@
...
+        arg.completer = ServiceNameCompleter(
+            include_hidden_services_key='include_hidden_services')
+        arg = parser.add_argument(
+            'service_type', nargs='?',
+            help="Type of the ROS service (e.g. 'std_srvs/srv/Empty')")
```
jacobperron, line 75: Since we're using `/add_two_ints` as the example service name, I think we should use `example_interfaces/srv/AddTwoInts` as the example here, to be consistent.

ros2service/ros2service/verb/echo.py
```diff
@@ -0,0 +1,236 @@
...
+        arg.completer = ServiceTypeCompleter(
+            service_name_key='service_name')
```
jacobperron, line 77: Looks like this can fit on one line
```suggestion
        arg.completer = ServiceTypeCompleter(service_name_key='service_name')
```

ros2service/ros2service/verb/echo.py
```diff
@@ -0,0 +1,236 @@
...
+        parser.add_argument(
+            '--full-length', '-f', action='store_true',
+            help='Output all elements for arrays, bytes, and string with a '
+                 "length > '--truncate-length', by default they are truncated "
+                 "after '--truncate-length' elements with '...''")
```
jacobperron, line 82:
```suggestion
                 "after '--truncate-length' elements with '...'")
```

ros2service/ros2service/verb/echo.py
```diff
@@ -0,0 +1,236 @@
...
+        parser.add_argument(
+            '--truncate-length', '-l', type=unsigned_int, default=DEFAULT_TRUNCATE_LENGTH,
+            help='The length to truncate arrays, bytes, and string to '
+                 '(default: %d)' % DEFAULT_TRUNCATE_LENGTH)
```
jacobperron, line 86:
```suggestion
                 f'(default: {DEFAULT_TRUNCATE_LENGTH})')
```

ros2service/ros2service/verb/echo.py
```diff
@@ -0,0 +1,236 @@
...
+                'If --include-message-info is also passed, the following fields are prepended: '
+                'source_timestamp, received_timestamp, publication_sequence_number,'
+                ' reception_sequence_number.'
+            )
+        )
+        parser.add_argument(
+            '--include-message-info', '-i', action='store_true',
+            help='Shows the associated message info.')
+        parser.add_argument(
+            '--client', action='store_true', help="Echo only request sent or response received by service client")
```
jacobperron, line 105:
```suggestion
            '--client', action='store_true', help="Echo only request sent or response received messages from service clients")
```

ros2service/ros2service/verb/echo.py
```diff
@@ -0,0 +1,236 @@
...
+        parser.add_argument(
+            '--server', action='store_true', help="Echo only request received or response sent by service server")
```
jacobperron, line 107:
```suggestion
            '--server', action='store_true', help="Echo only request received or response sent messages from service servers")
```

ros2service/ros2service/verb/echo.py
jacobperron, line 105: Renaming this to `--client-only` seems more intuitive to me, what do others think?
The same applies to the server option.

ros2service/ros2service/verb/echo.py
```diff
@@ -0,0 +1,236 @@
...
+
+        if args.service_type is None:
+            with NodeStrategy(args) as node:
+                self.srv_module = get_service_class(
+                    node, args.service_name, blocking=False, include_hidden_services=True)
+        else:
+            try:
+                self.srv_module = get_service(args.service_type)
+            except (AttributeError, ModuleNotFoundError, ValueError):
+                raise RuntimeError("The service type '%s' is invalid" % args.service_type)
```
jacobperron, line 126:
```suggestion
                raise RuntimeError(f"The service type '{args.service_type}' is invalid")
```

ros2service/ros2service/verb/echo.py
```diff
@@ -0,0 +1,236 @@
...
+                event_topic_name,
+                self.event_msg_type
+            )
+
+    def subscribe_and_spin(
+            self,
+            node: Node,
+            event_topic_name: str,
+            event_msg_type: MsgType,
+    ) -> Optional[str]:
```
jacobperron, line 146: It doesn't look like this function is returning anything.
```suggestion
    ) -> None:
```

ros2service/ros2service/verb/echo.py
```diff
@@ -0,0 +1,236 @@
...
+            event_msg_type,
+            event_topic_name,
+            self._subscriber_callback,
+            self.qos_profile)
+
+        rclpy.spin(node)
+
+    def _subscriber_callback(self, msg, info):
+        deserialize_msg_type = None
+        self.event_enum = msg.info.event_type
```
jacobperron, line 159: Seems a bit convoluted to add this data to the class object. Maybe just pass it directly to `format_output` as an argument

ros2service/ros2service/verb/echo.py
```diff
@@ -0,0 +1,236 @@
...
+
+        if self.event_enum is ServiceEventType.REQUEST_RECEIVED or \
+                self.event_enum is ServiceEventType.REQUEST_SENT:
+            deserialize_msg_type = self.srv_module.Request
+        elif self.event_enum is ServiceEventType.RESPONSE_RECEIVED or \
+                self.event_enum is ServiceEventType.RESPONSE_SENT:
+            deserialize_msg_type = self.srv_module.Response
+        else:  # TODO remove this once event enum is correct
+            print("Returning invalid service event type")
+            return
```
jacobperron, line 180: Should this case be removed? If we get to this case it sounds like it implies an error in the service introspection implementation. I think we should probably raise an error here, like
```py
raise ValueError(f'received unexpected service event type {event_enum}')
```

ros2service/ros2service/verb/echo.py
```diff
@@ -0,0 +1,236 @@
...
+
+        # csv
+        if self.csv:
+            to_print = message_to_csv(
+                msg,
+                truncate_length=self.truncate_length,
+                no_arr=self.no_arr,
+                no_str=self.no_str,
+                deserialize_msg_type=deserialize_msg_type
```
jacobperron, line 189: Following our discussion in https://github.com/ros2/rosidl_runtime_py/pull/17, I think we can continue to call `message_to_csv` here, deserialized the submessage here, and append the submessage to the output before printing. I think this works out because the request/response data is at the very end of the service event message.

Hopefully it's easy to do the same for the YAML format below.

ihasdapie:
@deepanshubansal01 I've pushed a commit which tidies up the code a bit and removes the serialization bit. I've also tided up the output.

Also, sorry for hijacking your branch. On a side note, if possible I'd like to close this PR and open up a new one so that we can have all of the PRs off a `service_introspection` branch which would make CI a little easier later.

deepanshubansal01:
> @deepanshubansal01 I've pushed a commit which tidies up the code a bit and removes the serialization bit. I've also tided up the output.
> 
> Also, sorry for hijacking your branch. On a side note, if possible I'd like to close this PR and open up a new one so that we can have all of the PRs off a `service_introspection` branch which would make CI a little easier later.

Sounds good to me. You can open a new PR, this PR also contains some changes related to unit tests which we can ignore in the new PR since they were in progress. Unit tests can be added in the new PR though.

ihasdapie:
I opened https://github.com/ros2/ros2cli/pull/745. Closing in lieu of that PR.

PR #749, opened 2022-08-19
Expand auto to the current time when passed to a Header field

### LINKED ISSUE #700: ros2 topic pub auto header
#### Feature description
ROS1 allowed automatic filling of header and timestamp using keywords 'now' and 'auto'. It would be great to have that functionality in ROS2

ex.
```
ros2 topic pub my_topic geometry_msgs/TwistStamped '{header: auto}'
```

#### Expected behavior
```
geometry_msgs.msg.TwistStamped(header=std_msgs.msg.Header(stamp=builtin_interfaces.msg.Time(sec=1646834300, nanosec=280516750), frame_id=''), twist=geometry_msgs.msg.Twist(linear=geometry_msgs.msg.Vector3(x=0.0, y=0.0, z=0.0), angular=geometry_msgs.msg.Vector3(x=0.0, y=0.0, z=0.0)))
```
#### Actual behavior
`Failed to populate field: Value 'auto' is expected to be a dictionary but is a str`

esteve:
This PR enables expansion of `auto` to a generated `Header` whose `stamp` field is set to the current time. This also expands `now` to the current time if the field is of type `builtin_interfaces.msg.Time`

These changes match the same behavior that `rostopic` has:

http://wiki.ros.org/ROS/YAMLCommandLine#Headers.2Ftimestamps

Closes #700

### DIFF
--- a/ros2topic/ros2topic/verb/pub.py
+++ b/ros2topic/ros2topic/verb/pub.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 import array
-from collections import defaultdict
 from functools import partial
 import time
 from typing import Any
@@ -62,7 +61,10 @@ def set_message_fields_expanded(node: Node, msg: Any, values: Dict[str, str]) ->
     :raises TypeError: If a message value does not match its field type.
     """
     timestamp_fields = []
-    def set_message_fields_expanded_internal(node: Node, msg: Any, values: Dict[str, str], timestamp_fields: List[Any]) -> List[Any]:
+
+    def set_message_fields_expanded_internal(
+            node: Node, msg: Any, values: Dict[str, str],
+            timestamp_fields: List[Any]) -> List[Any]:
         try:
             items = values.items()
         except AttributeError:
@@ -79,13 +81,14 @@ def set_message_fields_expanded(node: Node, msg: Any, values: Dict[str, str]) ->
                 value = numpy.array(field_value, dtype=field.dtype)
             elif type(field_value) is field_type:
                 value = field_value
-            # We can't import these types directly, so we use the qualified class name to distinguish
-            # them from other fields
+            # We can't import these types directly, so we use the qualified class name to
+            # distinguish them from other fields
             elif qualified_class_name == 'std_msgs.msg._header.Header' and field_value == 'auto':
                 stamp_now = node.get_clock().now().to_msg()
                 value = field_type(stamp=stamp_now, frame_id='')
                 timestamp_fields.append(partial(setattr, value, 'stamp'))
-            elif qualified_class_name == 'builtin_interfaces.msg._time.Time' and field_value == 'now':
+            elif qualified_class_name == 'builtin_interfaces.msg._time.Time' and \
+                    field_value == 'now':
                 value = node.get_clock().now().to_msg()
                 timestamp_fields.append(partial(setattr, msg, field_name))
             else:
@@ -93,7 +96,8 @@ def set_message_fields_expanded(node: Node, msg: Any, values: Dict[str, str]) ->
                     value = field_type(field_value)
                 except TypeError:
                     value = field_type()
-                    set_message_fields_expanded_internal(node, value, field_value, timestamp_fields)
+                    set_message_fields_expanded_internal(
+                        node, value, field_value, timestamp_fields)
             rosidl_type = get_message_slot_types(msg)[field_name]
             # Check if field is an array of ROS messages
             if isinstance(rosidl_type, AbstractNestedType):
@@ -101,7 +105,8 @@ def set_message_fields_expanded(node: Node, msg: Any, values: Dict[str, str]) ->
                     field_elem_type = import_message_from_namespaced_type(rosidl_type.value_type)
                     for n in range(len(value)):
                         submsg = field_elem_type()
-                        set_message_fields_expanded_internal(node, submsg, value[n], timestamp_fields)
+                        set_message_fields_expanded_internal(
+                            node, submsg, value[n], timestamp_fields)
                         value[n] = submsg
             setattr(msg, field_name, value)
     set_message_fields_expanded_internal(node, msg, values, timestamp_fields)
--- a/ros2topic/ros2topic/verb/pub.py
+++ b/ros2topic/ros2topic/verb/pub.py
@@ -46,10 +46,15 @@ def set_message_fields_expanded(node: Node, msg: Any, values: Dict[str, str]) ->
     """
     Set the fields of a ROS message.
 
+    This function is copied from :py:func:`rosidl_runtime_py.set_message_fields`, but with a
+    'Node' parameter, required for obtaining the current time.
     :param msg: The ROS message to populate.
     :param values: The values to set in the ROS message. The keys of the dictionary represent
-        fields of the message. If 'auto' is passed as a value to a 'std_msgs.msg.Header' field,
-        an empty Header will be instantiated with its 'stamp' field set to the current time.
+        fields of the message.
+        If 'auto' is passed as a value to a 'std_msgs.msg.Header' field, an empty Header will be
+        instantiated with its 'stamp' field set to the current time.
+        If 'now' is passed as a value to a 'builtin_interfaces.msg.Time' field, its value will be
+        expanded to the current time.
     :raises AttributeError: If the message does not have a field provided in the input dictionary.
     :raises TypeError: If a message value does not match its field type.
     """
@@ -69,9 +74,13 @@ def set_message_fields_expanded(node: Node, msg: Any, values: Dict[str, str]) ->
             value = numpy.array(field_value, dtype=field.dtype)
         elif type(field_value) is field_type:
             value = field_value
+        # We can't import these types directly, so we use the qualified class name to distinguish
+        # them from other fields
         elif qualified_class_name == 'std_msgs.msg._header.Header' and field_value == 'auto':
             stamp_now = node.get_clock().now().to_msg()
             value = field_type(stamp=stamp_now, frame_id='')
+        elif qualified_class_name == 'builtin_interfaces.msg._time.Time' and field_value == 'now':
+            value = node.get_clock().now().to_msg()
         else:
             try:
                 value = field_type(field_value)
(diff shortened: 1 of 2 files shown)


fujitatomoya (commented):
@esteve do we want this as option via command line like `ros2 topic pub --with-current-time`? if message does not have the field, probably fallback to ignore that option 1st. this way we can avoid the overhead for current users? (it seems that populating field every single time via timer callback would not be cost effective?) what do you think?

btw, test would be ideal.

fujitatomoya:
> do we want this as option via command line like ros2 topic pub --with-current-time? if message does not have the field, probably fallback to ignore that option 1st. this way we can avoid the overhead for current users? (it seems that populating field every single time via timer callback would not be cost effective?) what do you think?

@clalancette @ivanpauno could you share your thought?

esteve:
@fujitatomoya Thanks for the feedback.

> @esteve do we want this as option via command line like `ros2 topic pub --with-current-time`? if message does not have the field, probably fallback to ignore that option 1st. this way we can avoid the overhead for current users? (it seems that populating field every single time via timer callback would not be cost effective?) what do you think?

I thought about that, but it always felt to me that `--substitute-keywords` in ROS1's `rostopic` (which would be the equivalent of `--with-current-time`) was a workaround for https://github.com/ros/ros_comm/issues/55

In any case, I agree with you that for large messages updating the message very time might not be the best. I'll look into optimizing the code so that only timestamps are updated, and the rest of the message is left intact. However, I do think that the default behavior for `ros2 topic pub` should be to always update timestamps with current time if `now` is passed as a value, with no extra flags.

fujitatomoya:
@esteve thanks for sharing information.

basically, as user interface, i think

- `ros2 topic pub` option `--with-current-time` is easy to use, but this overrides all fields `stamp` and `builtin_interfaces.msg.Time`. If user wants to insert the current time in specific field of the message, it cannot support.
- `auto` and `now` statements in message field can allow user to manage the message field, more flexible.

i believe `auto` and `now` statements are okay.

> large messages updating the message very time might not be the best. I'll look into optimizing the code so that only timestamps are updated, and the rest of the message is left intact.

as implementation, i do not have any better way to do this at this moment, sorry.

personally, if it requires performance, i will built the binary instead of using script or shell. i think this can be acceptable, but just thought about the user who is not interested in this feature would want to avoid this overhead.

fujitatomoya (approved):
lgtm, probably it would be nice to add test.

esteve:
@fujitatomoya thanks for approving this PR, but I'll work today on optimizing it and adding a test

esteve:
@fujitatomoya I've added a bunch of tests, the rest of the logic in the `set_message_fields_expanded` function is already tested in `rosidl_runtime_py.set_message_fields`

I've optimized the logic for expanding `now` and `auto` so that only the appropriate `builtin_interfaces.msg.Time` fields are updated, and the the rest of the message is left intact, thus avoiding any extra overhead when continuously publishing messages.

Additionally, with this PR, you can pass `now` to any `builtin_interfaces.msg.Time` field and it'll be expanded to the current time, not only the stamp in the header.

ros2topic/ros2topic/verb/pub.py
```diff
@@ -27,13 +34,85 @@
...
+from rosidl_parser.definition import NamespacedType
+from rosidl_runtime_py.convert import get_message_slot_types
+from rosidl_runtime_py.import_message import import_message_from_namespaced_type
 from rosidl_runtime_py.utilities import get_message
 import yaml
 
 MsgType = TypeVar('MsgType')
 
 
+def set_message_fields_expanded(node: Node, msg: Any, values: Dict[str, str]) -> List[Any]:
```
ivanpauno, line 47: can you move most of the new logic to https://github.com/ros2/ros2cli/blob/0a33a5bd7fed869530b9dcf2e8771ff553d019dc/ros2topic/ros2topic/api/__init__.py?

ivanpauno: or maybe this could be added to `rosidl_runtime_py`, so it can be reused in other tools

esteve:
@fujitatomoya I need to make some more changes, I just realized that `set_message_fields_expanded` can be simplified a bit more, it doesn't really need a `Node` argument if the timestamps are updated elsewhere.

ros2topic/ros2topic/verb/pub.py
fujitatomoya: makes sense!

ros2topic/ros2topic/verb/pub.py
```diff
@@ -27,13 +34,85 @@
...
+            elif field_type is numpy.ndarray:
+                value = numpy.array(field_value, dtype=field.dtype)
+            elif type(field_value) is field_type:
+                value = field_value
+            # We can't import these types directly, so we use the qualified class name to
+            # distinguish them from other fields
+            elif qualified_class_name == 'std_msgs.msg._header.Header' and field_value == 'auto':
+                stamp_now = node.get_clock().now().to_msg()
+                value = field_type(stamp=stamp_now, frame_id='')
+                timestamp_fields.append(partial(setattr, value, 'stamp'))
```
fujitatomoya, line 89: can we add `continue` here, since `partial`ed function object will be called and overwrite this feild?

```suggestion
                timestamp_fields.append(partial(setattr, value, 'stamp'))
                continue
```

ros2topic/ros2topic/verb/pub.py
```diff
@@ -27,13 +34,85 @@
...
+            elif qualified_class_name == 'builtin_interfaces.msg._time.Time' and \
+                    field_value == 'now':
+                value = node.get_clock().now().to_msg()
+                timestamp_fields.append(partial(setattr, msg, field_name))
```
fujitatomoya, line 93: same here?

ros2topic/ros2topic/verb/pub.py
esteve: Done in bdd896bdc89c46b86b397ff3d8d3730479a7e39d

ros2topic/ros2topic/verb/pub.py
esteve: Done in bdd896bdc89c46b86b397ff3d8d3730479a7e39d

ros2topic/ros2topic/verb/pub.py
esteve: @ivanpauno I prefer to move it to `rosidl_runtime_py` so that there's no duplicate logic. Moving to `ros2topic` may not be that useful since no other verbs use the original `set_message_fields` function

esteve:
@fujitatomoya @ivanpauno can you have another look? I'm ok with merging the logic in `set_message_fields_expanded` with `rosidl_runtime_py.set_message_fields`, I'll submit a PR to `rosidl_runtime_py`, but I want to make sure that the rest of the code here is ok.

esteve:
@fujitatomoya @ivanpauno here's the PR for the changes in `rosidl_runtime_py` https://github.com/ros2/rosidl_runtime_py/pull/19

ros2topic/ros2topic/verb/pub.py
```diff
@@ -44,26 +44,32 @@
...
-        If 'auto' is passed as a value to a 'std_msgs.msg.Header' field, an empty Header will be
-        instantiated with its 'stamp' field set to the current time.
-        If 'now' is passed as a value to a 'builtin_interfaces.msg.Time' field, its value will be
-        expanded to the current time.
+    :param expand_header_auto: If enabled and 'auto' is passed as a value to a
+        'std_msgs.msg.Header' field, an empty Header will be instantiated and a setter function
+        will be returned so that its 'stamp' field can be set to the current time.
+    :param expand_time_now: If enabled and 'now' is passed as a value to a
+        'builtin_interfaces.msg.Time' field, a setter function will be returned so that
+        its value can be set to the current time.
```
ivanpauno, line 61: this doesn't seem to be accurate for the function defined here (though accurate for the `rosidl_runtime_py` PR).

ivanpauno (approved):
LGTM!
I only have one minor comment.

I think it's okay to merge this before the `rosidl_runtime_py` PR if that one gets delayed, the code here can be easily updated when that one is merged.

fujitatomoya (approved):
lgtm

fujitatomoya:
@esteve is this still draft?

ros2topic/ros2topic/verb/pub.py
esteve: I'm afraid I'm not following, both docstrings are the same. Can you elaborate what should be updated here? Thanks.

esteve:
@fujitatomoya thanks for the reminder. This is now ready, but I asked @ivanpauno to expand on his feedback.

ros2topic/ros2topic/verb/pub.py
ivanpauno: Sorry, I think I got confused when reading the diff.
The docs are fine

ivanpauno:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17229/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11772/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17691/)

ivanpauno:
@esteve could you check the `flake8` failures?

fujitatomoya:
CI retry:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17239/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11779/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17699/)

esteve:
@fujitatomoya @ivanpauno the flake8 test should now pass

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17241/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11782/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17701/)

esteve:
@ivanpauno thanks for triggering CI, I see that all the tests pass now. Any other improvements that I should address? /cc @fujitatomoya

ivanpauno (approved):

ivanpauno:
Thanks for the contribution @esteve !!!

> @ivanpauno thanks for triggering CI, I see that all the tests pass now. Any other improvements that I should address? /cc @fujitatomoya

Not right now, but when https://github.com/ros2/rosidl_runtime_py/pull/19 is ready it would be nice to update this as well.

Issue #534, opened 2020-06-13
Autocomplete fails while using zsh.


evshary:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - binaries
- Version or commit hash:
  - Dashing / Eloquent
- DDS implementation:
  - ALL
- Client library (if applicable):
  - N/A

#### Steps to reproduce issue
I use zsh as my shell, but I find some problems while using `<tab>` to autocomplete.

* While I type `ros2 run <tab>`, it works well.
* However when I type `ros2 run d<tab>`, it can't show the packages which starts with 'd'.

#### Additional information
After some debugging, I find that the behavior is related to argcomplete library in Python 3.
If I remove the library `rm -rf ~/.local/lib/python3.6/site-packages/argcomplete*`, zsh works well again.
I'm not pretty sure what the root cause is.
Maybe argcomplete library does not support zsh well.

dirk-thomas:
The fact that you have `argcomplete` under `~/.local/` indicates that you installed it via `pip` rather than (or in addition to) the Debian package.

Please share the version numbers of `argcomplete` reported by these two command:

* `pip3 freeze | grep arcomplete`
* `dpkg -l | grep argcomplete`

evshary:
@dirk-thomas  You're right. I've install `argcomplete` with the command `python3 -m pip install -U argcomplete` according to [the tutorial](https://index.ros.org/doc/ros2/Installation/Foxy/Linux-Development-Setup/) before.

My result for the two commands:
```bash
$ pip3 freeze | grep arcomplete
# Nothing output
$ dpkg -l | grep argcomplete
ii  python3-argcomplete                                 1.8.1-1ubuntu1                                  all          bash tab completion for argparse (for Python 3)
ii  python3-colcon-argcomplete                          0.3.3-1                                         all          Completion for colcon command lines using argcomplete.
```

I guess you want to see the version of `argcomplete` under `~/.local/`

```
$ python3 -m pip freeze | grep argcomplete
argcomplete==1.11.1
colcon-argcomplete==0.3.3
```

dirk-thomas:
I can reproduce the problem with Eloquent (using `5.4.2-3ubuntu3.1`). In Foxy it is working though (using `5.8-3ubuntu1`).

[...]

PR #750, opened 2022-08-22
ZSH argcomplete: call compinit only if needed

mjbogusz:
Ref #534

### DIFF
--- a/ros2cli/completion/ros2-argcomplete.zsh
+++ b/ros2cli/completion/ros2-argcomplete.zsh
@@ -12,7 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-autoload -U +X compinit && compinit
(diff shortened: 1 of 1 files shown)


gbiggs (approved):

nachovizzo:
I manually patched this file and didn't solve the problem in my case. I just threw this into my .zshrc in the meantime (and it's working)

```sh
function ros2_setup () {
  source /opt/ros/humble/setup.zsh
  complete -o nospace -o default -F _python_argcomplete "ros2"
}
```

mjbogusz:
This is only one of several places, where `compinit` is called - see the list in referenced discussion. In particular I think sourcing ROS by default pulls in ros2cli, rosidl and ament_index, but depending on colcon hooks of your installed packages this may vary.

PR #753, opened 2022-09-01
Add interactive mode for message creation

### LINKED ISSUE #742: Message template autocompletion broken
When publishing a topic via `ros2 topic pub` or calling a service via `ros2 service call` the tab completion for the fields of the message is broken. The completion for the topic and message type works fine. The issue seems to be different between zsh and bash.

In zsh:
The field of the message are all handled are possible cli parameters that can be autocompleted pressing tab iterates through the fields. Each of the fields has its escaped indentation in the beginning and is individually selectable. Creating a valid message is impossible even for small msgs.
Example: 
```zsh
🐋 /home/florian/Projekt/bitbots/colcon_ws  ros2 topic pub /camera_info sensor_msgs/msg/CameraInfo -\ 0.0\
-\ 0.0\                        \ \ do_rectify:\ false\        --keep-alive                   p:\                            --qos-reliability              \ \ stamp:\                    width:\ 0\                   
-1                             \ \ frame_id:\ \'\'\           -n                             --print                        -r                             -t                             \ \ x_offset:\ 0\            
binning_x:\ 0\                 header:\                       \ \ \ \ nanosec:\ 0\           --qos-depth                    r:\                            --times                        \ \ y_offset:\ 0\            
binning_y:\ 0\                 \ \ height:\ 0\                --node-name                    --qos-durability               --rate                         -w                                                          
d:\ []\                        height:\ 0\                    --once                         --qos-history                  roi:\                          --wait-matching-subscriptions                               
distortion_model:\ \'\'\       k:\                            -p                             --qos-profile                  \ \ \ \ sec:\ 0\               \ \ width:\ 0\                                              
```

In bash:
Bash looks overall better as the message template is printed as one piece, but the completed text is not editable. Instead, the user is thrown into a new terminal line containing the command before the tab completion. 

Example:
```bash
[florian@flovat colcon_ws]$ ros2 topic pub /camera_info sensor_msgs/msg/CameraInfo 
-1
header:\^J\ \ stamp:\^J\ \ \ \ sec:\ 0\^J\ \ \ \ nanosec:\ 0\^J\ \ frame_id:\ \'\'\^Jheight:\ 0\^Jwidth:\ 0\^Jdistortion_model:\ \'\'\^Jd:\ []\^Jk:\^J-\ 0.0\^J-\ 0.0\^J-\ 0.0\^J-\ 0.0\^J-\ 0.0\^J-\ 0.0\^J-\ 0.0\^J-\ 0.0\^J-\ 0.0\^Jr:\^J-\ 0.0\^J-\ 0.0\^J-\ 0.0\^J-\ 0.0\^J-\ 0.0\^J-\ 0.0\^J-\ 0.0\^J-\ 0.0\^J-\ 0.0\^Jp:\^J-\ 0.0\^J-\ 0.0\^J-\ 0.0\^J-\ 0.0\^J-\ 0.0\^J-\ 0.0\^J-\ 0.0\^J-\ 0.0\^J-\ 0.0\^J-\ 0.0\^J-\ 0.0\^J-\ 0.0\^Jbinning_x:\ 0\^Jbinning_y:\ 0\^Jroi:\^J\ \ x_offset:\ 0\^J\ \ y_offset:\ 0\^J\ \ height:\ 0\^J\ \ width:\ 0\^J\ \ do_rectify:\ false\
--keep-alive
-n
--node-name
--once
-p
--print
--qos-depth
--qos-durability
--qos-history
--qos-profile
--qos-reliability
-r
--rate
-t
--times
-w
--wait-matching-subscriptions
[florian@flovat colcon_ws]$ ros2 topic pub /camera_info sensor_msgs/msg/CameraInfo
```

I observed this behavior on all ros2 machines I have used so far, making the cli unusable for publishing topics.

## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 22.04 LTS
- Installation type:
  - binaries 
- Version or commit hash:
  - 0.19.0-1jammy.20220711.162709
- DDS implementation:
  - Cyclone & FastDDS

#### Steps to reproduce issue
- source your terminal (zsh or bash)
- type `ros2 topic pub /camera_info sensor_msgs/msg/CameraInfo` and press `TAB`

#### Expected behavior
A nicely formatted and editable string containing the message fields with their defaults should appear. 

#### Actual behavior
Either the message is split and autocompleted in pieces (zsh) or not editable (bash)

I am not familiar with terminal formatting.

Flova:
As discussed in #742 this PR adds an interactive TUI to the `ros2 topic pub` command. 

The cli is mostly kept as it is, only an additional `-i` or `--interactive` flag is added. This allows the old behavior for scripting etc. while enhancing the experience in a manual use case.

### Added dependencies
- `python3-prompt-toolkit`
- `python3-pygments`

### Features
- Interactive editing of the message in simple terminal editor
- YAML syntax highlighting
- Mouse support in the editor
- Storage of the last send message of a given type, so large inputs can be easily reused

### Other changes:
- Slight internal restructure
- Removes parsing from the `publish` method
- Updates doc strings

Closes #742 

The implementation for the `topic pub` command serves as a proof of concept. We can expand it to services and actions if the details are ironed out.

## Demo
[asciicast](https://asciinema.org/a/8PMM29iWklVU4QtfIqD94T1Pp)

### DIFF
--- a/ros2topic/ros2topic/verb/pub.py
+++ b/ros2topic/ros2topic/verb/pub.py
@@ -12,9 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import hashlib
-import os
-import tempfile
 import time
 from typing import List, Optional, Tuple, TypeVar
 
@@ -164,19 +161,14 @@ def main(args) -> Optional[str]:
 
     if args.interactive:
         print('Interactive mode...')
-        # Read last message that was sent if it exists in temp
-        content = get_last_message_content(args.message_type)
         # Show the tui
-        orig_msg, orig_timestamp_fields = parse_msg(args.message_type, content)
-        default_msg, default_timestamp_fields = parse_msg(args.message_type)
-        content = show_interactive_tui(message_to_yaml(orig_msg), message_to_yaml(default_msg))
-        # Load msg YAML now to be sure it does not fail and we store a broken message
-        msg, timestamp_fields = parse_msg(args.message_type, content)
-        # Store the user input so we are able to load it the next time
-        store_message_content(args.message_type, content)
+        default_msg, default_timestamp_fields = parse_msg(args.message_type, args.values)
+        content = show_interactive_tui(message_to_yaml(default_msg))
     else:
-        # Parse the yaml string and get a message object of the desired type
-        msg, timestamp_fields = parse_msg(args.message_type, content)
+        content = args.values
+
+    # Parse the yaml string and get a message object of the desired type
+    msg, timestamp_fields = parse_msg(args.message_type, content)
 
     with DirectNode(args, node_name=args.node_name) as node:
         return publisher(
@@ -193,63 +185,7 @@ def main(args) -> Optional[str]:
             args.keep_alive)
 
 
-def get_history_file(msg_type: str) -> str:
-    """
-    Get paths for semi persistent history based on message name.
-
-    :param msg_type: Name of the message type
-    :returns: The path where a history file would be located if it exists
-    """
-    # Get temporary directory name
-    msg_history_cache_folder_path = os.path.join(
-        tempfile.gettempdir(), "ros_interactive_msg_cache")
-    # Create temporary history dir if needed
-    os.makedirs(msg_history_cache_folder_path, exist_ok=True)
-    # Create a file based on the message name
-    return os.path.join(
-        msg_history_cache_folder_path,
-        f'{hashlib.sha224(msg_type.encode()).hexdigest()[:20]}.yaml')
-
-
-def get_last_message_content(msg_type: str) -> str:
-    """
-    Retrieve the last message of the given type that was sent using the tui if it exists.
-
-    :param msg_type: Name of the message type
-    :returns: The YAML representation containing the last message or an empty dict
-    """
-    content = "{}"
-    try:
-        history_path = get_history_file(msg_type)
-        # Load previous values for that message type
-        if os.path.exists(history_path):
-            with open(history_path, 'r') as f:
-                content = f.read()
-    except OSError:
-        print('Unable load history...')
-    return content
-
-
-def store_message_content(msg_type: str, content: str) -> None:
-    """
-    Store the YAML for the current message in a semi persistent file.
-
-    :param msg_type: Name of the message type
-    :param content: The YAML entered by the user
-    """
-    try:
-        history_path = get_history_file(msg_type)
-        # Clear cache
-        if os.path.exists(history_path):
-            os.remove(history_path)
-        # Store last message in cache
-        with open(history_path, 'w') as f:
-            f.write(content)
-    except OSError:
-        print('Unable to store history')
-
-
-def show_interactive_tui(msg_str: str, default_msg_str: Optional[str] = None) -> str:
+def show_interactive_tui(default_msg_str: str) -> str:
     """
     Show a tui to edit a given message yaml.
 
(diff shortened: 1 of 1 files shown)


clalancette (changes requested):
I'm sorry for the long delay in getting to look at this.  I do think this is a valuable feature to have, so thanks for getting it started.

That said, I have a couple of concerns with the implementation:
1.  The storing of temporary files to reuse next time is clever, but I think it is doing too much in one PR.  I'd rather us just get the interactive feature in first, then we can think about adding in caching separately.
2. I had to rebase this whole thing because the newly added 'auto' (for std_msgs/Header) and 'now' (for builtin_interfaces/Time) keywords touch the same code.  I did that, and fixed the resulting problems, and also ended up removing the caching mechanism.
3. The way that this feature interacts with 'auto' and 'now' isn't ideal.  In particular, there is no way to specify them in interactive mode.  Additionally, if you pass them on the command line, and also enter interactive mode, they don't populate properly.  I'm not quite sure how to fix this, but it would be really good to have them integrated.

fujitatomoya (commented):
> Storage of the last send message of a given type, so large inputs can be easily reused

where is this come from? is this provided by python package?

> I did that, and fixed the resulting problems, and also ended up removing the caching mechanism.

Ah, okay isee.

> In particular, there is no way to specify them in interactive mode. Additionally, if you pass them on the command line, and also enter interactive mode

When we are using the interactive mode, can we just ask the user to fill in the contents via editor?

clalancette:
> When we are using the interactive mode, can we just ask the user to fill in the contents via editor?

Let me show you an example where it wouldn't be ideal.

Right now in Rolling, you can do this:

```
$ ros2 topic pub /foo nav_msgs/msg/Path '{header: {stamp: {sec: 25, nanosec: 30}}}'
```

And you'lll get this output:

```
publisher: beginning loop
publishing #1: nav_msgs.msg.Path(header=std_msgs.msg.Header(stamp=builtin_interfaces.msg.Time(sec=25, nanosec=30), frame_id=''), poses=[])

publishing #2: nav_msgs.msg.Path(header=std_msgs.msg.Header(stamp=builtin_interfaces.msg.Time(sec=25, nanosec=30), frame_id=''), poses=[])
```

(that's the same as in Foxy, Galactic, and Humble).

But in Rolling, you can also do this:

```
$ ros2 topic pub /foo nav_msgs/msg/Path '{header: {stamp: now}}'
```

Which will get you this:

```
publisher: beginning loop
publishing #1: nav_msgs.msg.Path(header=std_msgs.msg.Header(stamp=builtin_interfaces.msg.Time(sec=1666012466, nanosec=830824864), frame_id=''), poses=[])

publishing #2: nav_msgs.msg.Path(header=std_msgs.msg.Header(stamp=builtin_interfaces.msg.Time(sec=1666012467, nanosec=831668171), frame_id=''), poses=[])
```

(notice that the time is increasing on every publish).

If we only allow interactive mode to specify numbers for the timestamps, then we can get the previous behavior, but not the new one.  It would be much nicer if this was integrated together so that it would work for users like they expect.

exMachina316:
@Flova Just checking in to see if this pr needs any more features(i.e. the new `now` feature in rolling for time) to be merged? I would am willing to help out in any way I can!!

Flova:
I would be happy if we could merge this feature. But I don't remember exactly what the blockers where for this one.

Flova:
I can have a look at it again in a few days.

PR #765, opened 2022-09-21
Add interface type filters to ros2 interface package

DLu:
Seemed like it was missing.

### DIFF
--- a/ros2interface/ros2interface/verb/package.py
+++ b/ros2interface/ros2interface/verb/package.py
@@ -21,16 +21,41 @@ class PackageVerb(VerbExtension):
     """Output a list of available interface types within one package."""
 
     def add_arguments(self, parser, cli_name):
+        parser.add_argument(
+            '-m', '--only-msgs', action='store_true',
+            help='Print out only the message types')
+
+        parser.add_argument(
+            '-s', '--only-srvs', action='store_true',
+            help='Print out only the service types')
+
+        parser.add_argument(
+            '-a', '--only-actions', action='store_true',
+            help='Print out only the action types')
+
         arg = parser.add_argument(
             'package_name',
             help="Name of the ROS package (e.g. 'example_interfaces')")
         arg.completer = package_name_completer
 
     def main(self, *, args):
+        types_to_filter = []
+        if args.only_msgs or args.only_srvs or args.only_actions:
+            if not args.only_msgs:
+                types_to_filter.append('msg')
+            if not args.only_srvs:
+                types_to_filter.append('srv')
+            if not args.only_actions:
+                types_to_filter.append('action')
+
         try:
             interfaces = get_interfaces([args.package_name])
         except LookupError as e:
             return str(e)
         for package_name in sorted(interfaces):
             for interface_name in interfaces[package_name]:
+                if types_to_filter:
+                    interface_type = interface_name[:interface_name.index('/')]
+                    if interface_type in types_to_filter:
+                        continue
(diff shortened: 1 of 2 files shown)


ros2interface/ros2interface/verb/package.py
```diff
@@ -21,16 +21,41 @@ class PackageVerb(VerbExtension):
...
         try:
             interfaces = get_interfaces([args.package_name])
         except LookupError as e:
             return str(e)
         for package_name in sorted(interfaces):
             for interface_name in interfaces[package_name]:
+                if types_to_filter:
+                    interface_type = interface_name[:interface_name.index('/')]
+                    if interface_type in types_to_filter:
+                        continue
```
clalancette, line 60: I don't feel like this is the right way to filter here.

In particular, while we almost certainly aren't going to change the prefix for `msg`, `srv`, and `action`, we also have dedicated functions for getting this information.  Thus, I think we should refactor this whole block to be something like:

```
interfaces = []
if not args.only_msgs and not args.only_srvs and not args.only_actions:
    interfaces = get_interfaces([args.package_name])
else:
    if args.only_msgs:
        interfaces += get_message_interfaces([args.package_name])
    if args.only_srvs:
        interfaces += get_service_interfaces([args.package_name])
    if args.only_actions:
        interfaces += get_action_interfaces([args.package_name])

for package_name in sorted(interfaces):
    for interface_name in interfaces[package_name]:
        print(f'{package_name}/{interface_name}')

```

clalancette (changes requested):

ros2interface/ros2interface/verb/package.py
```diff
@@ -11,24 +11,52 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+import collections
```
clalancette, line 14: You need whitespace both before and after this to pass flake8.

```suggestion

import collections

```

ros2interface/test/test_cli.py
```diff
@@ -200,6 +200,22 @@ def test_package_on_test_msgs(self):
...
+        assert interface_command.exit_code == launch_testing.asserts.EXIT_OK
+        output_lines = interface_command.output.splitlines()
+        assert launch_testing.tools.expect_output(
+            expected_lines=itertools.repeat(
+                re.compile(r'test_msgs/srv/[A-z0-9_]+'), len(output_lines)
+            ),
+            lines=output_lines,
+            strict=True
+        )
+        assert all(interface in output_lines for interface in some_services_from_test_msgs)
```
clalancette, line 217: It is probably overkill, but I'd feel a little better about this if we had 3 tests, one for each of `-s`, `-m`, and `-a`.  Also, I'm concerned about a 2 second timeout for the shutdown, particularly on Windows; I'll suggest 5 seconds instead.

clalancette (changes requested):
A couple more minor things to fix, then I think this will be ready for CI.

fujitatomoya (commented):
lgtm with @clalancette comment addressed

clalancette (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18768/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13291/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19509/)

PR #767, opened 2022-09-29
Implement 'latched' launch argument for echo verb.

RobertBlakeAnderson:
You can now conveniently echo latch-like topics via:
`ros2 topic echo --latched <topic>`

### DIFF
--- a/ros2topic/ros2topic/api/__init__.py
+++ b/ros2topic/ros2topic/api/__init__.py
@@ -217,3 +217,7 @@ def add_qos_arguments_to_argument_parser(
         help='Quality of service durability setting to {} with '
              '(overrides durability value of --qos-profile option, default: {})'
              .format(verb, default_profile.durability.short_key))
+    parser.add_argument(
+        '--latched',
+        action='store_true',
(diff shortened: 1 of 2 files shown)


clalancette (commented):
So we already have this functionality, both automatically (since Humble), and via `--qos-durability transient_local` (though I agree that is somewhat annoying to type).

Additionally, I don't think we should use the term "latched" here, as `transient_local` isn't exactly the same thing as the concept of `latched` from ROS 1 (see https://github.com/ros2/ros2/issues/464 for lots of details).

Because of all of this, I don't really think we should add this, but I'd be curious to hear from others.  @ros2/team, thoughts?

jacobperron:
I agree that we should avoid using the term "latched".

If we want a shortcut for `--qos-durability transient_local`, we could add an option `-t` or `-d` (for "durable"). Though since durability is already decided automatically in later distros, IMO it's not worth adding this feature.

clalancette:
Given the above, I'm going to close this one out.  But if you still think this is valuable, please feel free to keep commenting or reopen.

PR #782, opened 2022-12-12
I want to add `ament_auto` option to ros2 pkg create

Ar-Ray-code:
Our team uses ament_cmake_auto as buildtools, but since it cannot be generated with the ros2 pkg create command, it is copied from an existing package.
Please consider incorporating the following differences to simplify package generation.


clalancette (changes requested):
So the main issue with this is that `ament_cmake_auto` isn't a build type.  Build types are the things that our tools (colcon, buildfarm, etc) understand so that they can properly prepare a package.  Right now, that only includes `cmake`, `ament_cmake`, and `ament_python`, I believe.

What you are looking to do is to use additional `ament_cmake` helpers, which is a different thing.  There are two ways I can see going about this:

1.  Just adding some additional comments to the existing https://github.com/ros2/ros2cli/blob/rolling/ros2pkg/ros2pkg/resource/ament_cmake/CMakeLists.txt.em , explaining more about how to use `ament_auto`.
2. Adding a new command-line option to select `ament-auto`.  This is a little dicey as it doesn't apply when the `cmake` or `ament_python` types are used, but we could detect and error on that.

Ar-Ray-code:
Thank you for your response.
I’ll following your 2nd suggestion.

Ar-Ray-code:
I added `--ament-auto` option and revert `--build-type`.

Test commands used are shown below.


### ament_cmake_auto (success)

```bash
ros2 pkg create --build-type ament_cmake --node-name my_node my_package --dependencies rclcpp std_msgs sensor_msgs  --license Apache-2.0 --library-name libname --description "test pkg" --maintainer-name "Example" --maintainer-email "example@example.com" --package-format 3 --ament-auto
```


### ament_python ( or cmake, error)

```bash
root@a633c9050ddb:/ros2_ws# ros2 pkg create --build-type ament_python --node-name my_node my_package --dependencies rclpy std_msgs sensor_msgs --license Apache-2.0 --library-name libname --description "test pkg" --maintainer-name "Example" --maintainer-email "example@example.com" --package-format 3 --ament-auto
[ERROR] ament_auto can only be used with ament_cmake packages
root@a633c9050ddb:/ros2_ws# 
```

What do you think?

tomkimsour:
Any reason why this PR has been closed ? Would it make sense to open it again ?

PR #791, opened 2023-01-17
Refactor common types

arjo129:
A lot of types across verbs in ros2 topic are shared. This PR refactors the types to be used by different verbs preventing redefinition of the same types in different verbs. There is no behaviour changed introduced.

Signed-off-by: Arjo Chakravarty <arjo@openrobotics.org>

Issue #529, opened 2020-06-10
Feature Request - ros2 topic echo --once


craigh92:
## Feature request

Add an option `--once` to the `ros2 topic echo` command, that will cause the command to exit after it has received and printed one message. Also add a `--timeout ` option that will cause the command to exit if a certain period of time has elapsed.

#### Feature description

If messages are being published representing the state of the system, then the `ros2 topic echo` is useful for querying the current state. The last message is often all that the user is interested in. `ros2 topic echo --qos-durability=transient_local --once` will be a quick way of sending a "Get" command, that gets the current value of something in the system.

It will also be useful for writing tests. e.g

```bash
source /opt/ros/eloquent/setup.bash
source ./install/setup.sh

#Launch all nodes in a background process
ros2 launch -a launch/my_launch.py &>ros_out_log &
pid=$!

#Test initial state of robot1
pos=$(ros2 topic echo --qos-durability=transient_local --once robot1/position msgs/PositionMsg)
    #test $pos...

#Send move command
ros2 service call robot1/move srvs/MoveSrv {x: 1, y: 2}

sleep 2

#Test new poistion of robot1
pos=$(ros2 topic echo --qos-durability=transient_local --once robot1/position msgs/PositionMsg)
    #test $pos...

#Shutdown all nodes

kill -2 $pid

echo "waiting for nodes to shutdown..."

nodes=$(ros2 node list)
numnodes=$(echo $nodes | wc -w)
while [[ $numnodes != 0 ]]; do
  nodes=$(ros2 node list)
  numnodes=$(echo $nodes | wc -w)
  echo "$numnodes remaining"
  sleep 1
done

```

This is much easier than having to create subscribers, clients, etc in a python file and testing it that way.

#### Implementation considerations

This should be pretty easy to implement. Instead of `rclpy.spin(node)` [here](https://github.com/ros2/ros2cli/blob/f7bebe0f88e811104271758f038cef8abb5285d9/ros2topic/ros2topic/verb/echo.py#L117), replace with `rclpy.spin_untill_future_complete()` (which already has a timeout feature built into it), and then set the future as complete in the subscription callback.

I've made a start here: https://github.com/craigh92/ros2cli/commit/10e716dacfd838799a0510bbf67f45f87eff152c
I've not contributed before so I'm not sure if I'm using `parser.add_argument` correctly.

craigh92:
@dirk-thomas How do I test my PR?

[...]

PR #792, opened 2023-01-17
Adds a timeout feature to rostopic echo

arjo129:
Builds on top of:
* #791  - Review and merge this and you'll see the change is minimal.

This PR adds a `--time-out` feature which allows you to set the amount of time that rostopic echo should wait. This feature was
requested by https://github.com/ros2/ros2cli/issues/529, however, we only added `--once` and no time out. I have a use case where this timeout could be useful hence I added it

### DIFF
--- a/ros2topic/ros2topic/api/__init__.py
+++ b/ros2topic/ros2topic/api/__init__.py
@@ -38,6 +38,21 @@ def unsigned_int(string):
         raise ArgumentTypeError('value must be non-negative integer')
     return value
 
+def positive_int(string):
+    try:
+        value = int(string)
+    except ValueError:
+        value = -1
+    if value <= 0:
+        raise ArgumentTypeError('value must be a positive integer')
+    return value
+
+def positive_float(inval):
+    ret = float(inval)
+    if ret <= 0.0:
+        # The error message here gets completely swallowed by argparse
+        raise ValueError('Value must be positive')
+    return ret
 
 def get_topic_names_and_types(*, node, include_hidden_topics=False):
     topic_names_and_types = node.get_topic_names_and_types()
--- a/ros2topic/ros2topic/api/__init__.py
+++ b/ros2topic/ros2topic/api/__init__.py
@@ -52,12 +52,11 @@ def positive_int(string):
 def positive_float(inval):
     try:
         ret = float(inval)
-        if ret <= 0.0:
-            raise ArgumentTypeError('Value must be positive')
-    except:
+    except ValueError:
         raise ArgumentTypeError('Expects a floating point number')
+    if ret <= 0.0:
+        raise ArgumentTypeError('Value must be positive')
     return ret
-
 
 def get_topic_names_and_types(*, node, include_hidden_topics=False):
     topic_names_and_types = node.get_topic_names_and_types()
--- a/ros2topic/ros2topic/api/__init__.py
+++ b/ros2topic/ros2topic/api/__init__.py
@@ -50,10 +50,12 @@ def positive_int(string):
 
 
 def positive_float(inval):
-    ret = float(inval)
-    if ret <= 0.0:
-        # The error message here gets completely swallowed by argparse
-        raise ArgumentTypeError('Value must be positive')
+    try:
+        ret = float(inval)
+        if ret <= 0.0:
+            raise ArgumentTypeError('Value must be positive')
+    except:
+        raise ArgumentTypeError('Expects a floating point number')
     return ret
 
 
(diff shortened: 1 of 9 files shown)


ros2topic/ros2topic/verb/echo.py
```diff
@@ -103,6 +104,8 @@ def add_arguments(self, parser, cli_name):
                                                  'as well as m (the message).')
         parser.add_argument(
             '--once', action='store_true', help='Print the first message received and then exit.')
+        parser.add_argument(
+            '--time-out', type=positive_float, help='Set a timeout for waiting')
```
fujitatomoya, line 108: I see the option `--timeout` in daemon, not hyphen separated between. probably use the same format?

```suggestion
            '--timeout', metavar='N', type=positive_float, help='Set a timeout in seconds for waiting')
```

probably set the default? what happens if user issues `ros2 topic echo --timeout /chatter`?

fujitatomoya (changes requested):
test would be ideal, to check certain published messages are printed in specific time window until it times out.

ros2topic/ros2topic/verb/echo.py
arjo129: Thanks! Addressed in ecde453 and cc58815

fujitatomoya:
> test would be ideal, to check certain published messages are printed in specific time window until it times out.

i think you can refer to,

https://github.com/ros2/ros2cli/blob/cf020c0f0f90ded9aca6ebac806c78a387347a25/ros2topic/test/test_echo_pub.py#L393-L435

arjo129:
Thanks working on it.

ros2topic/test/test_echo_pub.py
```diff
@@ -433,3 +434,26 @@ def publish_message():
             # Cleanup
             self.node.destroy_timer(publish_timer)
             self.node.destroy_publisher(publisher)
+
+    @launch_testing.markers.retry_on_failure(times=5)
+    def test_echo_timeout(self, launch_service, proc_info, proc_output):
```
fujitatomoya, line 439: this verifies that the command times out eventually within 10 seconds, but specified time window.

I would do,

- create publisher publishes message every 2 seconds. (2 seconds is just an example)
- issue `ros2 topic echo --timeout 5`.
- check the output message that has 2 published message in it.

what do you think?

clalancette (changes requested):

ros2topic/test/test_echo_pub.py
fujitatomoya: > but we have to make sure not to make it flaky (particularly on Windows

👍

ros2topic/ros2topic/api/__init__.py
```diff
@@ -50,13 +50,14 @@ def positive_int(string):
...
+    try:
+        ret = float(inval)
+    except ValueError:
+        raise ArgumentTypeError('Expects a floating point number')
     if ret <= 0.0:
-        # The error message here gets completely swallowed by argparse
         raise ArgumentTypeError('Value must be positive')
     return ret
 
-
```
clalancette, line 59: I'm pretty sure you need to keep this whitespace to pass flake8 checks.

ros2topic/test/test_echo_pub.py
clalancette: This comment is still open, I believe.  For my part, I actually think that we don't need the publication; we are testing that elsewhere.  Just testing that the timeout works is enough for me.  But let's see if @fujitatomoya is OK with that.

clalancette (changes requested):

ros2topic/test/test_echo_pub.py
```diff
@@ -433,3 +433,24 @@ def publish_message():
...
+        topic = '/clitest/topic/echo_timeout'
+
+        command_action = ExecuteProcess(
+            cmd=['ros2', 'topic', 'echo', '--timeout', '5', topic, 'std_msgs/msg/String'],
```
clalancette, line 442: If we are just going to test that the timeout works, I'll suggest reducing this to 2 (or maybe even 1) seconds.  Just to keep CI times down.

arjo129: [5c620d6](https://github.com/ros2/ros2cli/pull/792/commits/5c620d6b3ae283e803ad9164ce5454397a412551)

clalancette (approved):
This looks good to me.  I'll run CI on it next.

But I'd like to get @fujitatomoya's opinion before I merge.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18016/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12542/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18651/)

fujitatomoya (approved):
@arjo129 LTGM!

PR #797, opened 2023-01-31
Adds a `--max-wait-time` option to ros2 topic pub

arjo129:
When we run `ros2 topic pub -w 1 ...` we can wait for a fixed number
of connections prior to shutting down. This is useful in many scenarios,
however when writing automated tests it may be useful to time out if the
correct number of connections are not recieved in a given amount of
time. This PR adds a `--max-wait-time` option which allows users to set
a timeout on how long to wait for subscribers.

Includes changes from #792 

### DIFF
--- a/ros2cli/completion/ros2-argcomplete.zsh
+++ b/ros2cli/completion/ros2-argcomplete.zsh
@@ -12,7 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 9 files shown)


clalancette:
With #792 now merged, this needs a rebase.  I'll do a review once that happens (and it comes out of draft).

arjo129:
Closing in favor of #800

PR #800, opened 2023-02-03
Adds a `--max-wait-time` option to `ros2 topic pub` 

arjo129:
Supercedes #797

When we run `ros2 topic pub -w 1 ...` we can wait for a fixed number
of connections prior to shutting down. This is useful in many scenarios,
however when writing automated tests it may be useful to time out if the
correct number of connections are not recieved in a given amount of
time. This PR adds a `--max-wait-time` option which allows users to set
a timeout on how long to wait for subscribers.

### DIFF
--- a/ros2topic/ros2topic/verb/pub.py
+++ b/ros2topic/ros2topic/verb/pub.py
@@ -156,7 +156,7 @@ def publisher(
     pub = node.create_publisher(msg_module, topic_name, qos_profile)
 
     if wait_matching_subscriptions == 0 and max_wait_time is not None:
-        return '--max-wait-time-secs option is only effective' +
+        return '--max-wait-time-secs option is only effective' \
             ' with --wait-matching-subscriptions, --once or --times'
 
     times_since_last_log = 0
--- a/ros2topic/ros2topic/verb/pub.py
+++ b/ros2topic/ros2topic/verb/pub.py
@@ -154,7 +154,7 @@ def publisher(
 
     pub = node.create_publisher(msg_module, topic_name, qos_profile)
 
-    if wait_matching_subscriptions is None and max_wait_time is not None:
+    if wait_matching_subscriptions == 0 and max_wait_time is not None:
         return '--max-wait-time-secs option is only effective with --wait-matching-subscriptions, --once or --times'
 
     times_since_last_log = 0
--- a/ros2topic/ros2topic/verb/pub.py
+++ b/ros2topic/ros2topic/verb/pub.py
@@ -167,7 +167,7 @@ def publisher(
             print(
                 f'Waiting for at least {wait_matching_subscriptions} matching subscription(s)...')
         if max_wait_time is not None and max_wait_time <= total_wait_time:
-            return f'Timed out waiting for subscribers: Expected {wait_matching_subscriptions}' +
+            return f'Timed out waiting for subscribers: Expected {wait_matching_subscriptions}' \
                 f' subcribers but only got {pub.get_subscription_count()} subscribers'
         times_since_last_log = (times_since_last_log + 1) % 10
         time.sleep(DEFAULT_WAIT_TIME)
--- a/ros2topic/ros2topic/verb/pub.py
+++ b/ros2topic/ros2topic/verb/pub.py
@@ -33,6 +33,7 @@ from rosidl_runtime_py.utilities import get_message
 import yaml
 
 MsgType = TypeVar('MsgType')
+DEFAULT_WAIT_TIME = 0.1
 
 
 def nonnegative_int(inval):
@@ -83,9 +84,10 @@ class PubVerb(VerbExtension):
                 'Wait until finding the specified number of matching subscriptions. '
                 'Defaults to 1 when using "-1"/"--once"/"--times", otherwise defaults to 0.'))
         parser.add_argument(
-            '--max-wait-time', type=positive_float, default=None,
+            '--max-wait-time-secs', type=positive_float, default=None,
             help=(
-                'This sets the maximum wait time if --wait-until-matching-subscriptions is set'))
+                'This sets the maximum wait time in seconds if --wait-until-matching-subscriptions is set.'
+                'By default, this flag is not set meaning the subscriber will wait endlessly.'))
         parser.add_argument(
             '--keep-alive', metavar='N', type=positive_float, default=0.1,
             help='Keep publishing node alive for N seconds after the last msg '
@@ -112,6 +114,9 @@ def main(args):
     if args.once:
         times = 1
 
+    if args.wait_matching_subscriptions is None and args.max_wait_time_secs is not None:
+        return '--max-wait-time-secs option is only effective with --wait-matching-subscriptions'
+
     with DirectNode(args, node_name=args.node_name) as node:
         return publisher(
             node.node,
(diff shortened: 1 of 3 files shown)


clalancette:
Just as an FYI, in general we prefer rebases of old PRs rather than new PRs.  That is, it would have been preferable here to keep #797 open, and just rebase it, rather than close it and make this new one.  That keeps all of the conversations in the same place and ensures it is easier to follow later.

That said, it is already done here, so let's just go forward with this one.

ros2topic/ros2topic/verb/pub.py
```diff
@@ -82,6 +82,10 @@ def add_arguments(self, parser, cli_name):
             help=(
                 'Wait until finding the specified number of matching subscriptions. '
                 'Defaults to 1 when using "-1"/"--once"/"--times", otherwise defaults to 0.'))
+        parser.add_argument(
+            '--max-wait-time', type=positive_float, default=None,
+            help=(
+                'This sets the maximum wait time if --wait-until-matching-subscriptions is set'))
```
clalancette, line 88: I think we should be clear about the units here.  Presumably this is in seconds, so at the very least we should say that in the help.  We also might consider making the argument `--max-wait-time-sec`, to make it really clear.  Same goes for the variable names below.

ros2topic/test/test_echo_pub.py
```diff
@@ -176,6 +176,32 @@ def message_callback(msg):
...
+            },
+            output='screen'
+        )
+        with launch_testing.tools.launch_process(
+            launch_service, command_action, proc_info, proc_output,
+            output_filter=launch_testing_ros.tools.basic_output_filter(
+                filtered_rmw_implementation=get_rmw_implementation_identifier()
+            )
+        ) as command:
+            assert command.wait_for_shutdown(timeout=2)
```
clalancette, line 196: Let's increase this timeout to 5, again to make sure Windows has time to actually launch the process.

clalancette (changes requested):

ros2topic/ros2topic/verb/pub.py
fujitatomoya: besides, probably add default behavior explanation, that is wait forever until it finds expected matching subscription.

ros2topic/ros2topic/verb/pub.py
```diff
@@ -147,14 +153,19 @@ def publisher(
...
 
     times_since_last_log = 0
+    total_wait_time = 0
     while pub.get_subscription_count() < wait_matching_subscriptions:
         # Print a message reporting we're waiting each 1s, check condition each 100ms.
         if not times_since_last_log:
             print(
                 f'Waiting for at least {wait_matching_subscriptions} matching subscription(s)...')
+        if max_wait_time is not None and max_wait_time <= total_wait_time:
+            return 'Timed out waiting for subscribers'
```
fujitatomoya, line 163: it would be helpful to print how many subscriptions we are missing when it times out?

ros2topic/ros2topic/verb/pub.py
```diff
@@ -147,14 +153,19 @@ def publisher(
...
         # Print a message reporting we're waiting each 1s, check condition each 100ms.
         if not times_since_last_log:
             print(
                 f'Waiting for at least {wait_matching_subscriptions} matching subscription(s)...')
+        if max_wait_time is not None and max_wait_time <= total_wait_time:
+            return 'Timed out waiting for subscribers'       
         times_since_last_log = (times_since_last_log + 1) % 10
-        time.sleep(0.1)
-
+        WAIT_TIME = 0.1
```
fujitatomoya, line 165: I am not familiar with how python interprets the code at runtime. but considering based C/C++, probably we can avoid store the same data to the variable? instead, we can declare this `WAIT_TIME` after `import` statement like `DEFAULT_WAIT_TIME` and use it in this while loop? that would be better for code maintenance.

ros2topic/ros2topic/verb/pub.py
```diff
@@ -119,6 +123,7 @@ def main(args):
             times,
             args.wait_matching_subscriptions
             if args.wait_matching_subscriptions is not None else int(times != 0),
+            args.max_wait_time,
```
fujitatomoya, line 126: how about checking the option dependencies before calling in here?

```python
if args.wait_matching_subscriptions is None and args.max_wait_time is not None:
    return '--max-wait-time option is only effective with --wait-matching-subscriptions'
```

maybe too much, cz we can actually ignore this `--max-wait-time` option.

ros2topic/ros2topic/verb/pub.py
arjo129: See c6dda6e

ros2topic/test/test_echo_pub.py
arjo129: See c6dda6e

ros2topic/ros2topic/verb/pub.py
arjo129: See c6dda6e

ros2topic/ros2topic/verb/pub.py
arjo129: c6dda6e

ros2topic/ros2topic/verb/pub.py
arjo129: c6dda6e

fujitatomoya (approved):

arjo129:
Added tests in ccb2864

ros2topic/test/test_echo_pub.py
```diff
@@ -176,6 +176,88 @@ def message_callback(msg):
...
+        self.node.create_subscription(String, topic, message_callback, 1)
+        with launch_testing.tools.launch_process(
+            launch_service, command_action, proc_info, proc_output,
+            output_filter=launch_testing_ros.tools.basic_output_filter(
+                filtered_rmw_implementation=get_rmw_implementation_identifier()
+            )
+        ) as command:
+            self.executor.spin_until_future_complete(future, timeout_sec=10)
+            assert command.wait_for_shutdown(timeout=20)
+        assert self.recv_msg_count == 1
```
clalancette, line 260: Is this line right?  We never set `self.recv_msg_count` in the code above.  Shouldn't we be checking `received_message_count`?

arjo129: [9c2ad47](https://github.com/ros2/ros2cli/pull/800/commits/9c2ad47d4f35edb2b46bfd89a22c0aefeb2b59bf)

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18047/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12577/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18692/)

clalancette:
@arjo129 It looks like one of the new tests is triggering CI.  Can you take a look?

arjo129:
Is it possible to check CI again? I've changed the problematic test case.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18096/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12628/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18740/)

clalancette:
Another try with flake8 warnings fixed:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18098/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12630/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18742/)

PR #801, opened 2023-02-03
Resolve #790

### LINKED ISSUE #790: Subdirectories not installed with ros2pkg's ament_python template
Currently, the ament_python template has ``packages=[package_name]``, as seen below:

https://github.com/ros2/ros2cli/blob/cf020c0f0f90ded9aca6ebac806c78a387347a25/ros2pkg/ros2pkg/resource/ament_python/setup.py.em#L8

However, with this configuration, it seems that subdirectories are not installed when running ``colcon build``.

Perhaps it is better to have in the template: 

```python
from setuptools import find_packages  # at top of setup.py.em

packages=find_packages(exclude=['test']),
```

like many ros2 python packages do, including most packages in ros2cli do. Happy to raise a PR if others agree with this change.

Related issue: https://github.com/ros-sports/soccer_ipm/pull/7

ijnek:
Resolves: #790

### DIFF
--- a/ros2pkg/ros2pkg/resource/ament_python/setup.py.em
+++ b/ros2pkg/ros2pkg/resource/ament_python/setup.py.em
@@ -1,11 +1,11 @@
-from setuptools import setup
+from setuptools import find_packages, setup
 
 package_name = '@project_name'
 
 setup(
     name=package_name,
     version='0.0.0',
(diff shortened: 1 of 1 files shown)


gbiggs (approved):

ijnek:
Thanks @gbiggs !

Issue #799, opened 2023-02-02
Check for existence of Node in verbs can cause false failures when calling


destogl:
Hi,

in ros2_control we [were using CLI calls to load parameters](https://github.com/ros-controls/ros2_control/pull/910) and has issues, especially with FastDDS that calls were failing because the node could not be discovered on time. We have overcome this in the referenced PR, still it would be good to make CLI more robust. The issue is partially caused by a single check of node existence done in these lines:

https://github.com/ros2/ros2cli/blob/bcb1414ea87ba164ea62c918f23b2b761f4c5a4a/ros2param/ros2param/verb/load.py#L45-L50

This check depends on strongly how fast discovery of DDS implementation is working. There are two possible ways to make this more robust:

1. Remove this check because it is actually not needed – since the method called has already timeout if service cannot be found. Service should time out in both cases, if target node does not exist and also if service does not exist.
     https://github.com/ros2/ros2cli/blob/bcb1414ea87ba164ea62c918f23b2b761f4c5a4a/ros2param/ros2param/verb/load.py#L53
2. Add timeout to this check, repeating it multiple times until the timeout. This would make this check much more robust against slower DDS discovery capabilities.

This issue is repeated on multiple verbs in `ros2param` library, and it would make sense to adjust it everywhere. 

Which of the proposed solution do you think is preferred?

fujitatomoya:
@destogl thanks for posting issue.

using `NodeStrategy`, there is an assumption that `ros 2 daemon` can response quickly the node availability based on cached node graph if ros 2 daemon is running the that host machine. Do you have this problem when ros 2 daemon is running?

> Which of the proposed solution do you think is preferred?

I would keep node check before service call, since `node is not available` and `service is not available` is not the same meaning for user. I am inclined to take an improvement `2` that makes sense. we can add an `timeout` option until it times out or node and service becomes available.

destogl:
> @destogl thanks for posting issue.
> 
> using `NodeStrategy`, there is an assumption that `ros 2 daemon` can response quickly the node availability based on cached node graph if ros 2 daemon is running the that host machine. Do you have this problem when ros 2 daemon is running?

yes, we could observe the issue with and without daemon running. The issue that we were starting plenty of nodes at once. 
I know that our approach was not well-designed, so we changed it because of that. The idea here is to increase robustness. 

> > Which of the proposed solution do you think is preferred?
> 
[...]

PR #802, opened 2023-02-08
add timeout option for ros2param to find node.

fujitatomoya:
address https://github.com/ros2/ros2cli/issues/799

### DIFF
--- a/ros2param/ros2param/verb/describe.py
+++ b/ros2param/ros2param/verb/describe.py
@@ -16,8 +16,8 @@ from ros2cli.node.direct import DirectNode
 from ros2cli.node.strategy import add_arguments
 from ros2cli.node.strategy import NodeStrategy
 from ros2node.api import get_absolute_node_name
-from ros2node.api import get_node_names
 from ros2node.api import NodeNameCompleter
+from ros2node.api import wait_for_node
 from ros2param.api import call_describe_parameters
 from ros2param.api import get_parameter_type_string
 from ros2param.api import ParameterNameCompleter
@@ -38,16 +38,16 @@ class DescribeVerb(VerbExtension):
             help='Consider hidden nodes as well')
         arg = parser.add_argument(
             'parameter_names', nargs='+', help='Names of the parameters')
+        parser.add_argument(
+            '--timeout', metavar='N', type=int, default=1,
+            help='Wait for N seconds until node becomes available (default %(default)s sec)')
         arg.completer = ParameterNameCompleter()
 
     def main(self, *, args):  # noqa: D102
-        with NodeStrategy(args) as node:
-            node_names = get_node_names(
-                node=node, include_hidden_nodes=args.include_hidden_nodes)
-
         node_name = get_absolute_node_name(args.node_name)
-        if node_name not in {n.full_name for n in node_names}:
-            return 'Node not found'
+        with NodeStrategy(args) as node:
+            if not wait_for_node(node, node_name, args.include_hidden_nodes, args.timeout):
+                return 'Node not found'
 
         with DirectNode(args) as node:
             response = call_describe_parameters(
--- a/ros2param/test/test_verb_load.py
+++ b/ros2param/test/test_verb_load.py
@@ -269,11 +269,23 @@ class TestVerbLoad(unittest.TestCase):
             strict=False
         )
 
+    def test_verb_load_timeout(self):
+        with self.launch_param_load_command(
+            arguments=['invalid_node', 'invalid_path', '--timeout', '2']
+        ) as param_load_command:
+            assert param_load_command.wait_for_shutdown(timeout=TEST_TIMEOUT)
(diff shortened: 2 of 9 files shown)


fujitatomoya:
@audrow @gbiggs what do you think? I will add `timeout` option to other `ros2 param` sub-command as well.

ros2param/test/test_verb_load.py
```diff
@@ -269,11 +269,23 @@ def test_verb_load_invalid_path(self):
...
+            text=param_load_command.output,
+            strict=False
+        )
+
     def test_verb_load(self):
         with tempfile.TemporaryDirectory() as tmpdir:
             filepath = self._write_param_file(tmpdir, 'params.yaml')
             with self.launch_param_load_command(
-                arguments=[f'{TEST_NAMESPACE}/{TEST_NODE}', filepath]
+                arguments=[f'{TEST_NAMESPACE}/{TEST_NODE}', filepath, '--timeout', '3']
```
fujitatomoya, line 288: these unrelated tests are flaky on my local environment. with `timeout` option from this PR, it can avoid the unstable test result.

iuhilnehc-ynos (commented):
Just a tiny minor suggestion.

iuhilnehc-ynos:
If the `wait_for_node` is a helper code that will be used in somewhere else, what about using `wait_for` in the `wait_for_node`?

iuhilnehc-ynos:
BTW:

https://github.com/ros2/ros2cli/blob/1e3f1a0cb2a7686c9a5d01dea0622149c3b8c4a6/ros2cli/ros2cli/helpers.py#L40-L44
should be updated, because `predicate()` will be called twice if `predicate()` returns True at the beginning.
(It reminded me of the logic similar to https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/include/std/condition_variable#L348-L351.)

```python
    while not predicate():
        if time.time() > deadline:
            return predicate()
        time.sleep(period)
    return True
```

fujitatomoya:
I believe that same option can be applied at lease `ros2param` sub command, since other sub commands have possibly same problem.

ros2param/test/test_verb_load.py
```diff
@@ -269,11 +269,23 @@ def test_verb_load_invalid_path(self):
             strict=False
         )
 
+    def test_verb_load_timeout(self):
```
fujitatomoya, line 272: I think `timeout` option verification is completed here, it uses the same exact login and methods underneath, so that we do not need to integrate the test for each subcommand with `timeout` option.

iuhilnehc-ynos (approved):

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18033/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12557/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18678/)

fujitatomoya:
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12564/)

fujitatomoya:
@audrow @gbiggs friendly ping, requesting another review.

ros2param/ros2param/verb/describe.py
```diff
@@ -38,16 +38,16 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
             help='Consider hidden nodes as well')
         arg = parser.add_argument(
             'parameter_names', nargs='+', help='Names of the parameters')
+        parser.add_argument(
+            '--timeout', metavar='N', type=int, default=1,
```
destogl, line 42: I propose to add a longer default timeout, e.g., 3 or 5 seconds. Nevertheless, I am uncertain if this will poolin result. In the [`spawner`](https://github.com/ros-controls/ros2_control/blob/master/controller_manager/controller_manager/spawner.py#L59-L67) of controller manager, we are waiting for 10 seconds, but check every 0.2 seconds if node is available.

destogl (approved):
Thanks for following up on this!

I have just a small proposal to increase the default timeout. As you see in the tests, 1, second could be too low.

ros2param/ros2param/verb/describe.py
fujitatomoya: @destogl thank you for reviewing and sharing the feedback.

> I propose to add a longer default timeout, e.g., 3 or 5 seconds.

I would like to stick with `1 sec` in default, since

- user can have option to set the timeout. (this is bottom line, i think)
- how long timeout should be is really use case dependent. probably `1 sec` is too long for some users, too short for some other users...
- i think it would be probably nicer to response the result for user so that user does not wait for several seconds to see the result at 1st step.

audrow (approved):
Looks good to me - thanks, @fujitatomoya!

fujitatomoya:
@audrow thanks for the review, i will go ahead to merge this.

PR #804, opened 2023-02-10
Fix printing of integer and double arrays.

clalancette:
In the ros2param code, we get integer and double arrays as literal Python arrays.  When printing that out in ros2 param get, they look something like:

Double values are: array('d', [1.4, 1.5])

But that isn't really what we want the user to see.  So call .tolist() on these before we print, so it looks like:

Double values are: [1.4, 1.5]

### DIFF
--- a/ros2param/ros2param/verb/set.py
+++ b/ros2param/ros2param/verb/set.py
@@ -57,7 +57,7 @@ class SetVerb(VerbExtension):
 
         with DirectNode(args) as node:
             parameter = Parameter()
-            Parameter.name = args.parameter_name
+            parameter.name = args.parameter_name
             parameter.value = get_parameter_value(string_value=args.value)
 
             response = call_set_parameters(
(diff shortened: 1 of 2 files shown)


ros2param/ros2param/verb/set.py
```diff
@@ -57,7 +57,7 @@ def main(self, *, args):  # noqa: D102
 
         with DirectNode(args) as node:
             parameter = Parameter()
-            Parameter.name = args.parameter_name
+            parameter.name = args.parameter_name
```
fujitatomoya, line 60: just curious  that this is bug? did it work okay?

fujitatomoya (approved):
lgtm

ros2param/ros2param/verb/set.py
clalancette: That's a good question.  I just noticed it while looking at the code.

It seemed to have worked fine before, but I think that was because we were setting the `name` field on the `Parameter` *class*, not the object.  The proper thing to do here is to set it on the object itself, which is what this change does.  Does that make sense?

PR #807, opened 2023-02-27
Fix two small flake8 errors.

clalancette:
These somehow slipped through CI earlier.


mjcarroll (approved):

clalancette:
Closing this as it is the same as #808 (which was opened in the same minute).

PR #812, opened 2023-03-20
Add QoS parameter to `ros2 service call`

bastianhjaeger:
In case of service setups diverging from the default QoS on a service, a service may not be called by cli due to mismatching QoS profiles. 
This PR should fix it.


ros2service/ros2service/verb/call.py
```diff
@@ -16,15 +16,26 @@
...
 import yaml
 
 
+def get_pub_qos_profile():
+    return QoSProfile(
+        reliability=QoSReliabilityPolicy.RELIABLE,
+        durability=QoSDurabilityPolicy.TRANSIENT_LOCAL,
+        depth=1)
+
+
```
fujitatomoya, line 38: Why we need to statically declare default profile only for `ros2 service call`? instead, can we use `rclpy.qos.QoSPresetProfiles.get_from_short_key(qos_profile_name)` and qos_profile_name is set `default` in default?

ros2service/ros2service/api/__init__.py
```diff
@@ -84,3 +88,33 @@ def __init__(self, *, service_type_key=None):
...
+
+def qos_profile_from_short_keys(
+    preset_profile: str, reliability: str = None, durability: str = None,
+    depth: Optional[int] = None, history: str = None,
+) -> rclpy.qos.QoSProfile:
+    """Construct a QoSProfile given the name of a preset, and optional overrides."""
+    # Build a QoS profile based on user-supplied arguments
+    profile = rclpy.qos.QoSPresetProfiles.get_from_short_key(preset_profile)
+    profile_configure_short_keys(profile, reliability, durability, depth, history)
+    return profile
```
fujitatomoya, line 120: there methods are almost duplication of `ros2topic.api`, can we reuse the already existed methods? probably move those methods to `ros2node.api` the called from else where? if more appropriate location to put, that would be better but i am not really sure...

ros2service/ros2service/verb/call.py
```diff
@@ -49,17 +60,53 @@ def add_arguments(self, parser, cli_name):
...
+            choices=rclpy.qos.QoSReliabilityPolicy.short_keys(),
+            help='Quality of service reliability setting to publish with '
+                 '(overrides reliability value of --qos-profile option, default: {})'
+                 .format(default_profile.reliability.short_key))
+        parser.add_argument(
+            '--qos-durability',
+            choices=rclpy.qos.QoSDurabilityPolicy.short_keys(),
+            help='Quality of service durability setting to publish with '
+                 '(overrides durability value of --qos-profile option, default: {})'
+                 .format(default_profile.durability.short_key))
```
fujitatomoya, line 89: these command line extensions are duplication of `ros2topic.api.add_qos_arguments`. as described above, can we avoid this duplication not to do double maintenance?

fujitatomoya (changes requested):
besides adding test cases in `ros2service.test.test_cli` would be better.

ros2service/ros2service/api/__init__.py
bastianhjaeger: I will have a look into this. Thanks

ros2service/ros2service/verb/call.py
bastianhjaeger: I will have a look into this. Thanks

ros2service/ros2service/verb/call.py
bastianhjaeger: Agreed. I was not sure how to implement this. Do you have some pointer how to solve this?

fujitatomoya:
@bastianhjaeger are you gonna work on this?

fujitatomoya:
@bastianhjaeger thank you for the contribution, i will look into this.

ros2service/package.xml
```diff
@@ -18,6 +18,7 @@
 
   <depend>rclpy</depend>
   <depend>ros2cli</depend>
+  <depend>ros2topic</depend>
```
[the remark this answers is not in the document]
fujitatomoya: besides, adding dependency on `ros2topic` from `ros2service`. how about moving `profile_configure_short_keys` and `qos_profile_from_short_keys` to `ros2cli.node.qos.py` and `ros2topic` and `ros2service` can import from there? In that case, we can avoid additional dependency something like this.

QoS related setting or methods are for node, so i would suggest that approach.

ros2service/ros2service/verb/call.py
```diff
@@ -49,17 +51,23 @@ def add_arguments(self, parser, cli_name):
         parser.add_argument(
             '-r', '--rate', metavar='N', type=float,
             help='Repeat the call at a specific rate in Hz')
+        add_qos_arguments(parser, 'services_default')
```
fujitatomoya, line 54: I think `add_qos_arguments` should be extended to take `service` and other entities, not statically expecting either `publisher` or `subscription`.

fujitatomoya (changes requested):

fujitatomoya:
@bastianhjaeger sorry to be late to get back to you, i think implementation looks good, but could be updated for future maintenance and flexibility. thanks

ros2service/ros2service/verb/call.py
bastianhjaeger: No sure I understand this. Could you elaborate your intention, please?

fujitatomoya: I mean,

`def add_qos_arguments(parser: ArgumentParser, subscribe_or_publish: str, default_profile_str):` to `def add_qos_arguments(parser: ArgumentParser, entity_type: str, default_profile_str):`

so that we can use it any entities such as publisher, subscription, services, actions for future use.

ros2service/package.xml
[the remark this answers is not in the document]
bastianhjaeger: Done.

ros2service/ros2service/verb/call.py
bastianhjaeger: So far I have 
```
def add_qos_arguments(parser: ArgumentParser, default_profile_str: str):
```
and think this is the best way to go, as it is independent on the entity type.

fujitatomoya (changes requested):

fujitatomoya:
> match current

fujitatomoya:
sorry accidentally closed, reopened.

fujitatomoya:
> @fujitatomoya I actually just realized, that the merging of all the QoS arguments and its handling is not so desirable. For the `ros2 topic echo` you actually want the default to match current publishers QoS (reliability and durability). For `ros2 topic pub` it is different. This results in different descriptions in how arguments are handles, thus the arg description is different.
> 
> Moving all the QoS argument handling in one file would result in one file, but the file would contain something like
> 
> ```
> if "topic echo"
>   // automatically adapt reliability/durability to current publisher
> else if "topic pub"
>   //...
> ```
> 
> and this is, in my opinion not desirable. My approach now would be to revert everything for the topic section and add a specific QoS arg handling for `service call`. What do you think?

I am not clear on this. originally this PR is meant to introduce the QoS argument for `ros2 service call`.
but are you saying that we should introduce the new logic to see the most appropriate QoS setting for `ros2 xxx xxx` command individually? e.g https://github.com/ros2/rclcpp/issues/1868

i might be mistaken, could you elaborate it?

fujitatomoya:
@bastianhjaeger are you still working on this?

fujitatomoya:
closing in favor of https://github.com/ros2/ros2cli/pull/966

bastianhjaeger:
> @bastianhjaeger are you still working on this?

Sorry. I needed to drop it. Good to see the work in https://github.com/ros2/ros2cli/pull/966
Thx.

fujitatomoya:
@bastianhjaeger no worries, i fixed your concern as well. if you could review it, that would be really appreciated.

PR #817, opened 2023-04-04
ros2node: Add verb `get_type_description` (rep2011)

achim-k:
Part of https://github.com/ros2/ros2/issues/1159

Adds the verb `get_type_description` which calls the given node's `~/get_type_description` service (if available) and prints out the response. This verb is basically a shorthand for `ros2 service call <node>/get_type_description`

### DIFF
--- a/ros2node/package.xml
+++ b/ros2node/package.xml
@@ -19,6 +19,7 @@
 
   <exec_depend>rclpy</exec_depend>
   <exec_depend>ros2cli</exec_depend>
+  <exec_depend>rosidl_runtime_py</exec_depend>
 
   <test_depend>ament_copyright</test_depend>
   <test_depend>ament_flake8</test_depend>
--- a/ros2node/package.xml
+++ b/ros2node/package.xml
@@ -20,6 +20,7 @@
   <exec_depend>rclpy</exec_depend>
   <exec_depend>ros2cli</exec_depend>
   <exec_depend>rosidl_runtime_py</exec_depend>
+  <exec_depend>type_description_interfaces</exec_depend>
 
   <test_depend>ament_copyright</test_depend>
   <test_depend>ament_flake8</test_depend>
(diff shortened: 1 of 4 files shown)


ros2node/ros2node/verb/get_type_description.py
```diff
@@ -0,0 +1,64 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import rclpy
+from ros2cli.node import NODE_NAME_PREFIX
+from ros2node.api import NodeNameCompleter
+from ros2node.verb import VerbExtension
+from rosidl_runtime_py import message_to_yaml
+from type_description_interfaces.srv import GetTypeDescription
```
achim-k, line 20: missing a dependency in package.xml

ros2node/package.xml
```diff
@@ -20,6 +20,7 @@
   <depend>ros2cli</depend>
```
clalancette, line 20:
```suggestion
  <exec_depend>ros2cli</exec_depend>
```

ros2node/package.xml
```diff
@@ -20,6 +20,7 @@
...
 
   <exec_depend>rclpy</exec_depend>
+  <exec_depend>rosidl_runtime_py</exec_depend>
```
clalancette, line 23:
```suggestion
  <exec_depend>rosidl_runtime_py</exec_depend>
  <exec_depend>type_description_interfaces</exec_depend>
```

clalancette (dismissed):

ros2node/package.xml
achim-k: I believe `<depend>` is correct here. All other ros2* packages have it as well, I guess the main reason is to register the command verbs with `ros2cli` (setup.py)?

clalancette: It's not necessary for that.  There is never a reason to use `<depend>` in a pure Python package; it has no build step.

achim-k: OK I see. Is it OK for you to leave it as is, since the change would be somewhat unrelated to this PR?

clalancette (commented):
This generally looks good to me, but I'm going to hold off on approving until we finish https://github.com/ros2/rcl/pull/1052 .

achim-k:
Closing, this will be implemented in https://github.com/ros2/ros2cli/pull/846

Issue #725, opened 2022-06-16
ros2 topic echo: No option to clear screen


ahemwe:
#### Feature description

In ROS 1 `rostopic echo` has an option "-c"
`$ rostopic echo -c /topic_name`

Meaning:
_Clear the screen after each message is published. Cannot be used with -p._

See: http://wiki.ros.org/rostopic#rostopic_echo

This is very helpful when using a terminal session to monitor ros messages in a terminal.

In ROS 2 this option is not available as of now. I would highly appreciate if this could be implemented.

torrhen:
Would require something like the `"\033[2J\033[;H"` ANSI control sequence prefixed for each message callback to clear display and reposition cursor.

No `-p` argument in this case, but `--csv` looks similar. Are they equivalent given this context?

PR #819, opened 2023-04-11
feat(echo --clear): add --clear option to echo

Guillaumebeuzeboc:
Adds the `--clear/-c` option to echo.

Solves this issue: https://github.com/ros2/ros2cli/issues/725

This was previously available in ROS.

This clears the screen before a new message is displayed.

`ESC[2J` corresponds to `erase entire screen`
`ESC[H` corresponds to `moves cursor to home position (0, 0)`

I had to build rolling from source to successfully run the tests

### DIFF
--- a/ros2topic/ros2topic/verb/echo.py
+++ b/ros2topic/ros2topic/verb/echo.py
@@ -12,9 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import os
-import platform
-
 from typing import Optional
 from typing import TypeVar
 
@@ -336,7 +333,4 @@ def _message_lost_event_callback(message_lost_status):
 
 
 def clear_terminal():
-    if platform.system() == 'Windows':
-        os.system('cls')
-    else:
-        os.system('clear')
+    print('\x1b[H\x1b[2J')
--- a/ros2topic/ros2topic/verb/echo.py
+++ b/ros2topic/ros2topic/verb/echo.py
@@ -12,6 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import os
+import platform
+
 from typing import Optional
 from typing import TypeVar
 
@@ -109,7 +112,7 @@ class EchoVerb(VerbExtension):
             help='Shows the associated message info.')
         parser.add_argument(
             '--clear', '-c', action='store_true',
-            help='clear screen before printing next message')
+            help='Clear screen before printing next message')
 
     def choose_qos(self, node, args):
 
@@ -188,7 +191,7 @@ class EchoVerb(VerbExtension):
         self.no_str = args.no_str
         self.flow_style = args.flow_style
         self.once = args.once
-        self.prefix = '\033[2J\033[H' if args.clear else ''
+        self.clear_screen = args.clear
 
         self.filter_fn = None
         if args.filter_expr:
@@ -284,15 +287,17 @@ class EchoVerb(VerbExtension):
         if self.future is not None and self.once:
             self.future.set_result(True)
 
-        to_print = ''
+        # Clear terminal screen before print
+        if self.clear_screen:
+            clear_terminal()
 
         if not hasattr(submsg, '__slots__'):
             # raw
             if self.include_message_info:
-                to_print = '---Got new message, message info:---\n'
-                to_print = f'{to_print}{info}\n'
-                to_print = f'{to_print}---Message data:---\n'
-            print(f'{self.prefix}{to_print}{submsg}', end='\n---\n')
+                print('---Got new message, message info:---')
+                print(info)
+                print('---Message data:---')
+            print(submsg, end='\n---\n')
             return
 
         if self.csv:
@@ -303,15 +308,16 @@ class EchoVerb(VerbExtension):
                 no_str=self.no_str)
             if self.include_message_info:
                 to_print = f'{",".join(str(x) for x in info.values())},{to_print}'
-            print(f'{self.prefix}{to_print}')
+            print(to_print)
             return
         # yaml
         if self.include_message_info:
-            to_print = f'{yaml.dump(info)}---\n'
-        to_print += message_to_yaml(
+            print(yaml.dump(info), end='---\n')
+        print(
+            message_to_yaml(
                 submsg, truncate_length=self.truncate_length,
-                no_arr=self.no_arr, no_str=self.no_str, flow_style=self.flow_style)
-        print(f'{self.prefix}{to_print}---\n')
+                no_arr=self.no_arr, no_str=self.no_str, flow_style=self.flow_style),
+            end='---\n')
 
 
 def _expr_eval(expr):
@@ -327,3 +333,10 @@ def _message_lost_event_callback(message_lost_status):
         f'\n\ttotal count: {message_lost_status.total_count}',
         end='---\n'
     )
+
+
+def clear_terminal():
+    if platform.system() == 'Windows':
+        os.system('cls')
+    else:
+        os.system('clear')
--- a/ros2topic/ros2topic/verb/echo.py
+++ b/ros2topic/ros2topic/verb/echo.py
@@ -107,6 +107,9 @@ class EchoVerb(VerbExtension):
         parser.add_argument(
             '--include-message-info', '-i', action='store_true',
             help='Shows the associated message info.')
+        parser.add_argument(
+            '--clear', '-c', action='store_true',
+            help='clear screen before printing next message')
 
     def choose_qos(self, node, args):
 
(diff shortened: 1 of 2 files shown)


ros2topic/ros2topic/verb/echo.py
```diff
@@ -288,13 +292,15 @@ def _subscriber_callback(self, msg, info):
...
             # raw
             if self.include_message_info:
-                print('---Got new message, message info:---')
-                print(info)
-                print('---Message data:---')
-            print(submsg, end='\n---\n')
+                to_print = '---Got new message, message info:---\n'
+                to_print = f'{to_print}{info}\n'
+                to_print = f'{to_print}---Message data:---\n'
+            print(f'{self.prefix}{to_print}{submsg}', end='\n---\n')
```
fujitatomoya, line 303: having escape sequence was what we do in ROS 1 rostopic echo, so this would be probably good.

instead, how about taking advantage of `os.system()` to clear the screen? so that we can just rely on system shell that gives us more flexibility for the future?

```python
import os
import system

if platform.system() == 'Windows':
    os.system('cls')
else:
    os.system('clear')
```

fujitatomoya (commented):
implementation looks good to me but many tests are failing.

could you address the following failures?

https://build.ros2.org/job/Rpr__ros2cli__ubuntu_jammy_amd64/238/testReport/ros2topic.ros2topic.test/test_cli/test_cli/

https://build.ros2.org/job/Rpr__ros2cli__ubuntu_jammy_amd64/238/testReport/ros2topic.ros2topic.test/test_echo_pub/test_echo_pub/

ros2topic/ros2topic/verb/echo.py
clalancette: I agree that the escape sequence is potentially problematic, particularly on Windows.

That said, I'm really not to enamored of doing `os.system`.  That almost always leads to trouble down the line, as it is difficult to deal with errors and easy to trick it into doing something nefarious.

I think it is worthwhile here to step back and maybe look at some of the python command-line libraries and see what they do in this situation.  If we can find a solution to work on all shells and in Windows (both `cmd` and `powershell`), and does not require `os.system`, that would be the ideal.

ahemwe:
Does that PR require further changes?

clalancette:
> Does that PR require further changes?

Yes, we need to come to a conclusion on this [discussion](https://github.com/ros2/ros2cli/pull/819#discussion_r1165043306).

ros2topic/ros2topic/verb/echo.py
VArt3m: I think this is a very convenient feature and I would love to see this discussion and PR resolved and merged. Can provide help if it's wanted.
For example, very popular [click](https://github.com/pallets/click/blob/d73083e3960567429428578859d7f7c8fbe0e0a7/src/click/termui.py#L450) python library uses `\033[2J\033[1;1H` escape sequence to clear the screen and reset the cursor position.

fujitatomoya: @Guillaumebeuzeboc @clalancette i just pushed the fix based on the review comment here, can you check?

fujitatomoya:
Pulls: ros2/ros2cli#819
Gist: https://gist.githubusercontent.com/fujitatomoya/c7569c301740a3cb51951715927637b3/raw/ece81984450ed3dd68f812e402c55ef6353b55d4/ros2.repos
BUILD args:  --packages-above-and-dependencies ros2topic
TEST args:  --packages-above ros2topic
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14693
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21981/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16246/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1495/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22711/)

ros2topic/ros2topic/verb/echo.py
clalancette: I'm still not a fan of the use of `os.system` in here.  I'd much rather we figure out the correct terminal codes to do this, if we can.

fujitatomoya:
Okay the CI is failing with the following error. (my local environment which does have the terminal environment, there is no problem)

https://ci.ros2.org/job/ci_linux/21981/testReport/junit/ros2topic.ros2topic.test/test_cli/test_cli/

```bash
  File "/home/jenkins-agent/workspace/ci_linux/ws/install/launch_testing/lib/python3.12/site-packages/launch_testing/markers.py", line 61, in _wrapper
    return func(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jenkins-agent/workspace/ci_linux/ws/src/Guillaumebeuzeboc/ros2cli/ros2topic/test/test_cli.py", line 411, in test_topic_echo_clear
    assert topic_command.wait_for_output(functools.partial(
AssertionError: Output does not match: 'unknown': unknown terminal type.
  data: 'Hello World: 78'
  ---
  'unknown': unknown terminal type.
  data: 'Hello World: 79'
  ---
```

I think we need to check if the terminal environmental variable is there with checking `TERM` for unix system. but i am not sure how we can do that with Windows... any thoughts?

ros2topic/ros2topic/verb/echo.py
Guillaumebeuzeboc: The "rich" python library is also using [the escape sequence to clear the screen](https://github.com/Textualize/rich/blob/0f2f51b872d14588de3a65968f5cdde6fb5694a3/rich/control.py#L38).

@clalancette what do you mean by "terminal codes"?
It seems that `os.system` and escape characters are the only solutions that are used around.

Guillaumebeuzeboc: The [Wikipedia page about ANSI escape code](https://en.wikipedia.org/wiki/ANSI_escape_code#DOS_and_Windows) mentions that escape characters in strings are only supported since PowerShell6 (the default PowerShell for windows10 &11 is 5.1).
I installed power shell 7 and I can confirm that the string ``"`e[2J`e[H"`` is clearing the screen. As you can see, [the escape character is different on Windows](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_special_characters?view=powershell-7.4#escape-e). Also, none of that is going to work natively on Windows since the version of PowerShell is too old.
From here I am really not sure what can be done.

VArt3m: image
Escape sequence in click library (in `clear()` function) worked perfectly in PowerShell 5.1 for me on Windows 11. Just doing `print("\033[2J\033[1;1H")` works too. So I don't think it's a problem? I think that's an escape character for _strings_ that contain ANSI escape sequences, so you could construct such strings in PowerShell commands. But we don't write PowerShell commands here :)
As Wikipedia article you linked mentions the support for ANSI escape sequences was added and eabled by default since PowerShell 5.1, so I think it's gonna be alright.

ros2topic/ros2topic/verb/echo.py
```diff
@@ -322,3 +333,10 @@ def _message_lost_event_callback(message_lost_status):
         f'\n\ttotal count: {message_lost_status.total_count}',
         end='---\n'
     )
+
+
+def clear_terminal():
+    if platform.system() == 'Windows':
+        os.system('cls')
+    else:
+        os.system('clear')
```
clalancette, line 342: As far as I can tell from some local testing, we should be able to replace this with an ANSI escape sequence (tested to work on bash, tcsh, zsh, fish, Windows command prompt, and Windows Powershell):

```suggestion
def clear_terminal():
    print('\x1b[H\x1b[2J')
```

Can others try it locally and see if it works for them?

clalancette (changes requested):

fujitatomoya:
I ended up not having the test case for this because of https://github.com/ros2/ros2cli/pull/819#issuecomment-2407719260 and checking the escape sequence generates unterminated error during check.

fujitatomoya:
Pulls: ros2/ros2cli#819
Gist: https://gist.githubusercontent.com/fujitatomoya/e3ceab25625387dd621a69ef49e4acc2/raw/ece81984450ed3dd68f812e402c55ef6353b55d4/ros2.repos
BUILD args:  --packages-above-and-dependencies ros2topic
TEST args:  --packages-above ros2topic
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14770
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22074/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16331/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1593/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22809/)

PR #823, opened 2023-04-14
Make all of the dependencies in pure Python packages exec_depend.

clalancette:
That is, they have no build step, so they shouldn't export a dependency for that.

@cottsay can you double-check my work on this one and make sure I'm not mistaken?  Thanks.


cottsay (approved):
Looks good, thanks!

PR #846, opened 2023-07-28
Add ros2 node get_type_description initial implementation

emersonknapp:
Implements a first pass at `ros2 node get_type_description`. It does its best to autofill the type hash so the user doesn't have to, but given the limitations of the current API this is only possible for `msg` on pub/sub topics, not for srv/action. For those, the hash must be provided manually by the user, though it is not discoverable via any existing API.

I think we should move forward with putting this in, as it covers the majority use case. Further discussion follows:

1. It's awkward that we need to get `names_and_types`, then use that to go to `get_x_info_by_topic` in order to get `TopicEndpointInfo` to find the hash. https://github.com/ros2/rmw/issues/356 would make this part easier and support srv/action
2. The fact that we need to do this graph-munging to "discover" the hash for a typename seems unnecessary. Given that a specific node has a 1:1 relationship of type name and type hash, it seems like `GetTypeDescription` should allow for either (but not both) to be empty in the request, so that the requesting client doesn't necessarily even need to know both pieces of information. From a user point of view, it seems very natural to say "you node, give me your description for mypackage/msg/Foo". It also seems like it could come up, though less commonly, to request "you node, what is the type that has hash RIHS01_e28ce254ca8abc06abf92773b74602cdbf116ed34fbaf294fb9f81da9f318eac". The current underlying implementation in `rcl` assumes the hash is the more important piece of information, using it as the key to its cache, but this usage reveals that it's probably more likely we wish to fetch this information by typename, using the hash as a potential validation. HOWEVER - all that aside, with https://github.com/ros2/rmw/issues/356 the "discovery" would be much simpler and so alleviate some of my concern with this point.


emersonknapp:
@clalancette could I get a reviewer assignment for this? (or @audrow @gbiggs friendly ping)

ros2node/ros2node/verb/get_type_description.py
```diff
@@ -0,0 +1,156 @@
...
+        parser.add_argument(
+            '--include-sources', action='store_true',
+            help='Fetch and display the raw source text as well')
+
+    def main(self, *, args):
+        service_name = f'{args.node_name}/get_type_description'
+
+        rclpy.init()
+        node = rclpy.create_node(f'{NODE_NAME_PREFIX}_td_requester_{os.getpid()}')
+        cli = node.create_client(GetTypeDescription, '/talker/get_type_description')
```
fujitatomoya, line 131:
```suggestion
        cli = node.create_client(GetTypeDescription, service_name)
```

ros2node/ros2node/verb/get_type_description.py
```diff
@@ -0,0 +1,156 @@
...
+        if not cli.service_is_ready():
+            print('Waiting for service to become available...')
+            if not cli.wait_for_service(timeout_sec=5.0):
+                raise RuntimeError(f'Service {service_name} not found')
```
fujitatomoya, line 135: how about waiting until service is ready, unless user sends the signal to stop? i would have timeout option instead of static timeout that user cannot even see how long it waits.

```suggestion
            cli.wait_for_service()
```

fujitatomoya (commented):
it would be nice to have test for topic.

ros2node/ros2node/verb/get_type_description.py
```diff
@@ -0,0 +1,156 @@
...
+from ros2node.api import (
+    NodeNameCompleter,
+    parse_node_name,
+)
+from ros2node.verb import VerbExtension
+from type_description_interfaces.msg import FieldType
+from type_description_interfaces.srv import GetTypeDescription
+
+
+def print_field_type(ft):
```
MichaelOrlov, line 29: The function name `print_field_type` is a bit misleading since it doesn't print anything.
IMO it would be more appropriate to rename it to the `get_field_type_str` or something similar.

ros2node/ros2node/verb/get_type_description.py
```diff
@@ -0,0 +1,156 @@
...
+    if not type_name:
+        raise RuntimeError(f'Could not match FieldType type_id {ft.type_id} to a name')
+
+    type_str = type_name
+    if 'NESTED_TYPE' in type_name:
+        type_str += f' ({ft.nested_type_name})'
+    if ft.string_capacity:
+        type_str += f' - string_capacity {ft.string_capacity}'
+    if ft.capacity:
+        type_str += f' - capacity {ft.capacity}'
```
MichaelOrlov, line 45:
```suggestion
    if ft.string_capacity:
        type_str += f' string_capacity({ft.string_capacity})'
    if ft.capacity:
        type_str += f' capacity({ft.capacity})'
```
IMO formatting with `- capacity 255` could be a bit confusing especially if afterward will be printed out the dafult value. It is gonna look like `FIELD_TYPE_INT8_ARRAY - capacity 25 = 0`

ros2node/ros2node/verb/get_type_description.py
```diff
@@ -0,0 +1,156 @@
...
+        print(' ' * indent + line)
+
+    print()
+    p(f'{itd.type_name}')
+    p('Fields:')
+    indent += 2
+    for field in itd.fields:
+        field_string = f'{field.name}: {print_field_type(field.type)}'
+        if field.default_value:
+            field_string += f' = {field.default_value}'
```
MichaelOrlov, line 60:
```suggestion
        if field.default_value:
            field_string += f' default({field.default_value})'
```

ros2node/ros2node/verb/get_type_description.py
```diff
@@ -0,0 +1,156 @@
...
+        p(field_string)
+    indent -= 2
+
+
+def print_type_description(td):
+    print_individual_type_description(td.type_description)
+    print()
+    print('Referenced Type Descriptions:')
+    for rtd in td.referenced_type_descriptions:
+        print_individual_type_description(rtd, indent=2)
```
MichaelOrlov, line 70: Printing `'Referenced Type Descriptions:'` make sense only if `td.referenced_type_descriptions` is not empty

ros2node/ros2node/verb/get_type_description.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+def discover_hash_for_type(node, remote_node_name, type_name):
+    # TODO(emersonknapp) get hashes from names_and_types when that is implemented
```
MichaelOrlov, line 74: Please add a reference to the tracking issue https://github.com/ros2/rmw/issues/356 in the TODO comment.

MichaelOrlov (changes requested):
@emersonknapp In general looks good to me.  
I only have a few minor concerns, especially about formatting output
>  formatting with `- capacity 255` could be a bit confusing especially if the afterward will be printed out the default value. It is gonna look like FIELD_TYPE_INT8_ARRAY - capacity 25 = 0

I agree that if we had https://github.com/ros2/rmw/issues/356 implemented it would significantly simplify the search for hashes.  
I also thinking about why we require to provide a hash when doing a service request for `get_type_description`?  
What if make the hash field optional and the service will return a list of types with hashes?  
It will be a client-side responsibility to verify hashes and determine which type is really interesting. 
IMO this way everybody's life would be much easier :smiley: 

Update: @emersonknapp  it would be nice to have some tests.

amparo-siemens:
I need the hash to obtain the information of an interface in runtime using the GetTypeDescription service. With the type hash discovery for msg is really easy since just executing the "ros2 topic info -v /topic" I obtain the required hash to call the service. But afaik there is no type hash discovery for actions and services. Because of that, to use the GetTypeDescription I need to check the hash manually in the install folder, so it is only possible if you have access to the source code. Is there any plan to include the type hash discovery for actions and services to be able to take the advantage of knowing the structure of the interfaces in runtime (without access to the code) using the GetTypeDescription service?

fujitatomoya:
> Is there any plan to include the type hash discovery for actions and services to be able to take the advantage of knowing the structure of the interfaces in runtime (without access to the code) using the GetTypeDescription service?

AFAIK, nobody is currently working on this and neither the plan. https://github.com/ros2/ros2/issues/1159 does not include this requirement, but there is the issue for that https://github.com/ros2/rmw/issues/356

Issue #849, opened 2023-08-09
Command to monitor and pretty print the rosout logs


guillaumeautran:
## Enhancement Request

**Required Info:**

- Operating System:
  - OS Agnostic
- Installation type:
  - N/A
- Version or commit hash:
  - N/A
- DDS implementation:
  - N/A
- Client library (if applicable):
  - rclpy
#### Feature description
Monitoring the rosout logs of a running system from the command line involves running the `ros2 topic echo /rosout` and dealing with the content of the message AS-IS. 
```
$ ros2 topic echo /rosout
---
stamp:
  sec: 1691614804
  nanosec: 921514406
level: 20
name: rosbag2_player
msg: Set rate to 2.35795
file: /build/source/src/rosbag2_transport/player.cpp
function: set_rate
line: 295
---
```
Although this works, the results are less than ideal and confusing to read for most users. Doing grep to search the output is made difficult as well and filtering the content for the logs of a specific node is also difficult.

Instead, it would be great to be able to pretty print the rosout logs in a way that makes it appealing and easy to follow. 
For example: 
```
 $ ros2 rosout print
[2023-08-09 17:02:22.539865596] [INFO ] [rosbag2_player]: Set rate to 2.59374
```

ros-discourse:
This issue has been mentioned on **Open Robotics Discourse**. There might be relevant details there:

https://discourse.openrobotics.org/t/draft-rep-0156-2026-ros2log-command-line-interface-call-for-discussion/51913/1

PR #850, opened 2023-08-09
Command to monitor and pretty print the rosout logs #849

guillaumeautran:
Creation of a rosout print command that generates pretty prints of the content of the `/rosout` topic. The commands also allows for the filtering of nodes based on a regular expression applied to the node name itself. Output color can be disabled (enabled by default) and the function information detail can also be displayed.

issue: #849

### DIFF
--- a/ros2rosout/README.md
+++ b/ros2rosout/README.md
@@ -1,0 +1,19 @@
+# ros2rosout
+
+This is the `ros2 rosout print` utility command which displaies the content of `/rosout` topic in a nicely formatted and colorized output.
+
+## Usage 
+
+Run `ros2 rosout print` to get the live stream of the logs.
+
+Run `ros2 rosout print -h/--help` to print all available command arguments.
+
+
+## Output format
+
+The command outputs the log line with the following format: 
+`[` _datetime_ `] [` _level_ `] [` _nodename_ `]: ` _log_ `:[` _file_ `:` _line_ `(` _function_ `)]` 
+
+File, line and function are all optional and are displays when the `-f/--function-detail` switch is provided. 
+
+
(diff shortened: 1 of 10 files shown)


clalancette:
I'm honestly not sure this is a feature we need to have in `ros2` itself.

You can easily parse this kind of information using a YAML parser.  For instance, I downloaded https://github.com/mikefarah/yq , and now I can do:

```
$ ros2 topic echo /rosout | yq '.msg'
```

Which just prints out the message field.  You can also easily use something like it to select only the nodes you are interested in:

```
$ ros2 topic echo /rosout | yq 'select(.name == "talker") .msg'
```

So given that there are other tools out there that can do this, I'd rather not maintain this in the core itself.

guillaumeautran:
That's fair. Just thought I'd bring it here to see if there was any interest.

clalancette:
Sorry, I was maybe a little more terse than I meant to be.

Everything we put into the ROS 2 core has a cost, mostly in maintenance.  That stuff always has to work before we can do a release, so keeping the size of the core smaller means it is somewhat easier to maintain.  That's the main reason I always look for other tools to do things, rather than write our own.

However, the good news is that we've attempted to make ROS 2 as pluggable as possible.  So if you find value in this extension, then I'd encourage you to put this in a separate repository and release it into Humble, Iron, and Rolling.  That neatly sidesteps the problem of having it in the core, while still making the functionality available.  Over time if we then find that this is a must-have feature, we can then promote it into the core.

Does that make sense?

guillaumeautran:
I do understand the need to keep the core lean to reduce the maintenance cost. And although you are correct that you can use other tools to achieve very similar functionality, I do see the benefit of such command (and a similar one for the diagnostics topic as well).  

Anyway, yes it is my plan to create a separate repo for these type of tools. Here at Clearpath, we use this utility all the time to follow the logs live. Will post the link to the repo once it is up. 

Thanks for your feedback.

guillaumeautran:
FYI: https://github.com/clearpathrobotics/clearpath_ros2cli
More to come :)

PR #854, opened 2023-09-22
catch ExternalShutdownException ros2cli main.

fujitatomoya:
address https://github.com/ros2/rclpy/issues/1166

### DIFF
--- a/ros2cli/ros2cli/cli.py
+++ b/ros2cli/ros2cli/cli.py
@@ -19,6 +19,8 @@ import functools
 import signal
 import sys
 
+from rclpy.executors import ExternalShutdownException
+
 from ros2cli.command import add_subparsers_on_demand
 
 
@@ -89,6 +91,8 @@ def main(*, script_name='ros2', argv=None, description=None, extension=None):
         rc = extension.main(parser=parser, args=args)
     except KeyboardInterrupt:
         rc = signal.SIGINT
+    except ExternalShutdownException:
+        rc = signal.SIGTERM
     except RuntimeError as e:
         rc = str(e)
     return rc


Yadunund (approved):
LGTM

ionutnechita-intel (commented):
With this change i have this at the end of the execution:

failed to initialize wait set: the given context is not valid, either rcl_init() was not called or rcl_shutdown() was called., at ./src/rcl/wait.c:130

timeout 5s ros2 topic hz /camera/color/image_raw

ros2cli/ros2cli/cli.py
```diff
@@ -89,6 +91,8 @@ def main(*, script_name='ros2', argv=None, description=None, extension=None):
         rc = extension.main(parser=parser, args=args)
     except KeyboardInterrupt:
         rc = signal.SIGINT
+    except ExternalShutdownException:
+        rc = signal.SIGTERM
     except RuntimeError as e:
         rc = str(e)
```
fujitatomoya, line 97: > failed to initialize wait set: the given context is not valid, either rcl_init() was not called or rcl_shutdown() was called., at ./src/rcl/wait.c:130

that is caught on here as RuntimeError. i think that this is expected exception since deferred signal handler shuts down the context asynchronously, and if that is racy condition to create `WaitSet::WaitSet`, this exception is generated in `rcl_wait_set_init`.

i am not sure what else we could d here, any thoughts?

iuhilnehc-ynos: After referring to `rclcpp`, it seems that we need to make the https://github.com/ros2/rclpy/blob/159ced49bb904511154d9f6e595b64e1f5c0d8c0/rclpy/rclpy/executors.py#L643 as a data 
member of `Executor`, and then call `rcl_wait_set_resize` instead, which means the https://github.com/ros2/rclpy/blob/1843c2259260787d94975f658ed195a82f6eb90c/rclpy/src/rclpy/wait_set.cpp#L255 need to export the `resize` method for `rcl_wait_set_resize`.

fujitatomoya: @iuhilnehc-ynos thanks for pointing that out, that can be another enhancement.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19588/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14099/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20312/)

fujitatomoya:
linux unstable failures are unrelated, i will go ahead to merge this.

fujitatomoya:
@Yadunund thanks for the review.

fujitatomoya:
@mergifyio backport iron humble

Issue #806, opened 2023-02-27
Ros2 doctor fails to call some class functions 


JvE-Lely:
## Bug report

**Required Info:**

- Operating System:
  - Debian buster
- Installation type:
  - source
- Version or commit hash:
  - Humble 
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  -  N/A

#### Steps to reproduce issue

```
ros2 doctor --report
```

#### Actual behavior
```
ros2 doctor --report
/opt/ros/humble/lib/python3.7/site-packages/ros2doctor/api/__init__.py: 154: UserWarning: Fail to call PackageReport class functions.
/opt/ros/humble/lib/python3.7/site-packages/ros2doctor/api/__init__.py: 154: UserWarning: Fail to call RosdistroReport class functions.
```
and
```
ros2 doctor
/opt/ros/humble/lib/python3.7/site-packages/ros2doctor/api/__init__.py: 118: UserWarning: Fail to call PackageCheck class functions.
/opt/ros/humble/lib/python3.7/site-packages/ros2doctor/api/__init__.py: 118: UserWarning: Fail to call PlatformCheck class functions.
```

#### Additional information
The following packages are installed:
```
 python3-catkin-pkg-modules_0.5.2-1_all.deb
 python3-lark-parser_0.7.2-1osrf~buster_all.deb
 python3-netifaces_0.10.4-1+b1_armhf.deb
 python3-numpy_1.16.2-1_armhf.deb
 python3-rosdistro_0.9.0-100_all.deb
 python3-rosdistro-modules_0.9.0-1_all.deb
 python3-rospkg_1.4.0-100_all.deb
 python3-rospkg-modules_1.4.0-1_all.deb
```

See also https://answers.ros.org/question/406296/optroshumblelibpython310site-packagesros2doctorapi__init__py-118-userwarning-fail-to-call-packagecheck-class-functions/

mjcarroll:
I cannot reproduce on jammy, so it may be something `buster` specific in this case.  I will try to spin a docker container to reproduce.

zhuzy-2018:
i've encoutered this issue as well using humble on ubuntu 22.04.

fujitatomoya:
I cannot reproduce this issue with ubuntu 22.04 and Humble.

```bash
root@tomoyafujita:~# ros2 doctor

All 5 checks passed

root@tomoyafujita:~# ros2 doctor --report
...<snip>
   PLATFORM INFORMATION
[...]

PR #860, opened 2023-10-09
(ros2doctor) fix PackageCheck

MatthijsBurgh:
Fixes https://github.com/ros2/ros2cli/issues/806

Maybe backport to humble and iron too.

### DIFF
--- a/ros2doctor/ros2doctor/api/package.py
+++ b/ros2doctor/ros2doctor/api/package.py
@@ -97,15 +97,15 @@ def compare_versions(result: Result, local_packages: dict, distro_packages: dict
     :param: boolean value determines which output to populate, msgs or report
     :return: list of warning messages
     """
-    missing_req = ''
-    missing_local = ''
+    missing_req = []
+    missing_local = []
     for name, local_ver_str in local_packages.items():
         if not local_ver_str:
-            missing_local += ' ' + name
+            missing_local += name
             local_ver_str = ''
         latest_ver_str = distro_packages.get(name, '')
         if not latest_ver_str:
-            missing_req += ' ' + name
+            missing_req += name
         local_ver = version.parse(local_ver_str).base_version
         latest_ver = version.parse(latest_ver_str).base_version
         if local_ver < latest_ver:
@@ -114,26 +114,29 @@ def compare_versions(result: Result, local_packages: dict, distro_packages: dict
                 f' local: {local_ver} <'
                 f' latest: {latest_ver}')
             result.add_warning()
-    if missing_req:
-        if len(missing_req) > 100:
+
+    missing_req_str = ' '.join(missing_req)
+    if missing_req_str:
+        if len(missing_req_str) > 100:
             doctor_warn(
                 'Cannot find the latest versions of packages: ' +
-                textwrap.shorten(missing_req, width=100) +
+                textwrap.shorten(missing_req_str, width=100) +
                 ' Use `ros2 doctor --report` to see full list.')
         else:
             doctor_warn(
                 'Cannot find the latest versions of packages: ' +
-                missing_req)
-    if missing_local:
-        if len(missing_local) > 100:
+                missing_req_str)
+    missing_local_str = ' '.join(missing_local)
+    if missing_local_str:
+        if len(missing_local_str) > 100:
             doctor_warn(
                 'Cannot find local versions of packages: ' +
-                textwrap.shorten(missing_local, width=100) +
+                textwrap.shorten(missing_local_str, width=100) +
                 ' Use `ros2 doctor --report` to see full list.')
         else:
             doctor_warn(
                 'Cannot find local versions of packages: ' +
-                missing_local)
+                missing_local_str)
 
 
 class PackageCheck(DoctorCheck):
--- a/ros2doctor/ros2doctor/api/package.py
+++ b/ros2doctor/ros2doctor/api/package.py
@@ -101,17 +101,22 @@ def compare_versions(result: Result, local_packages: dict, distro_packages: dict
     missing_local = []
     for name, local_ver_str in local_packages.items():
         if not local_ver_str:
-            missing_local += name
-            local_ver_str = ''
+            missing_local.append(name)
         latest_ver_str = distro_packages.get(name, '')
         if not latest_ver_str:
-            missing_req += name
+            missing_req.append(name)
+
+        if not local_ver_str or not latest_ver_str:
+            continue  # Logging is already done by missing_local/missing_req
 
         try:
             local_ver = version.parse(local_ver_str).base_version
+        except version.InvalidVersion:
+            doctor_warn(f"Failed to parse local version '{local_ver_str}' of package '{name}'")
+        try:
             latest_ver = version.parse(latest_ver_str).base_version
         except version.InvalidVersion:
-            continue  # Logging is already done by missing_local/missing_req
+            doctor_warn(f"Failed to parse latest version '{latest_ver_str}' of package '{name}'")
 
         if local_ver < latest_ver:
             doctor_warn(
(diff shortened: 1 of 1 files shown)


MatthijsBurgh:
@mjcarroll you assigned https://github.com/ros2/ros2cli/issues/806 to yourself. So maybe you want to review this too.

ros2doctor/ros2doctor/api/package.py
```diff
@@ -97,43 +97,51 @@ def compare_versions(result: Result, local_packages: dict, distro_packages: dict
...
     :return: list of warning messages
     """
-    missing_req = ''
-    missing_local = ''
+    missing_req = []
+    missing_local = []
     for name, local_ver_str in local_packages.items():
         if not local_ver_str:
-            missing_local += ' ' + name
+            missing_local += name
```
clalancette, line 104: This should be:

```suggestion
            missing_local.append(name)
```

ros2doctor/ros2doctor/api/package.py
```diff
@@ -97,43 +97,51 @@ def compare_versions(result: Result, local_packages: dict, distro_packages: dict
...
             local_ver_str = ''
```
clalancette, line 105: I know you didn't add this, but there doesn't seem to be a reason for this line of code.  We already know that the local_ver_str is empty by this point.

```suggestion
```

ros2doctor/ros2doctor/api/package.py
```diff
@@ -97,43 +97,51 @@ def compare_versions(result: Result, local_packages: dict, distro_packages: dict
...
         latest_ver_str = distro_packages.get(name, '')
         if not latest_ver_str:
-            missing_req += ' ' + name
-        local_ver = version.parse(local_ver_str).base_version
-        latest_ver = version.parse(latest_ver_str).base_version
+            missing_req += name
```
clalancette, line 108:
```suggestion
            missing_req.append(name)
```

ros2doctor/ros2doctor/api/package.py
```diff
@@ -97,43 +97,51 @@ def compare_versions(result: Result, local_packages: dict, distro_packages: dict
...
+
+        try:
```
clalancette, line 110: I like having this `try..except` block to deal with arbitrary version parsing errors.

However, if either `local_ver_str` or `latest_ver_str` is empty, there is really no reason to bother here at all; we know it is going to fail.  I think we should just skip the version parsing and comparison in that case.

ros2doctor/ros2doctor/api/package.py
```diff
@@ -97,43 +97,51 @@ def compare_versions(result: Result, local_packages: dict, distro_packages: dict
...
-                textwrap.shorten(missing_req, width=100) +
+                textwrap.shorten(missing_req_str, width=100) +
                 ' Use `ros2 doctor --report` to see full list.')
         else:
             doctor_warn(
                 'Cannot find the latest versions of packages: ' +
-                missing_req)
-    if missing_local:
-        if len(missing_local) > 100:
+                missing_req_str)
```
clalancette, line 133: Again, I know you didn't add all of this, but there doesn't seem to be a reason to have two branches here for <= 100 and > 100.  We can just unconditionally use `textwrap.shorten` here (and I think it is reasonable to print the 'Use `ros2 doctor --report` to see full list.' in both cases).

clalancette (changes requested):
Thanks for bringing this up; this is definitely an improvement.  I've left some stuff inline.

Besides these changes, I think we need another change in the `PackageReport` class down below to completely fix #806.  In particular, if the `latest_ver` is empty, we have a similar problem to what you are fixing here so we should skip the `version.parse` there as well.

ros2doctor/ros2doctor/api/package.py
```diff
@@ -97,15 +97,18 @@ def compare_versions(result: Result, local_packages: dict, distro_packages: dict
...
         latest_ver_str = distro_packages.get(name, '')
         if not latest_ver_str:
-            missing_req += ' ' + name
+            missing_req.append(name)
+
+        if not local_ver_str or not latest_ver_str:
+            continue  # Logging is already done by missing_local/missing_req
+
         local_ver = version.parse(local_ver_str).base_version
         latest_ver = version.parse(latest_ver_str).base_version
```
clalancette, line 113: Sorry, I wasn't clear on my previous comment.  I think we should preserve the `try..except` block here so that *other* errors in version parsing are caught and a warning printed.

```suggestion
        try:
            local_ver = version.parse(local_ver_str).base_version
        except version.InvalidVersion:
            doctor_warn(f'Failed to parse local version '{local_ver_str}' of package '{name}')
        try:
            latest_ver = version.parse(latest_ver_str).base_version
        except version.InvalidVersion:
            doctor_warn(f'Failed to parse latest version '{latest_ver_str}' of package '{name}')
```

ros2doctor/ros2doctor/api/package.py
```diff
@@ -173,9 +169,11 @@ def report(self):
...
-        for name, local_ver_str in local_package_vers.items():
-            latest_ver_str = distro_package_vers.get(name, '')
-            local_ver = version.parse(local_ver_str).base_version
-            latest_ver = version.parse(latest_ver_str).base_version
+        for name, local_ver in local_package_vers.items():
+            latest_ver = distro_package_vers.get(name, '')
+            if local_ver:
+                local_ver = version.parse(local_ver).base_version
+            if latest_ver:
+                latest_ver = version.parse(latest_ver).base_version
```
clalancette, line 177:
```suggestion
            if latest_ver:
                latest_ver = version.parse(latest_ver).base_version
            else:
                latest_ver = 'N/A'
```

(I think this just looks nicer on the report)

clalancette (changes requested):
Thanks for iterating!  This looks a lot closer to what I think we should have.  I've left two more fixes, then I'm happy to run CI on this.

ros2doctor/ros2doctor/api/package.py
```diff
@@ -97,43 +97,46 @@ def compare_versions(result: Result, local_packages: dict, distro_packages: dict
...
-        latest_ver = version.parse(latest_ver_str).base_version
+            missing_req.append(name)
+
+        if not local_ver_str or not latest_ver_str:
+            continue  # Logging is already done by missing_local/missing_req
+
+        try:
+            local_ver = version.parse(local_ver_str).base_version
+        except version.InvalidVersion:
+            doctor_warn(f'Failed to parse local version \'{local_ver_str}\' of package \'{name}\'')
```
clalancette, line 115: Linters are complaining about this.  Fix it with:

```suggestion
            doctor_warn(f"Failed to parse local version '{local_ver_str}' of package '{name}'")
```

ros2doctor/ros2doctor/api/package.py
```diff
@@ -97,43 +97,46 @@ def compare_versions(result: Result, local_packages: dict, distro_packages: dict
...
+        try:
+            latest_ver = version.parse(latest_ver_str).base_version
+        except version.InvalidVersion:
+            doctor_warn(f'Failed to parse latest version \'{latest_ver_str}\' of package \'{name}\'')
```
clalancette, line 119:
```suggestion
            doctor_warn(f"Failed to parse latest version '{latest_ver_str}' of package '{name}'")
```

clalancette: Flake8 is also complaining that this line is too long.  I don't have an exact solution on how to fix it right now, but that needs to be fixed before we run more full CI on this.

clalancette (changes requested):

MatthijsBurgh:
@clalancette The linelength issue should be solved with change the quotes.

clalancette (changes requested):
Sorry, two more cosmetic changes I noticed when testing. After this I'll run CI.

fujitatomoya (approved):
lgtm with green CI

MatthijsBurgh:
@clalancette done

clalancette (approved):
Thank you for all of the iteration!  This looks good to me, I'll run CI on it next.

clalancette:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19666/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14173/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20390/)

MatthijsBurgh:
@clalancette could you please create a new release with this fix for the next sync?

PR #870, opened 2023-12-01
Document auto header timestamping.

contradict:
Discovered this while reading the code to add a similar feature, seems too useful to stay hidden.

### DIFF
--- a/ros2topic/ros2topic/verb/pub.py
+++ b/ros2topic/ros2topic/verb/pub.py
@@ -64,7 +64,10 @@ class PubVerb(VerbExtension):
             'values', nargs='?', default='{}',
             help='Values to fill the message with in YAML format '
                  "(e.g. 'data: Hello World'), "
-                 'otherwise the message will be published with default values')
+                 'otherwise the message will be published with default values.'
+                 ' If the message contains a std_msgs.msg.Header, using the '
+                 ' value "auto" for this field will set the timestamp '
+                 '(e.g. \'header: "auto"\').')
         arg.completer = TopicMessagePrototypeCompleter(
             topic_type_key='message_type')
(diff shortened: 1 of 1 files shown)


ros2topic/ros2topic/verb/pub.py
```diff
@@ -64,7 +64,10 @@ def add_arguments(self, parser, cli_name):
             'values', nargs='?', default='{}',
             help='Values to fill the message with in YAML format '
                  "(e.g. 'data: Hello World'), "
-                 'otherwise the message will be published with default values')
+                 'otherwise the message will be published with default values.'
+                 ' If the message contains a std_msgs.msg.Header, using the '
+                 ' value "auto" for this field will set the timestamp '
+                 '(e.g. \'header: "auto"\').')
```
fujitatomoya, line 70: there is also 'now' which can be passed as a value to a 'builtin_interfaces.msg.Time' field, this will be be set to the current time. i am not inclined to add the each message type specific behavior to here, instead can we add some description to,

- https://github.com/ros2/rcl_interfaces/blob/rolling/builtin_interfaces/msg/Time.msg
- https://github.com/ros2/common_interfaces/blob/rolling/std_msgs/msg/Header.msg

any thought?

fujitatomoya: > In what situations other than using this tool would these values be useful?

yeah, i thought about this too. good question.

> Perhaps I should add this to the tutorial [here](https://docs.ros.org/en/rolling/Tutorials/Beginner-CLI-Tools/Understanding-ROS2-Topics/Understanding-ROS2-Topics.html#ros2-topic-pub)?

sounds reasonable to me.

Issue #739, opened 2022-07-18
Add support for ament_cmake_python in ros2 pkg create


RFRIEDM-Trimble:
#### Feature description

ROS2 CLI allows creating packages for `cmake`, `ament_cmake`, `ament_python`. Considering there isn't a complete example in the documentation for [ament_cmake_python](https://docs.ros.org/en/humble/How-To-Guides/Ament-CMake-Python-Documentation.html) , it would be helpful to support generating one for users. 

#### Implementation considerations

Follow best practices for using ament. 
Supply a working example with both C++ and python code. 
Ensure dependencies are exported correctly.

bismuth01:
I have a few questions regarding the issue in my attempt to solve it: -
- Do I strictly follow the folder structure in the documentation as it seems having `include` and `src` directory like in `ament_cmake` might be better.
- Is it only for the rolling branch ?
- Any other suggestions or coding style or best practices I have to adhere to ?

christophebedard:
Someone attempted this before; you should take a look at the PR and the discussion: https://github.com/ros2/ros2cli/pull/904.

And yes, you should target the `rolling` branch, and we can consider backporting it to other branches afterwards.

bismuth01:
I read the discussion, so to solve the issue, rather than implement a separate `ament_cmake_python` package type, a template type of `python` must be implemented with build type `ament_cmake`.

So, what are the new requirements to solve the issue and is it still open ?

christophebedard:
I think that's about it! It should still use `--build-type ament_cmake` because a package using `ament_cmake_python` is still a package of build type `ament_cmake`: [`ament_cmake_python` is just a set of CMake utilities for `ament_cmake` packages that contain Python code](https://docs.ros.org/en/jazzy/How-To-Guides/Ament-CMake-Python-Documentation.html). We just want to use a different template file, and we'd like for it to be selected using `--template-name python` (+ `--build-type ament_cmake`).

In summary:

1. `--build-type ament_cmake`: uses the [current template](https://github.com/ros2/ros2cli/blob/rolling/ros2pkg/ros2pkg/resource/ament_cmake/CMakeLists.txt.em)
2. `--build-type ament_cmake --template-name cmake`: uses the current template too (we assume that `--template-name cmake` is the default with `--build-type ament_cmake`)
[...]

PR #904, opened 2024-04-29
Support `ament_cmake_python` for `pkg create`

m12watanabe1a:
# About this PR
- Implement new argument `--template-name` for selecting different template
- Giving '?' shows the list of the available template
- Support new template `ament_cmake_python`

## Related Issue
https://github.com/ros2/ros2cli/issues/739

## How to use
```bash
ros2 pkg create --build-type ament_cmake --template-name ament_cmake_python my_project
```

This command will genrate,
```
.
└── my_project
    ├── CMakeLists.txt
    ├── package.xml
    └── my_project
        └── __init__.py
```

Both `<node_name>` or `<library_name>` given in the cli will generate node or library in **cpp**.

### DIFF
--- a/ros2pkg/ros2pkg/api/create.py
+++ b/ros2pkg/ros2pkg/api/create.py
@@ -112,10 +112,6 @@ def create_package_environment(package, destination_directory):
     if package.get_build_type() == 'ament_python':
         print('creating source folder')
         source_directory = _create_folder(package.name, package_directory)
-    if package.get_build_type() == 'ament_cmake_python':
-        print('creating source folder')
-        source_directory = _create_folder('src', package_directory)
-        include_directory = _create_folder(package.name, package_directory + os.sep + 'inlclude')
 
     return package_directory, source_directory, include_directory
 
--- a/ros2pkg/ros2pkg/api/create.py
+++ b/ros2pkg/ros2pkg/api/create.py
@@ -318,4 +318,4 @@ def populate_ament_cmake_python(package, package_directory, cpp_node_name, cpp_l
                           '__init__.py',
                           {})
 
-    test_directory = _create_folder('test', package_directory)
+    _ = _create_folder('test', package_directory)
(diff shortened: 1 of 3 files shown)


clalancette (changes requested):
As far as I can see, there is no package type called `ament_cmake_python`.  For all of our python packages that are built using CMake, the package.xml type is just `ament_cmake` ([one example](https://github.com/ros2/rclpy/blob/rolling/rclpy/package.xml))

@m12watanabe1a Can you explain more about what you are trying to accomplish here?

ros2pkg/ros2pkg/api/create.py
```diff
@@ -112,6 +112,10 @@ def create_package_environment(package, destination_directory):
     if package.get_build_type() == 'ament_python':
         print('creating source folder')
         source_directory = _create_folder(package.name, package_directory)
+    if package.get_build_type() == 'ament_cmake_python':
```
clalancette, line 115: There are currently no instances of ROS 2 packages where this would be true.  That is, `package.get_build_type()` will never return `ament_cmake_python`.

That said, it would still be worthwhile to have some way to create a package template for a Python package using CMake.  However, it will have to be done a different way because it will use a `build_type` of `ament_cmake`.  Maybe we can add in a new command-line argument that tells it to generate Python installation stuff in the CMakeLists.txt?

clalancette (changes requested):

sloretz:
I added an agenda item to the next ROS 2 meeting to get feedback on the naming of `--template-name` , afterwards the ros2cli maintainers will be assigned to review. 🧇

Thank you for your patience!

m12watanabe1a:
Thank you so much :pray:

christophebedard:
Just to follow up (https://github.com/ros2/ros2cli/pull/904#issuecomment-2103606131), this was discussed on May 14, 2024. The conclusion was:

1. Use `--build-type ament_cmake` and `--template-name python` (where `--template-name cmake`, or maybe simply the absence of `--template-name`, would be the default and correspond to the current `ament_cmake` template)
2. Make this template `ros2pkg/ros2pkg/resource/ament_cmake/python/CMakeLists.txt.em`

PR #912, opened 2024-06-16
Update daemon.py for Python 3.8 compatibility

AhmedMoaz:
Minor fix for containers running Pyhton version < 3.9.0 

Before and after: 
Fix

Tested on:
environment

PR #915, opened 2024-06-20
Humble corrections

nidhicj:
Added a fdlimit = int(line[len(string_to_find):].strip()) in ros2cli/ros2cli/node/daemon.py replacing use of "removeprefix" as it is supported only from python3.9


clalancette:
This is a duplicate of #912

Issue #830, opened 2023-05-11
Make possible for ros2 topic hz to monitor the rates of multiple topics simultaneously as ros1 rostopic hz does. 


martinakos:
#### Feature description
Make possible for ros2 topic hz to monitor the rates for of multiple topics simultaneously as ros1 rostopic hz does. 

In ROS1 I can do:

`rostopic hz /topic1 /topic2 /topic3`

In ROS2 Humble I can only check the rate for one topic at a time:

`ros2 topic hz /topic1`

fujitatomoya:
yeah, i think it is nice to have.

the following is output from `rostopic hz` from ROS 1 just a reference.

```bash
root@4b5f2b7e1de8:~# rostopic hz /chatter /rosout
subscribed to [/chatter]
subscribed to [/rosout]
 topic      rate   min_delta   max_delta    std_dev    window
=============================================================
/chatter   9.995   0.09971     0.1004      0.0001937   10
/rosout    9.996   0.09978     0.1004      0.0001766   10

 topic      rate   min_delta   max_delta    std_dev    window
=============================================================
/chatter   9.998   0.0997      0.1004      0.0001685   20
/rosout    9.998   0.09967     0.1004      0.0001661   20
....<snip>
```

yaswanth1701:
Hi  @fujitatomoya This feature sounds cool, If it's not implemented yet, I would like to give it a try. Can I ?

fujitatomoya:
i think @iuhilnehc-ynos created https://github.com/ros2/ros2cli/pull/831 already. it would be nice if you can do review for that 👍

yaswanth1701:
> i think @iuhilnehc-ynos created #831 already. it would be nice if you can do review for that 👍

I haven't done a review before for PR but I will give it a try, thanks!

felixdivo:
Just a side note: If once accepts new dependencies, it might make sense to not reinvent table printing here and use something proven. For example: [rich](https://rich.readthedocs.io/en/stable/tables.html) or [others](https://stackoverflow.com/a/26937531/3753684).

clalancette:
> Just a side note: If once accepts new dependencies, it might make sense to not reinvent table printing here and use something proven. For example: [rich](https://rich.readthedocs.io/en/stable/tables.html) or [others](https://stackoverflow.com/a/26937531/3753684).

[...]

PR #831, opened 2023-05-12
to support multiple topics for topic hz

iuhilnehc-ynos:
to support https://github.com/ros2/ros2cli/issues/830

PR #929, opened 2024-09-04
Support multiple topics via ros2 topic hz.

fujitatomoya:
replaces https://github.com/ros2/ros2cli/pull/831 and closes https://github.com/ros2/ros2cli/issues/830

Note: after the merge, we need to update the documentation

### DIFF
--- a/ros2topic/ros2topic/verb/hz.py
+++ b/ros2topic/ros2topic/verb/hz.py
@@ -34,11 +34,13 @@ from collections import defaultdict
 import functools
 import math
 import threading
+import time
 
 import rclpy
 
 from rclpy.clock import Clock
 from rclpy.clock import ClockType
+from rclpy.executors import ExternalShutdownException
 from rclpy.qos import qos_profile_sensor_data
 from ros2cli.node.direct import add_arguments as add_direct_node_arguments
 from ros2cli.node.direct import DirectNode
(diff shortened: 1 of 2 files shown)


ros2topic/ros2topic/verb/hz.py
```diff
@@ -225,43 +225,98 @@ def get_hz(self, topic=None):
...
+    for topic in topics:
+        msg_class = get_msg_class(
+            node, topic, blocking=True, include_hidden_topics=True)
 
-    if msg_class is None:
-        node.destroy_node()
-        return
+        if msg_class is None:
+            node.destroy_node()
+            return
```
fujitatomoya, line 298: I believe that what we can do better here is to skip the unknown message classes with printing the warning message for user, and proceed to measure the frequency with found ones.

TannerGilbert:
Thanks for the quick reply on the other PR. I have one more question. Will changes like these only be merged into rolling or will there also be backports to older still supported versions like Humble?

fujitatomoya:
@TannerGilbert 

basically we enforce API compatibility for released distribution, see more details for https://docs.ros.org/en/rolling/The-ROS2-Project/Contributing/Developer-Guide.html#versioning. and this PR breaks API combability including behavior change, so we would not backport the fix to released distributions.

If we want to backport, what i would suggest is that,

1. merge this fix into rolling 1st, then wait for a couple of weeks as soak time to make sure nothing happens in the main.
2. and then, consider backport fix with API compatibility to jazzy, iron and humble. (each distro could have base code difference)

CC: @clalancette

ros2topic/ros2topic/verb/hz.py
fujitatomoya: fixed with https://github.com/ros2/ros2cli/pull/929/commits/0d066fd18b55d38a27d000f4404cf86458e99258

fujitatomoya:
@ahcorde @clalancette @sloretz i would like to request the review.

Issue #930, opened 2024-09-11
ros2 param dump UnboundLocalError


reinzor:
```
$ ros2 param dump /kalman_filter_map 
Traceback (most recent call last):
  File "/opt/ros/jazzy/bin/ros2", line 33, in <module>
    sys.exit(load_entry_point('ros2cli==0.32.1', 'console_scripts', 'ros2')())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/ros/jazzy/lib/python3.12/site-packages/ros2cli/cli.py", line 91, in main
    rc = extension.main(parser=parser, args=args)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/ros/jazzy/lib/python3.12/site-packages/ros2param/command/param.py", line 39, in main
    return extension.main(args=args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/ros/jazzy/lib/python3.12/site-packages/ros2param/verb/dump.py", line 103, in main
    f"'{node_name.full_name}': {e}", file=sys.stderr)
                                ^
UnboundLocalError: cannot access local variable 'e' where it is not associated with a value
```

reinzor:
@fujitatomoya this issue can be reproduces by the scenario described in https://github.com/cra-ros-pkg/robot_localization/issues/903 . This discussion might be out of scope for this issue but just a friendly FYI as you are the maintainer of this repo :) .

PR #931, opened 2024-09-11
fix(ros2param-dump): UnboundLocalError

reinzor:
#930

### DIFF
--- a/ros2param/ros2param/verb/dump.py
+++ b/ros2param/ros2param/verb/dump.py
@@ -100,7 +100,7 @@ class DumpVerb(VerbExtension):
             if parameter_values is None:
                 print(
                     'Exception while calling get_parameters service of node '
-                    f"'{node_name.full_name}': {e}", file=sys.stderr)
+                    f"'{node_name.full_name}'", file=sys.stderr)
                 return
 
             for param_name, pval in zip(response, parameter_values):


clalancette (changes requested):
First of all, thank you for opening a PR to fix what is obviously a copy-n-paste error.

That said, I find the way that this entire function is called problematic.  First of all, `get_parameter_values` is defined as a static method on the class, but there is no need for that; it doesn't access anything in the class or the object.

But more problematic is that the error handling in `get_parameter_values` just seems wildly off.  It looks to me like `call_get_parameters` can throw an exception if it fails, and this doesn't handle that at all.  In point of fact, it looks to me like the only way we can end up in the situation that you ran into is when we call `ros2 param dump`, but there are no parameters.  That is a valid (though uncommon) situation.

My suggestion is that we rework this a bit more.  Move the functionality of `get_parameter_values` inline in the place where it is used.  Make sure to capture exceptions, and print an error.  And if there are no parameters, that is fine, just don't print any.

Does that make sense to you?

reinzor:
Makes sense. I will try to take a look at it tomorrow but I don't know how much time I can spend on refactoring this.

fujitatomoya:
@reinzor if you can take a look at https://github.com/ros2/ros2cli/pull/933 which replaces this PR, that would be really appreciated.

PR #933, opened 2024-09-16
cosmetic fixes for ros2param dump command.

fujitatomoya:
closes https://github.com/ros2/ros2cli/issues/930

replaces https://github.com/ros2/ros2cli/pull/931

### DIFF
--- a/ros2param/ros2param/verb/dump.py
+++ b/ros2param/ros2param/verb/dump.py
@@ -68,7 +68,6 @@ class DumpVerb(VerbExtension):
             yaml_output = {node_name.full_name: {'ros__parameters': {}}}
 
             # retrieve parameter names
-            response = None
             response = call_list_parameters(node=node, node_name=absolute_node_name)
             if response is None:
                 print(
--- a/ros2param/ros2param/verb/dump.py
+++ b/ros2param/ros2param/verb/dump.py
@@ -91,9 +91,9 @@ class DumpVerb(VerbExtension):
                     'Exception while calling get_parameters service of node '
                     f"'{node_name.full_name}': {e}", file=sys.stderr)
             if response.values is None:
-                # no parameters available
-                print(f"'{node_name.full_name}' does not possess any parameters.")
-                return
+                # pass through here, no parameters are available with this node.
+                # since this is not failure, it proceeds to print the yaml as consistent behavior.
+                pass
             parameter_values = [get_value(parameter_value=i) for i in response.values]
 
             # create dictionary with parameter names and values
--- a/ros2param/ros2param/verb/dump.py
+++ b/ros2param/ros2param/verb/dump.py
@@ -46,19 +46,6 @@ class DumpVerb(VerbExtension):
             '--timeout', metavar='N', type=int, default=1,
             help='Wait for N seconds until node becomes available (default %(default)s sec)')
 
-    @staticmethod
-    def get_parameter_values(node, node_name, params):
-        response = call_get_parameters(
-            node=node, node_name=node_name,
-            parameter_names=params)
-
-        # requested parameter not set
-        if not response.values:
-            return None
-
-        # extract type specific value
-        return [get_value(parameter_value=i) for i in response.values]
-
     def insert_dict(self, dictionary, key, value):
         split = key.split(PARAMETER_SEPARATOR_STRING, 1)
         if len(split) > 1:
@@ -80,7 +67,7 @@ class DumpVerb(VerbExtension):
         with DirectNode(args) as node:
             yaml_output = {node_name.full_name: {'ros__parameters': {}}}
 
-            # retrieve values
+            # retrieve parameter names
             response = call_list_parameters(node=node, node_name=absolute_node_name)
             if response is None:
                 print(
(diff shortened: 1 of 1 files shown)


fujitatomoya:
@clalancette could you take a look?

ros2param/ros2param/verb/dump.py
```diff
@@ -93,17 +80,24 @@ def main(self, *, args):  # noqa: D102
...
+            try:
+                response = call_get_parameters(
+                    node=node, node_name=absolute_node_name, parameter_names=parameter_names)
+            except RuntimeError as e:
                 print(
                     'Exception while calling get_parameters service of node '
                     f"'{node_name.full_name}': {e}", file=sys.stderr)
+            if response.values is None:
+                # no parameters available
+                print(f"'{node_name.full_name}' does not possess any parameters.")
```
fujitatomoya, line 95: Should we print this to user? i do not have strong opinion here though.

- keep this printing so that user can understand there are no parameter available with this node.
- no print message here, so that nothing comes up in the terminal.
- we do not return here, so that it outputs the yaml `[node_name.full_name]['ros__parameters']` only.

i think printing message would be the most comprehensive for user perspective.

clalancette: It's a tough one to answer.  I think it kind of depends on whether we intend this to be used as an end-user interface, or as part of a pipeline.  For an end-user interface, I think printing "<node> has no parameters" is best.  For a pipeline, I think the empty YAML is best.  I guess we could add a command-line option to control it, but that doesn't feel great either.

I guess my personal preference would be to remove the return and print the empty YAML.  But I'd also like to hear opinions from @ros2/team .

reinzor: I agree with @clalancette that an empty YAML is fine. This command should only write a yaml file to stdout.

sloretz: How about printing the error message to STDERR and an empty yaml on STDOUT? Then it can still be used as a pipeline and have a human readable error message.

fujitatomoya: thank you very much for the comments.

i decided to take `empty YAML` process. Having no parameters is NOT the error, that is just one of the success case. for consistent behavior, i think we can pass through this case and proceed to print the (almost) empty YAML.

Note: the question is `can we create a node with no parameters available?` the short answer is yes but only if user application creates the node with `rcl` interfaces. there are 2 default (aside from rclcpp QoS parameters) parameters available when it creates the node, that are `use_sim_time` and `start_type_description_service`. `use_sim_time` can be undeclared because it is not `read_only` parameter. but we cannot undeclare `start_type_description_service` parameter because it is declared with `read_only` parameter by `rclcpp` and `rclpy`. after all, this `no parameters available` situation is so unlikely to happen. because of this, i ended up not having the test case for this either.

fujitatomoya:
Pulls: ros2/ros2cli#933
Gist: https://gist.githubusercontent.com/fujitatomoya/26e84d2c1dd65ac3bc6ac4ce41d96a89/raw/2ac6489ad24ed7fefa8115f92d9ea3b3b88f64c4/ros2.repos
BUILD args:  --packages-above-and-dependencies ros2param
TEST args:  --packages-above ros2param
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14574
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21851/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16121/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1357/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22562/)

fujitatomoya:
@sloretz CI is green, could you review?

ros2param/ros2param/verb/dump.py
```diff
@@ -93,17 +80,24 @@ def main(self, *, args):  # noqa: D102
...
+                response = call_get_parameters(
+                    node=node, node_name=absolute_node_name, parameter_names=parameter_names)
+            except RuntimeError as e:
                 print(
                     'Exception while calling get_parameters service of node '
                     f"'{node_name.full_name}': {e}", file=sys.stderr)
-                return
-
-            for param_name, pval in zip(response, parameter_values):
+            if response.values is None:
```
clalancette, line 93: I think this line of code is a bug.  If `call_get_parameters` throws an exception, then we'll fall through to here, but `response` isn't assigned to anything.  So I think we'll end up with a `NameError`.  Maybe what we should do is set a variable called `response_values = None` before `call_get_parameters`, and then only if call_get_parameters succeeds assign result.values to `result_values`.

clalancette (changes requested):

ros2param/ros2param/verb/dump.py
fujitatomoya: right, this is bug. addressed in https://github.com/ros2/ros2cli/pull/933/commits/bb452ccbebc630528c27a721ebd388aff019cbca

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21881/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16149/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1386/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22594/)

ros2param/ros2param/verb/dump.py
```diff
@@ -80,7 +67,8 @@ def main(self, *, args):  # noqa: D102
         with DirectNode(args) as node:
             yaml_output = {node_name.full_name: {'ros__parameters': {}}}
 
-            # retrieve values
+            # retrieve parameter names
+            response = None
```
clalancette, line 71: This one shouldn't be needed; we are assigning it to something in the next call.

```suggestion
```

clalancette (changes requested):
One more minor fix, then I think this looks good.

ros2param/ros2param/verb/dump.py
fujitatomoya: fixed in https://github.com/ros2/ros2cli/pull/933/commits/5b47d97452e60f7ce8fcd9abce2d8c0fd58ad277

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21887/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16155/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1395/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22605/)

clalancette (approved):

fujitatomoya:
@Mergifyio backport humble iron jazzy

PR #947, opened 2024-11-18
Allow reuse address when firing up the ros2cli daemon.

clalancette:
Especially in tests where the daemon is being repeatedly created and destroyed, it can take some time for the kernel to actually allow the address to be rebinded (even after the old process has exited).  This can lead to some situations where we fail to spawn the daemon.

To fix this, set "allow_reuse_address" inside the LocalXMLRPCServer, which will set SO_REUSADDR on the socket.

### DIFF
--- a/ros2cli/ros2cli/daemon/__init__.py
+++ b/ros2cli/ros2cli/daemon/__init__.py
@@ -151,6 +151,13 @@ def serve(server: LocalXMLRPCServer, *, timeout: int = 2 * 60 * 60):
             pass
 
 
+def serve_and_close(server: LocalXMLRPCServer, *, timeout: int = 2 * 60 * 60):
+    try:
+        serve(server, timeout=timeout)
+    finally:
+        server.server_close()
+
+
 def main(*, argv=None):
     parser = argparse.ArgumentParser(
         formatter_class=argparse.ArgumentDefaultsHelpFormatter)
--- a/ros2cli/ros2cli/node/daemon.py
+++ b/ros2cli/ros2cli/node/daemon.py
@@ -46,7 +46,7 @@ class DaemonNode:
                 for method in self._proxy.system.listMethods()
                 if not method.startswith('system.')
             ]
-        except ConnectionRefusedError:
+        except (ConnectionRefusedError, ConnectionResetError):
             return False
         return True
 
--- a/ros2cli/ros2cli/node/daemon.py
+++ b/ros2cli/ros2cli/node/daemon.py
@@ -168,7 +168,7 @@ def spawn_daemon(args, timeout=None, debug=False):
             'rmw_implementation': rclpy.get_rmw_implementation_identifier()}
 
         daemonize(
-            functools.partial(daemon.serve, server),
+            functools.partial(daemon.serve_and_close, server),
             tags=tags, timeout=timeout, debug=debug)
     finally:
         server.server_close()
--- a/ros2cli/ros2cli/xmlrpc/local_server.py
+++ b/ros2cli/ros2cli/xmlrpc/local_server.py
@@ -42,15 +42,15 @@ class LocalXMLRPCServer(SimpleXMLRPCServer):
 
     def server_bind(self):
         # Prevent listening socket from lingering in TIME_WAIT state after close()
-        self.socket.setsockopt(
-            socket.SOL_SOCKET, socket.SO_LINGER, struct.pack('ii', 1, 0))
+        # self.socket.setsockopt(
+        #     socket.SOL_SOCKET, socket.SO_LINGER, struct.pack('ii', 1, 0))
         super(LocalXMLRPCServer, self).server_bind()
 
     def get_request(self):
         # Prevent accepted socket from lingering in TIME_WAIT state after close()
         sock, addr = super(LocalXMLRPCServer, self).get_request()
-        sock.setsockopt(
-            socket.SOL_SOCKET, socket.SO_LINGER, struct.pack('ii', 1, 0))
+        # sock.setsockopt(
+        #     socket.SOL_SOCKET, socket.SO_LINGER, struct.pack('ii', 1, 0))
         return sock, addr
 
     def verify_request(self, request, client_address):
--- a/ros2cli/ros2cli/xmlrpc/local_server.py
+++ b/ros2cli/ros2cli/xmlrpc/local_server.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import os
 import socket
 import struct
 # Import SimpleXMLRPCRequestHandler to re-export it.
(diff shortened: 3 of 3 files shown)


clalancette:
Pulls: https://github.com/ros2/ros2cli/pull/947
Gist: https://gist.githubusercontent.com/clalancette/c79668ce9bddc836dcdeae6fb121cc2a/raw/2342a02ab486617cbac78c10af6ae3bacbacd367/ros2.repos
BUILD args: 
TEST args: 
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14831
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22138/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16393/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1661/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22881/)

clalancette:
After discussing this with @cottsay a bit, we shouldn't do this on Windows; SO_REUSEADDR with multiple ports is basically UB.  See https://learn.microsoft.com/en-us/windows/win32/winsock/using-so-reuseaddr-and-so-exclusiveaddruse .  So I'm going to fix this to only make this change on Linux.  It does mean that on Windows we still may have the flaky test, but at least we'll fix it on Linux.  We'll have to think about how to do this on Windows separately.

clalancette:
And 7cb8b24 fixes it to do just that.

ahcorde (approved):

clalancette:
Hmpf.  After looking at this further, @cottsay found that this is reverting previous work: https://github.com/ros2/ros2cli/pull/622 .  So marking this as Draft for now.

fujitatomoya (commented):
IMO, it would make more sense to enable `allow_reuse_address` as this PR suggested?

the socket is set with `SO_LINGER` with timeout `0`, that means it causes the socket to send a TCP RST (Reset) upon closure and terminates the connection immediately, discarding any unsent data.

`SO_REUSEADDR` allows binding to a port that is already in use (e.g., in the `TIME_WAIT` state). It doesn't terminate the `TIME_WAIT` state but bypasses the restriction, enabling rebinding.

clalancette:
All right, with all of the latest pushes here, I got things to work far more reliably on both Linux and Windows.  I'm going to explain a bit about how we got here.

# What's the real problem we are trying to solve?
The original problem here is that running the `test_strategy.py` test in the nightly repeated jobs "sometimes" fails.  You can see that for example in https://ci.ros2.org/view/nightly/job/nightly_linux-rhel_repeated/2098/testReport/junit/ros2cli.ros2cli.test/test_strategy/test_with_daemon_spawn_2_3_/ .

# Why is that test failing?
What happens when that test fails is pretty simple; the daemon fails to spawn correctly.  Because we have no checking around this in `NodeStrategy`, it is basically a silent failure until we try to verify that we've connected in the tests.  Looking at https://github.com/ros2/ros2cli/blob/8766c334d4995a6b5214edc8ebe883fd9a0c1ddc/ros2cli/ros2cli/node/daemon.py#L104-L176 , I believe the only real reason the daemon can silently fail to spawn is if we get EADDRINUSE on the socket.  In turn, that happens because it can be the case that a socket from an "old" daemon that was/is shutting down still hasn't been cleaned up by the kernel.  In a bit more detail, the kernel can leave old listening sockets open in TCP TIME_WAIT state for a while after the socket was closed.  In that situation, trying to create a "new" daemon will fail with EADDRINUSE (at least, by default).  This also explains why we mostly see this in the tests; they are the only ones that are spawning and killing daemons in a loop.  For most normal users, this won't be the case.

# How did we get to where the current code?
There have been a few attempts to fix flakiness in the ros2 daemon in the past.  These include #620 , #622 , and #652 .  These all changed things in various ways, but the key PR was #652, which made spawning the daemon a reliable operation.

#622 made some changes to change the sockets to add SO_LINGER with a zero timeout.  That improved, but did not totally solve the situation.  It also has its own downsides, as SO_LINGER doesn't gracefully terminate connections and instead just sends RST on the socket and terminates it.

# What's the solution?
The solution here is in 3 parts, though one of the parts is platform-dependent.  This PR implements all 3 parts:

1.  When the daemon is exiting cleanly, it should *explicitly* shutdown the socket that it was using for the XMLRPC server.  That will cleanly shutdown the socket, and tell the kernel it can start the cleanup.  On its own, this does not completely solve the problem, but it reduces the amount of time that things are hanging about waiting for the Python interpreter and/or the kernel to implicitly clean things up.
2.  We should *not* specify SO_LINGER on the daemon sockets.  As mentioned above, this is actually something of an anti-pattern and does not properly terminate connections with FIN (it just sends RST).
3.  We should specify SO_REUSEADDR, but only on Unix.  On Unix, SO_REUSEADDR essentially means "allow binding to an address/port that is in TCP TIME_WAIT (but not that is otherwise in use)".  This is exactly the behavior we want.  On Windows, SO_REUSEADDR causes undefined behavior, as it can cause a socket to bind even if there is something else bound already.  Because of that, we want to set SO_REUSEADDR on Unix, but not Windows.

Finally, while testing here I had to add in one bugfix to make things reliable on Windows, which is to also catch `ConnectionResetError`.  That arises because we can attempt to "connect" to a daemon that is in the process of shutting down.  In that case, we should also consider the daemon not "connected".

With all of the fixes in this PR, I'm no longer able to reproduce the flakiness in the server on either Linux or Windows.  Next up I'm going to run regular CI on this, as well as an equivalent of the nightly repeated jobs to make sure things are happy.

clalancette:
Pulls: https://github.com/ros2/ros2cli/pull/947
Gist: https://gist.githubusercontent.com/clalancette/e73e1a2dd42b49fe06f374ccbb538d51/raw/2342a02ab486617cbac78c10af6ae3bacbacd367/ros2.repos
BUILD args: 
TEST args: 
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14838
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22145/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16400/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1669/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22888/)

clalancette:
Here are the repeated jobs:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22146/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16401/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1670/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22889/)

cottsay (approved):

fujitatomoya (approved):
@clalancette good study, thanks for sharing 👍

PR #948, opened 2024-11-20
[ros2pkg] Integrate features of 3rd-party ros2-pkg-create CLI tool

lreiher:
We have recently open-sourced our own package generation tool [`ros2-pkg-create`](https://github.com/ika-rwth-aachen/ros2-pkg-create). Please also see my introductory post on [ROS Discourse](https://discourse.ros.org/t/introducing-ros2-pkg-create-new-powerful-ros-2-package-generator/40669).

I'm opening this draft PR in response to [@tfoote suggesting to integrate `ros2-pkg-create`s features into the official CLI](https://discourse.ros.org/t/introducing-ros2-pkg-create-new-powerful-ros-2-package-generator/40669/2).

> [!NOTE]
> This PR is not carrying any changes just yet. I'm mainly opening this to share some insights into the features of `ros2-pkg-create`, how it's different than this repository's `ros2 pkg create`, and how an integration here could look like. Ideally, we could gather some feedback and suggestions before starting the actual integration efforts.

---

### Features of [`ros2-pkg-create`](https://github.com/ika-rwth-aachen/ros2-pkg-create)

- templates for three types of packages
  - C++ Package
  - Python Package
  - Interfaces Package
- support for basic and advanced features for both C++ and Python
  - publisher
  - subscriber
  - parameter loading
  - launch file (py/xml/yml)
  - service server
  - action servertimer callback, C++ component, C++ lifecycle, [docker-ros](https://github.com/ika-rwth-aachen/docker-ros)
- interactive CLI questionnaire for options not specified as CLI arguments

image

### DIFF
--- a/ros2pkg/ros2pkg/verb/create.py
+++ b/ros2pkg/ros2pkg/verb/create.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# TODO: integrate features of https://github.com/ika-rwth-aachen/ros2-pkg-create
+
 import getpass
 import os
 import shutil


clalancette:
So I would absolutely love to get some of these features into `ros2 pkg create`.

That said, I think we have to come to an agreement on the templating engine before we do anything.  In `ros2 pkg create`, we currently use empy.  I know that your `ros2-pkg-create` uses Jinja currently.  I'd really, really prefer to keep using `empy`, because that is what we have "standardized" in ROS for our templating engine.  Would you be willing to make changes to use empy rather than Jinja?

lreiher:
I definitely see your point regarding `empy`, but the limitation on our side is that [copier](https://copier.readthedocs.io/en/stable/) is based on Jinja. We see the copier integration, which provides the interactive questionnaire, as one of the core features that we wouldn't want to drop.

Taking this a step further, one could of course try to bring `empy` support to copier. This is out of scope of what we could contribute, though.

On the other hand, we could translate the Jinja templates to empy templates (I guess this should be pretty straighforward?), add some CLI arguments here, and be done with it. To me, this integration wouldn't feel complete, though, and we would need to think about how much effort we would want to put into this from our side. We would definitely be open to someone else porting the templates, though.

clalancette:
> On the other hand, we could translate the Jinja templates to empy templates (I guess this should be pretty straighforward?), add some CLI arguments here, and be done with it. To me, this integration wouldn't feel complete, though, and we would need to think about how much effort we would want to put into this from our side. We would definitely be open to someone else porting the templates, though.

Yeah, this is the kind of thing I would suggest here.  We'd have to do this anyway, because it is far easier to get things in one-at-a-time rather than as a giant chunk of work.  I'm not sure if you are up for doing that, but this is what I would suggest at least as a starting point.

Issue #951, opened 2024-12-04
[Question] - does ros2 topic echo support multiple fields


cwrx777:
Hi,

is it possible to echo more than one fields, e.g.
```bash
  ros2 topic echo /topic --field field_1 --field field_2
```

clalancette:
In the current code, it does not support this; it can only print out either the whole message, or a single field.

That said, I think it would be an OK feature to be able to select more than one.  If you'd like to contribute a PR to implement it, we'd be happy to review it.

CWAbhi:
@clalancette can i  work in this issue?

fujitatomoya:
@CWAbhi contributions are always welcome, thanks 👍

CWAbhi:
okay I ll raise a PR to solve this

CWAbhi:
i have raised a PR with the changes done #958 please check and if i need to edit or do any changes do tell me.

PR #958, opened 2024-12-29
Added support for echoing multiple fields in ros2 topic echo

CWAbhi:
#951 
# Now ros2 topic echo has support  for multiple fields
This PR extends the functionality of the ros2 topic echo command to allow users to specify and display multiple fields from a ROS topic message. Previously, the command supported either displaying the entire message or a single field using the --field argument. With this enhancement, users can specify multiple --field arguments to display specific fields, including nested fields.

### DIFF
--- a/ros2topic/ros2topic/verb/echo.py
+++ b/ros2topic/ros2topic/verb/echo.py
@@ -69,8 +69,8 @@ class EchoVerb(VerbExtension):
             )
         )
         parser.add_argument(
-            '--field', type=str, default=None,
-            help='Echo a selected field of a message. '
+            '--field', action='append', type=str, default=None,
+            help='Specify one or more fields to echo. Use multiple --field options for multiple fields.'
                  "Use '.' to select sub-fields. "
                  'For example, to echo the position field of a nav_msgs/msg/Odometry message: '
                  "'ros2 topic echo /odom --field pose.pose.position'",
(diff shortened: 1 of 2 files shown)


ros2topic/ros2topic/verb/echo.py
```diff
@@ -181,10 +181,15 @@ def main(self, *, args):
...
         # Validate field selection
         self.field = args.field
-        if self.field is not None:
-            self.field = list(filter(None, self.field.split('.')))
-            if not self.field:
-                raise RuntimeError(f"Invalid field value '{args.field}'")
+        if args.field:
+            for field in args.field:
+                try:
+                    value = get_field_value(message, field)
```
fujitatomoya: where is this message come from? originally this part is to validate the fields only, not printing the message. (there is no message received at here...)

fujitatomoya (commented):
@CWAbhi Does this really work??? If you are still working on this, probably you can keep this PR as draft?

besides, test would be ideal to add around,

https://github.com/ros2/ros2cli/blob/rolling/ros2topic/test/test_cli.py#L579-L638

ros2topic/ros2topic/verb/echo.py
CWAbhi: okay then i will remove this i thought i need to send the message

CWAbhi:
@fujitatomoya Yes it is working and i will add test file in it then commit it . I think the 1 check that was failing it was due to this only

ros2topic/ros2topic/verb/echo.py
CWAbhi: and i think this needed a little bit changes which i will do so i don't think to convert into draft as per my opinion. What is your opinion in this?

CWAbhi:
@fujitatomoya can you help me this failed check...

fujitatomoya:
this includes many unrelated changes, can you remove all unrelated changes? and can you reproduce the issue on your local environment?

CWAbhi:
should i change the echo.py and the test file too? or just in echo.py

CWAbhi:
@fujitatomoya i have solved the error still it is showing this and ubuntu build is not opening showing a 502 gateway tried multiple attempts can you help me out.

fujitatomoya:
@CWAbhi this is still draft and including many unrelated changes, and i think https://github.com/ros2/ros2cli/pull/964 is closer to this particular enhancement. could you review https://github.com/ros2/ros2cli/pull/964 instead, that would be appreciated.

CWAbhi:
@fujitatomoya Sure i will review it

fujitatomoya:
@CWAbhi i will go ahead to close this in favor of https://github.com/ros2/ros2cli/pull/964. thank you very much for trying and review for that.

PR #968, opened 2025-02-06
Try to use the git global user.name for maintainer-name

lgezelius:
Add code for "ros2 pkg create" so that the first default for the maintainer-name is the git user.name, which is likely a full name, instead of first using an OS login name. This creates more consistency, because the maintainer-email already defaults to the git user.email.

### DIFF
--- a/ros2pkg/ros2pkg/verb/create.py
+++ b/ros2pkg/ros2pkg/verb/create.py
@@ -113,13 +113,13 @@ class CreateVerb(VerbExtension):
             sys.exit(0)
 
         maintainer_name: str = (
-            args.maintainer_name 
-            or get_git_config('user.name') 
+            args.maintainer_name
+            or get_git_config('user.name')
             or getpass.getuser())
         maintainer = Person(maintainer_name)
         maintainer.email = (
-            args.maintainer_email 
-            or get_git_config('user.email') 
+            args.maintainer_email
+            or get_git_config('user.email')
             or f"{maintainer.name.replace(' ', '')}@todo.todo")
 
         node_name = None
--- a/ros2pkg/ros2pkg/verb/create.py
+++ b/ros2pkg/ros2pkg/verb/create.py
@@ -17,6 +17,7 @@ import os
 import shutil
 import subprocess
 import sys
+from typing import Optional
 
 import ament_copyright
 
@@ -90,7 +91,7 @@ class CreateVerb(VerbExtension):
 
     def main(self, *, args):
 
-        def get_git_config(key: str) -> str | None:
+        def get_git_config(key: str) -> Optional[str]:
             # retrieve a value from git config
             git = shutil.which('git')
             if git is not None:
--- a/ros2pkg/ros2pkg/verb/create.py
+++ b/ros2pkg/ros2pkg/verb/create.py
@@ -79,7 +79,7 @@ class CreateVerb(VerbExtension):
             '--maintainer-email',
             help='email address of the maintainer of this package'),
         parser.add_argument(
-            '--maintainer-name', default=getpass.getuser(),
+            '--maintainer-name',
             help='name of the maintainer of this package'),
         parser.add_argument(
             '--node-name',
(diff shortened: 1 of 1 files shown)


ros2pkg/ros2pkg/verb/create.py
```diff
@@ -97,13 +97,29 @@ def main(self, *, args):
...
+        if args.maintainer_name:
+            maintainer_name = args.maintainer_name
+        else:
+            # try getting the name from the global git config
+            if git is not None:
+                p = subprocess.Popen(
+                    [git, 'config', 'user.name'],
+                    stdout=subprocess.PIPE)
+                resp = p.communicate()
+                name = resp[0].decode().rstrip()
```
christophebedard: You should probably use [`subprocess.run()`](https://docs.python.org/3/library/subprocess.html#subprocess.run) instead.

> The recommended approach to invoking subprocesses is to use the [run()](https://docs.python.org/3/library/subprocess.html#subprocess.run) function for all use cases it can handle.

ros2pkg/ros2pkg/verb/create.py
```diff
@@ -97,13 +97,29 @@ def main(self, *, args):
             print('Supported licenses:\n%s' % ('\n'.join(available_licenses)))
             sys.exit(0)
 
-        maintainer = Person(args.maintainer_name)
+        git = shutil.which('git')
+
+        if args.maintainer_name:
+            maintainer_name = args.maintainer_name
+        else:
+            # try getting the name from the global git config
```
christophebedard: This is only the global git config if this isn't run in a git repo, otherwise it gets the `user.name` value from the local git repo's config file. I think it makes sense to just use whatever `git config user.name` returns, but then this comment shouldn't say "global"

christophebedard: I see that there's a similar comment below for `user.email`. Maybe you could remove "global" from that too.

christophebedard (commented):
Thanks for the PR. I think this makes sense. I just have a couple of minor comments.

ros2pkg/ros2pkg/verb/create.py
fujitatomoya: besides this, the whole process is almost copied from the following email one. i would have the internal function in this scope to get the information using key argument.

christophebedard: good point! That would indeed be nice

lgezelius: I made the recommended/suggested changes in my latest commit.

ros2pkg/ros2pkg/verb/create.py
lgezelius: I made the comment change.

christophebedard:
Pulls: ros2/ros2cli#968
Gist: https://gist.githubusercontent.com/christophebedard/01eefc42eca835a79d61cdbd95e7eacf/raw/82d20bad79a96d65b8363e722f98e719951e868b/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15144
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22477/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16729/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/2002/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/23264/)

christophebedard (approved):
Thanks for addressing the comments!

ros2pkg/ros2pkg/verb/create.py
```diff
@@ -89,6 +89,21 @@ def add_arguments(self, parser, cli_name):
             help='name of the empty library')
 
     def main(self, *, args):
+
+        def get_git_config(key: str) -> str | None:
```
christophebedard:
```suggestion
        def get_git_config(key: str) -> Optional[str]:
```

Make sure to import it (at the top of this file):

```py
from typing import Optional
```

christophebedard (changes requested):
The tests fail on Windows: https://ci.ros2.org/job/ci_linux-rhel/2002/consoleFull#console-section-425. Turns out `|` was only added in Python 3.10: https://peps.python.org/pep-0604/. And Windows uses Python 3.8: https://www.ros.org/reps/rep-2000.html#jazzy-jalisco-may-2024-may-2029.

christophebedard:
Thanks for the fix. New CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22506/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16756/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/2030/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/23295/)

christophebedard (approved):
All good now, thank you for the PR!

Issue #871, opened 2023-12-18
ROS2 topic hz gives incorrect value with camera image


EwingKang:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 22.04
- Installation type:
  - Binary/debian
- Version or commit hash:
  - Iron
- DDS implementation:
  - Default (fastDDS)
- Client library (if applicable):
  - rclpy (from `ros2 topic hz`)

#### Steps to reproduce issue

I'm using a Lenovo E16 laptop that comes with built-in camera
```bash
# Install camera node
sudo apt-get install ros-iron-usb-cam
# start camera
ros2 run usb_cam usb_cam_node_exe
# Start monitor
ros2 topic hz /image_raw
```

#### Expected behavior
Reporting 30 FPS
![image](https://github.com/ros2/ros2cli/assets/8307900/cee39670-cfc8-4949-9798-4e05f9c5fc31)
*The misspelling are purposely done to identify my modification to the installed package

#### Actual behavior
Reporting 15-22 FPS or even lower
![image](https://github.com/ros2/ros2cli/assets/8307900/4a630d9a-0f0f-4f5a-9e51-210077387c11)
*The misspelling are purposely done to identify my modification to the installed package

#### Additional information
Affecting source code: [here](https://github.com/ros2/ros2cli/blob/rolling/ros2topic/ros2topic/verb/hz.py)
Following the hint from [this answers.ros](https://answers.ros.org/question/350753/ros2-topic-hz-provides-wrong-rate-for-larger-msgs/) discussions, I try modifying the QoS setting directly in the installed location `/opt/ros/iron/lib/python3.10/site-packages/ros2topic/verb/hz.py`. By changing the [qos profile](https://github.com/ros2/ros2cli/blob/rolling/ros2topic/ros2topic/verb/hz.py#L260C9-L260C32) line to `rclpy.qos.qos_profile_parameters`, it is possible for `ros2 topic hz` to report correct frequency.

| QoS Profile | qos_profile_sensor_data | qos_profile_parameters |
| ---- |  ----  | ----  |
| Reported rate | 15-22  | 30.077-30.083 |
| CPU usage  | 8-10% | 11-14% |

Expected value: 30 Hz as configured by the camera node

[...]

Issue #843, opened 2023-07-18
ros2 topic hz perform worse than c++ code, also worse than result in another docker container


Chenxintang:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu18.04
- Installation type:
  - source
- Version or commit hash:
  - humble
- DDS implementation:
  - FastDDS

#### Steps to reproduce issue

- run two docker container using the following docker-compose.yml
```yaml
container1:
    image: image_name
    network_mode: "host"
    cpuset: 2-9
    cpu_shares: 2048
    shm_size: 4G
    container_name: container1
    runtime: nvidia
    privileged: true
    environment:
      - NVIDIA_VISIBLE_DEVICES=all
    logging:
      options:
        max-size: "${DOCKER_LOGS_MAX_SIZE:-100M}"
```
```yaml
container2:
    image: image_name
    network_mode: "host"
    privileged: true
    cpuset: 12-17
    cpu_shares: 2048
    shm_size: 2G
    container_name: container2
    environment:
      - NVIDIA_VISIBLE_DEVICES=all
    runtime: nvidia
    logging:
      options:
        max-size: "${DOCKER_LOGS_MAX_SIZE:-100M}"
```
- publish `sensor_msgs::msg::PointCloud2` type message with frequency 10hz in container1
- run subscriber and process data c++ code in container2, record the frequency subscription callback called
- use `ros2 topic hz` tool follow the frequency of topic in both containers
#### Expected behavior
- In both containers, `ros2 topic hz` shows the same frequency which is around 10hz
- Log of c++ code in container2 shows that pointcloud2 messages arrive each 100ms 
#### Actual behavior
- In container1 terminal, `ros2 topic hz` shows frequency slightly under 10hz
![ros2 topic hz result in container1](https://github.com/ros2/ros2cli/assets/75031872/b8f54c40-38fe-4369-9776-73c7804406da)
- In container2 terminal, `ros2 topic hz` shows frequency highly under 10hz, verity from 7hz to 9hz
![ros2 topic hz result in container2](https://github.com/ros2/ros2cli/assets/75031872/1821d199-0aec-4c77-a576-7467bba9562c)
- In log of container2, c++ code receive message each 100ms stably
![c++ code log of container2](https://github.com/ros2/ros2cli/assets/75031872/b4711c88-5ed7-43e6-8d55-971b8d9a2ab6)

#### Additional information
Since code in both container are quite complicate, I can't paste the code here.
[...]

PR #970, opened 2025-02-07
Hz verb: Warn users about wrong reported freq

VictorLamoine:
I think enough users have wasted time with this tool and I hope to be the last one.

https://github.com/ros2/ros2cli/issues/871
https://github.com/ros2/ros2cli/issues/843
https://robotics.stackexchange.com/questions/95613/ros2-topic-hz-provides-wrong-rate-for-larger-msgs

Please notify me if that modification is ok for you and I'll fix the tests

### DIFF
--- a/ros2topic/ros2topic/verb/hz.py
+++ b/ros2topic/ros2topic/verb/hz.py
@@ -93,7 +93,7 @@ def main(args):
     else:
         filter_expr = None
     print('Reported frequency might not be accurate! https://github.com/ros2/ros2cli/issues/871')
-    
+
     with DirectNode(args) as node:
         _rostopic_hz(node.node, topics, window_size=args.window_size, filter_expr=filter_expr,
                      use_wtime=args.use_wtime)
--- a/ros2topic/ros2topic/verb/hz.py
+++ b/ros2topic/ros2topic/verb/hz.py
@@ -92,7 +92,8 @@ def main(args):
         filter_expr = expr_eval(args.filter_expr)
     else:
         filter_expr = None
-
+    print('Reported frequency might not be accurate! https://github.com/ros2/ros2cli/issues/871')
+    
     with DirectNode(args) as node:
         _rostopic_hz(node.node, topics, window_size=args.window_size, filter_expr=filter_expr,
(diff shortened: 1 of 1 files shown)


fujitatomoya (commented):
@VictorLamoine thanks for creating PR, trying to add the notice.

IMO this message does not really help the user experience. meaning, if that just might not be correct, how can we know that is correct this time or that time? without that capability, i guess this is just what we cannot trust at all, that means pretty much same with `do not use this feature for the topic statistics, because that might be just incorrect information`... for me.

instead, i suggest the followings, what do you think?

- adding the qos configuration option for `ros2 topic hz`. since QoS configuration affects the data reception on subscription with `ros2 topic hz`, besides this is also reported in https://github.com/ros2/ros2cli/issues/871 to solve the problem. we also can add this note on the command help message as well.
- i also would want to add more description about how `ros2 topic hz` works. it actually creates the subscription and receives the data, and then calculate the frequency. with my experience from streaming framework, some developers would think this is statistics from the sender(publisher) side.

VictorLamoine:
I'm all in for improving the tool but until the problems are fixed warning the users is the least we can do.

fujitatomoya:
@VictorLamoine could you add QoS improvement in this PR?

audrow:
I've added this to discuss to the next ROS 2 PMC meeting.

fujitatomoya:
https://github.com/ros2/ros2cli/pull/935 is adding the QoS argument for `ros2 topic hz`.

fujitatomoya:
> @VictorLamoine could you add QoS improvement in this PR?

this has been addressed by https://github.com/ros2/ros2cli/pull/970

VictorLamoine:
Sorry; I didn't have time to deal with this. Well test again and report!

ros-discourse:

https://discourse.ros.org/t/ros-pmc-minutes-for-2025-03-12/42493/1

fujitatomoya:
@VictorLamoine can you close this one, instead i will come up with documentation described in https://github.com/ros2/ros2cli/issues/843#issuecomment-2716401376. if you can review, that would be appreciated.

PR #973, opened 2025-02-08
Use target_link_libraries instead of ament_target_dependencies

sloretz:
This shows one weakness of `target_link_libraries`: the target names are not automatically known. `ament_cmake` packages create a `_TARGETS` variable,  so this CL makes the new `ament_cmake` package script use that. There might be cases where a new package depending on a plain cmake library would have worked before with `ament_target_dependencies()`, but does not work with this change.

I think using pure CMake here is worth it because we can point users to CMake docs instead of our own, but I'm open to feedback.

https://github.com/ament/ament_cmake/issues/292

### DIFF
--- a/ros2lifecycle_test_fixtures/CMakeLists.txt
+++ b/ros2lifecycle_test_fixtures/CMakeLists.txt
@@ -19,8 +19,8 @@ find_package(rclcpp REQUIRED)
 add_executable(simple_lifecycle_node
   src/simple_lifecycle_node.cpp)
 
-ament_target_dependencies(simple_lifecycle_node
-  "rclcpp_lifecycle")
+target_link_libraries(simple_lifecycle_node PRIVATE
+  rclcpp_lifecycle::rclcpp_lifecycle)
 
 install(TARGETS
   simple_lifecycle_node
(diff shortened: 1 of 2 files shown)


fujitatomoya (commented):
> I think using pure CMake here is worth it because we can point users to CMake docs instead of our own, but I'm open to feedback.

i think this is okay, but we have lots of documentation that recommends `ament_target_dependencies` e.g https://docs.ros.org/en/rolling/How-To-Guides/Ament-CMake-Documentation.html#linking-to-dependencies. those can also be updated in the future.

ahcorde:
Pulls: ros2/ros2cli#973
Gist: https://gist.githubusercontent.com/ahcorde/f15dae00f131e9fbd94d2fbfc17f6b18/raw/95f3d9b052978fef1bdb4c0ffa74d3b81d521875/ros2.repos
BUILD args:  --packages-above-and-dependencies ros2lifecycle_test_fixtures ros2pkg
TEST args:  --packages-above ros2lifecycle_test_fixtures ros2pkg
ROS Distro: rolling
Job: ci_launcher

ahcorde (approved):

ahcorde:
Pulls: ros2/ros2cli#973
Gist: https://gist.githubusercontent.com/ahcorde/39b6009cbfdc2d70cd2e077bdcc0953a/raw/95f3d9b052978fef1bdb4c0ffa74d3b81d521875/ros2.repos
BUILD args:  --packages-above-and-dependencies ros2lifecycle_test_fixtures ros2pkg
TEST args:  --packages-above ros2lifecycle_test_fixtures ros2pkg
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15192
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22530/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16777/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/2051/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/23322/)

PR #982, opened 2025-02-20
* ros2pkg/resource/cpp/main.cpp.em: Use modern C++17 syntax.

shynur:
According to <https://github.com/shynur/ros2cli/blob/889fba5171393aae9b9577a1dc2b8cea8fc39a3f/ros2pkg/ros2pkg/resource/cmake/CMakeLists.txt.em#L29>, we use C++17 or higher standard.

- `[[maybe_unused]]` is a C++17 attribute, which is more portable.
- `return 0;` is unnecessary in C++, which can be removed to save space.

### DIFF
--- a/ros2pkg/ros2pkg/resource/cpp/main.cpp.em
+++ b/ros2pkg/ros2pkg/resource/cpp/main.cpp.em
@@ -1,10 +1,6 @@
 #include <cstdio>
 
-int main(int argc, char ** argv)
+int main([[maybe_unused]] int argc, [[maybe_unused]] char ** argv)
 {
-  (void) argc;
-  (void) argv;
-
   printf("hello world @(package_name) package\n");
-  return 0;
 }


ros2pkg/ros2pkg/resource/cpp/main.cpp.em
```diff
@@ -1,10 +1,6 @@
 #include <cstdio>
 
-int main(int argc, char ** argv)
+int main(int argc [[maybe_unused]], char ** argv [[maybe_unused]])
```
christophebedard: I think this makes sense.

However, `[[maybe_unused]]` should ideally go _before_ the type and name, see https://en.cppreference.com/w/cpp/language/attributes/maybe_unused.

shynur: > However, `[[maybe_unused]]` should ideally go *before* the type and name

See <https://en.cppreference.com/w/cpp/language/attributes>:

> In declarations, attributes may appear both before the whole declaration and *directly after the name* of the entity that is declared, in which case they are combined.

Putting the attribute and name together looks more intuitive IMO.

Anyway, you can edit the PR as you wish ;-)

christophebedard: Yes, it can be before or after. I've personally seen both, but I've seen it mostly being used before and that makes more sense to me. That also makes more sense to me when it's used for a variable (e.g., inside a function).

shynur: I will edit it... done

christophebedard:
Could you sign off your second commit, or squash both commits into a single signed-off commit?

shynur:
> Could you sign off your second commit, or squash both commits into a single signed-off commit?

It's a single signed-off commit now.  Please check.

christophebedard:
Pulls: ros2/ros2cli#982
Gist: https://gist.githubusercontent.com/christophebedard/9710dc2790a77933d0b98287cca5f0c5/raw/c1636a524072fef93e25f181f2320d80c835ea96/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15238
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22583/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16826/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/2098/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/23385/)

christophebedard (approved):
Thanks for the PR!

PR #987, opened 2025-03-13
Adjust topic hz and bw command description.

fujitatomoya:
part of https://github.com/ros2/ros2cli/issues/843#issuecomment-2716401376

replaces https://github.com/ros2/ros2cli/pull/970

### DIFF
--- a/ros2topic/ros2topic/verb/bw.py
+++ b/ros2topic/ros2topic/verb/bw.py
@@ -62,10 +62,10 @@ class BwVerb(VerbExtension):
 
     def add_arguments(self, parser, cli_name):
         parser.description = (
-            "Display bandwidth used by topic.\n\n"
-            "note:\n"
-            "  This bandwidth reflects the receiving rate on subscription, "
-            "which might be affected by platform resources and QoS configuration, "
+            'Display bandwidth used by topic.\n\n'
+            'note:\n'
+            '  This bandwidth reflects the receiving rate on subscription, '
+            'which might be affected by platform resources and QoS configuration, '
             "and may not exactly match the publisher's bandwidth."
         )
         arg = parser.add_argument(
--- a/ros2topic/ros2topic/verb/bw.py
+++ b/ros2topic/ros2topic/verb/bw.py
@@ -61,6 +61,13 @@ class BwVerb(VerbExtension):
     """Display bandwidth used by topic."""
 
     def add_arguments(self, parser, cli_name):
+        parser.description = (
+            "Display bandwidth used by topic.\n\n"
+            "note:\n"
+            "  This bandwidth reflects the receiving rate on subscription, "
+            "which might be affected by platform resources and QoS configuration, "
+            "and may not exactly match the publisher rate."
+        )
         arg = parser.add_argument(
             'topic',
             help='Topic name to monitor for bandwidth utilization')
--- a/ros2topic/ros2topic/verb/bw.py
+++ b/ros2topic/ros2topic/verb/bw.py
@@ -66,7 +66,7 @@ class BwVerb(VerbExtension):
             "note:\n"
             "  This bandwidth reflects the receiving rate on subscription, "
             "which might be affected by platform resources and QoS configuration, "
-            "and may not exactly match the publisher rate."
+            "and may not exactly match the publisher's bandwidth."
         )
         arg = parser.add_argument(
             'topic',
--- a/ros2topic/ros2topic/verb/hz.py
+++ b/ros2topic/ros2topic/verb/hz.py
@@ -57,10 +57,10 @@ class HzVerb(VerbExtension):
 
     def add_arguments(self, parser, cli_name):
         parser.description = (
-            "Print the average receiving rate to screen.\n\n"
-            "note:\n"
-            "  This rate reflects the receiving rate on subscription, "
-            "which might be affected by platform resources and QoS configuration, "
+            'Print the average receiving rate to screen.\n\n'
+            'note:\n'
+            '  This rate reflects the receiving rate on subscription, '
+            'which might be affected by platform resources and QoS configuration, '
             "and may not exactly match the publisher's rate."
         )
         arg = parser.add_argument(
--- a/ros2topic/ros2topic/verb/hz.py
+++ b/ros2topic/ros2topic/verb/hz.py
@@ -53,9 +53,16 @@ DEFAULT_WINDOW_SIZE = 10000
 
 
 class HzVerb(VerbExtension):
-    """Print the average publishing rate to screen."""
+    """Print the average receiving rate to screen."""
 
     def add_arguments(self, parser, cli_name):
+        parser.description = (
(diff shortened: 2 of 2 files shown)


ros2topic/ros2topic/verb/bw.py
```diff
@@ -61,6 +61,13 @@ class BwVerb(VerbExtension):
     """Display bandwidth used by topic."""
 
     def add_arguments(self, parser, cli_name):
+        parser.description = (
```
fujitatomoya, line 64:
```bash
root@tomoyafujita:~/ros2_ws/colcon_ws# ros2 topic bw -h
usage: ros2 topic bw [-h] [--window WINDOW] [--spin-time SPIN_TIME] [-s] topic

Display bandwidth used by topic.

note:
  This bandwidth reflects the receiving rate on subscription, which might be affected by platform resources and QoS configuration, and may not exactly match the publisher rate.

positional arguments:
  topic                 Topic name to monitor for bandwidth utilization

options:
  --window WINDOW, -w WINDOW
                        maximum window size, in # of messages, for calculating rate (default: 100)
  --spin-time SPIN_TIME
                        Spin time in seconds to wait for discovery (only applies when not using an already running
                        daemon)
```

ros2topic/ros2topic/verb/hz.py
```diff
@@ -53,9 +53,16 @@
...
+    """Print the average receiving rate to screen."""
 
     def add_arguments(self, parser, cli_name):
+        parser.description = (
+            "Print the average receiving rate to screen.\n\n"
+            "note:\n"
+            "  This rate reflects the receiving rate on subscription, "
+            "which might be affected by platform resources and QoS configuration, "
+            "and may not exactly match the publisher rate."
+        )
```
fujitatomoya:
```bash
root@tomoyafujita:~/ros2_ws/colcon_ws# ros2 topic hz -h
usage: ros2 topic hz [-h] [--window WINDOW] [--filter EXPR] [--wall-time] [--spin-time SPIN_TIME] [-s]
                     topic_name [topic_name ...]

Print the average receiving rate to screen.

note:
  This rate reflects the receiving rate on subscription, which might be affected by platform resources and QoS configuration, and may not exactly match the publisher rate.

positional arguments:
  topic_name            Names of the ROS topic to listen to (e.g. '/chatter')

options:
  --window WINDOW, -w WINDOW
                        window size, in # of messages, for calculating rate (default: 10000)
  --filter EXPR         only measure messages matching the specified Python expression
  --wall-time           calculates rate using wall time which can be helpful when clock is not published during
                        simulation
  --spin-time SPIN_TIME
                        Spin time in seconds to wait for discovery (only applies when not using an already running
                        daemon)
```

fujitatomoya (commented):
adding the note for `ros2 topic hz/bw` command that these statistics are what we can get on this subscription, not expected publisher rate or bandwidth.

fujitatomoya:
@mjcarroll @ahcorde @christophebedard @MichaelOrlov what do you think? at least, this could give some hint for users?

VictorLamoine:
This is great thank you 👍🏼 
Users will try to investigate their QoS configuration first instead of checking their publisher code.

jmachowinski:
I would also add a warning before the start of the rate output e.g.:
```shell
$ ros2 topic hz /clock
Warning: This rate reflects the receiving rate on subscription, which might be affected by platform resources and QoS configuration, and may not exactly match the publisher rate. This is especially true for big or high frequency messages.
average rate: 100.006
        min: 0.010s max: 0.010s std dev: 0.00008s window: 102
average rate: 100.004
        min: 0.010s max: 0.010s std dev: 0.00008s window: 203

```

fujitatomoya:
> I would also add a warning before the start of the rate output e.g.:

i like this one, i am inclined to do this. but one minor concern is that changing output format/message would break user tools, sometime developers have some tools to grep the output from ros2cli... but i think this is not the case.

ahcorde (approved):
I'm fine with the message in the help message. Anyhow I will wait for other opinions

fujitatomoya:
yeah it would be much better if we can address performance related enhancement like deserialization and executor.

btw, it is also important to know that statistics are different from sender and receiver. means, that if we want the precise statistics like this, we would need to provide the API/Interface to get the statistics on publisher and subscription each. (saying no matter we optimize the performance, there would be platform resource issue, network stability problem on receiver side.) i think that this is common for the use cases that rely on statistics such as live streaming and camera framework.
this additional note is meant to let the user know `this is what you can see on your subscription, not publisher statistics`.

fujitatomoya:
Pulls: ros2/ros2cli#987
Gist: https://gist.githubusercontent.com/fujitatomoya/18dd221066730bb7a47229a6a51d5288/raw/9a015b9681ac5414e1753157e00f1800880141c3/ros2.repos
BUILD args:  --packages-above-and-dependencies ros2topic
TEST args:  --packages-above ros2topic
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15374
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22729/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16971/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/2246/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/23604/)

MichaelOrlov (commented):
@fujitatomoya I think the statement about 
> Display bandwidth used by the topic

It is a bit misleading and unclear. Bandwidth usually refers to the number of bytes per second. However, here, we are talking about the messages rate without considering the number of bytes transferred.

Also, I would add a note that this command will subscribe to the topic and receive messages from subscription during the measurement, which can affect overall system performance.

MichaelOrlov:
Please disregard my concern above `Bandwidth` statement. I missed that it is actually in the **BW** (Bandwidth) command and not in the **Hz**.

fujitatomoya:
@MichaelOrlov no worries, thanks for the comments.

fujitatomoya:
Pulls: ros2/ros2cli#987
Gist: https://gist.githubusercontent.com/fujitatomoya/4bcb5cecf35c77b47388d1f791a553ff/raw/9a015b9681ac5414e1753157e00f1800880141c3/ros2.repos
BUILD args:  --packages-above-and-dependencies ros2topic
TEST args:  --packages-above ros2topic
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15382
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22737/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16979/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/2256/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/23612/)

fujitatomoya:
@Mergifyio backport humble jazzy

PR #1003, opened 2025-04-04
Fix 'ros2 topic hz'

christophebedard:
Fixes #1002

Follow-up to #998, which backported a commit from Rolling to Jazzy that included unrelated changes that Jazzy doesn't have:

1. `_rostopic_hz()` does not expect the `qos_args` arg, since Jazzy doesn't have #935
    - This is the original bug report from #1002
2. `args.topic_name` contains a single topic name, since Jazzy doesn't have #929, which also means that `_rostopic_hz()` only expects a single topic name, and not a list

Run a node:

```console
$ ros2 run examples_rclcpp_minimal_publisher publisher_member_function
```

Before:

```console
$ ros2 topic hz /topic
Traceback (most recent call last):
  File "/home/christophe.bedard/ros2_ws_jazzy/install/ros2cli/bin/ros2", line 33, in <module>
    sys.exit(load_entry_point('ros2cli', 'console_scripts', 'ros2')())
  File "/home/christophe.bedard/ros2_ws_jazzy/build/ros2cli/ros2cli/cli.py", line 91, in main
    rc = extension.main(parser=parser, args=args)
  File "/home/christophe.bedard/ros2_ws_jazzy/build/ros2topic/ros2topic/command/topic.py", line 41, in main
    return extension.main(args=args)
  File "/home/christophe.bedard/ros2_ws_jazzy/build/ros2topic/ros2topic/verb/hz.py", line 87, in main
    return main(args)
  File "/home/christophe.bedard/ros2_ws_jazzy/build/ros2topic/ros2topic/verb/hz.py", line 152, in main
    _rostopic_hz(node.node, topics, qos_args=args, window_size=args.window_size,
TypeError: _rostopic_hz() got an unexpected keyword argument 'qos_args'
```

If I only fix (1) and not (2), the topic name string gets split into single characters:

```console
$ ros2 topic hz topic --filter 'True'
WARNING: topic [t] does not appear to be published yet
```

After:

```console
$ ros2 topic hz topic --filter 'True'
average rate: 2.000
        min: 0.500s max: 0.500s std dev: 0.00011s window: 3
average rate: 2.000
        min: 0.499s max: 0.501s std dev: 0.00058s window: 6
```

Issue #1002, opened 2025-04-04
`_rostopic_hz() got an unexpected keyword argument 'qos_args'`


mjforan:
### Operating System:

Linux 7789fa2838e0 6.6.74+rpt-rpi-2712 #1 SMP PREEMPT Debian 1:6.6.74-1+rpt1 (2025-01-27) aarch64 GNU/Linux

### ROS version or commit hash:

jazzy

### RMW implementation (if applicable):

rmw_fastrtps_cpp

### 'ros2 doctor --report' output

```
   NETWORK CONFIGURATION
inet         : 127.0.0.1
inet4        : ['127.0.0.1']
inet6        : ['::1']
netmask      : 255.0.0.0
device       : lo
flags        : 73<LOOPBACK,RUNNING,UP>
mtu          : 65536
inet         : 172.17.0.3
inet4        : ['172.17.0.3']
ether        : ea:a8:23:8a:0d:c2
netmask      : 255.255.0.0
device       : eth0
flags        : 4163<MULTICAST,RUNNING,UP,BROADCAST>
mtu          : 1500
broadcast    : 172.17.255.255

   PACKAGE VERSIONS
rqt_bag_plugins                           : latest=1.5.4, local=1.5.4
rosbag2                                   : latest=0.26.6, local=0.26.6
rosbag2_storage_default_plugins           : latest=0.26.6, local=0.26.6
rosbag2_storage_mcap                      : latest=0.26.6, local=0.26.6
rosbag2_compression_zstd                  : latest=0.26.6, local=0.26.6
mcap_vendor                               : latest=0.26.6, local=0.26.6
zstd_vendor                               : latest=0.26.6, local=0.26.6
vision_msgs_rviz_plugins                  : latest=4.1.1, local=4.1.1
rviz_visual_testing_framework             : latest=14.1.8, local=14.1.8
rviz2                                     : latest=14.1.8, local=14.1.8
rviz_default_plugins                      : latest=14.1.8, local=14.1.8
rviz_common                               : latest=14.1.8, local=14.1.8
rqt_bag                                   : latest=1.5.4, local=1.5.4
rosbag2_examples_py                       : latest=0.26.6, local=0.26.6
launch_testing_examples                   : latest=0.19.5, local=0.19.5
ros2bag                                   : latest=0.26.6, local=0.26.6
rosbag2_py                                : latest=0.26.6, local=0.26.6
rosbag2_transport                         : latest=0.26.6, local=0.26.6
[...]

PR #998, opened 2025-04-02
jazzy: Backport Patch CVE-2024-42002

mjcarroll:

PR #1004, opened 2025-04-04
Revert "jazzy: Backport Patch CVE-2024-42002"

christophebedard:
Reverts ros2/ros2cli#998

See https://github.com/ros2/ros2cli/issues/1002. I attempted to fix some issues in https://github.com/ros2/ros2cli/pull/1003, but for now just revert it.

The same change was also reverted in Rolling: #1000


christophebedard:
This reverts the very last commit, and it looks like CI wasn't run when the Rolling PR was reverted (#1000), so I'll just merge this.

fujitatomoya (commented):
@christophebedard thanks, one step behind. but i agree with you, this is deeper than it looks...

PR #1058, opened 2025-06-17
Make sure to install py.typed files

christophebedard:
## Description

Similar to https://github.com/ros2/launch/pull/886, see https://github.com/ros2/launch/issues/885.

This will make `mypy` (e.g., through `ament_mypy`) be stricter about type annotation for downstream users of the packages with a `py.typed` file, such as `ros2cli`. For example, see https://github.com/ros2/ros2_tracing/issues/160#issuecomment-2971204451. This could be considered a breaking change. This is probably only a concern for `ros2cli`, since the other packages are leaf-y. However, the `ros2cli` API is pretty simple, so this shouldn't be a big headache.

No.

### Additional Information

It's important to test all core packages downstream of these.

### DIFF
--- a/ros2action/setup.py
+++ b/ros2action/setup.py
@@ -12,6 +12,7 @@ setup(
         ('share/ament_index/resource_index/packages',
             ['resource/' + package_name]),
     ],
+    package_data={'': ['py.typed']},
     install_requires=['ros2cli'],
     zip_safe=True,
     author='Jacob Perron',
--- a/ros2cli/setup.py
+++ b/ros2cli/setup.py
@@ -21,6 +21,7 @@ setup(
             'completion/ros2-argcomplete.zsh'
         ]),
     ],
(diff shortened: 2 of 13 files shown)


christophebedard:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/24012/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/18083/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/3414/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/24653/)

fujitatomoya (approved):
lgtm with green CI.

@christophebedard probably we need to restart the CI?

christophebedard:
I found the jobs that were automatically triggered after CI failures. The only test failures are known (including the Rpr job), so all good.

PR #1080, opened 2025-07-17
Added ament_cmake_python package creation

bismuth01:
## Description

Working on the feature requested at #739 . Added a template argument and creation of the required folder structure for `--build-type cmake --template-name python`

Now user can get a starting boilerplate where both python and cmake can be used in the same package.

No, everything was reviewed, written and tested by hand.

### Additional Information

It's still unclear how the `library-name` and `node-name` arguments will interact with this template given the ambiguity of python and cmake.


ahcorde (changes requested):
Flake8 is failing 
https://build.ros2.org/job/Rpr__ros2cli__ubuntu_noble_amd64/234/testReport/junit/ros2pkg.ros2pkg.test/test_flake8/test_flake8/

PR #1105, opened 2025-09-23
Fix ros2 topic pub --stdin

wentasah:
## Description

Running the following command currently fails:

    ros2 topic pub --stdin /test std_msgs/Bool <<< "{data: true}"

The error is:

      File "/python3.12-ros-jazzy-ros2cli-0.32.6-r1/bin/ros2", line 34, in <module>
        sys.exit(load_entry_point('ros2cli==0.32.6', 'console_scripts', 'ros2')())
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/ros-env/lib/python3.12/site-packages/ros2cli/cli.py", line 91, in main
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/ros-env/lib/python3.12/site-packages/ros2topic/command/topic.py", line 41, in main
               ^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/ros-env/lib/python3.12/site-packages/ros2topic/verb/pub.py", line 117, in main
        return main(args)
               ^^^^^^^^^^
      File "/ros-env/lib/python3.12/site-packages/ros2topic/verb/pub.py", line 138, in main
        return publisher(
               ^^^^^^^^^^
      File "/ros-env/lib/python3.12/site-packages/ros2topic/verb/pub.py", line 174, in publisher
        if '^J' in values:
           ^^^^^^^^^^^^^^
    TypeError: a bytes-like object is required, not 'str'

We fix it by decoding stdin to utf8 string to match the type of values passed via a command line argument.

yes

no

### DIFF
--- a/ros2topic/ros2topic/verb/pub.py
+++ b/ros2topic/ros2topic/verb/pub.py
@@ -138,7 +138,7 @@ def main(args):
         times = 1
 
     if args.stdin:
-        values = collect_stdin()
+        values = collect_stdin().decode()
     else:
         values = args.values
 


fujitatomoya (approved):
This allows subsequent code to work with a string instead of bytes, which is often necessary for text processing. test would be ideal to add somewhere like https://github.com/ros2/ros2cli/blob/rolling/ros2topic/test/test_echo_pub.py

wentasah:
I'm trying to write the test, but I have no idea how to pass something to process stdin. I see there is a `ProcessStdin` event, which is handled by the `ExecuteLocal` action, but I din't find a way to generate the event from the test. And from looking at `ExecuteLocal.__on_process_stdin`, I doubt that it can do anything useful here. I also tried to follow the documentation, which says that `ExecuteProcess` passes many arguments to `Popen`, but it seems not to be the case for `stdin`.

Perhaps, I'd be able to write a test without the complex launch machinery by using plain Python's `Popen`. Would that be acceptable?

PR #1151, opened 2025-12-21
Add fzf-based interactive selection to ros2cli commands

tonynajjar:
## Description

This PR adds optional fuzzy search functionality to ros2cli commands using fzf, inspired by the [ros2-aliases](https://github.com/tonynajjar/ros2-aliases) project. When arguments are omitted from commands, an interactive fzf selector is launched instead of throwing an error.

**Commands enhanced:**
- `ros2 topic echo` - Select topic interactively
- `ros2 topic hz` - Select topic interactively
- `ros2 topic info` - Select topic interactively
- `ros2 topic bw` - Select topic interactively
- `ros2 topic type` - Select topic interactively
- `ros2 node info` - Select node interactively
- `ros2 param get` - Cascading selection (node → parameter)
- `ros2 interface show` - Select interface interactively
- `ros2 run ` - Cascading selection (package → executable)

**Implementation:**
- Added `interactive_select()` helper function in `ros2cli/helpers.py`
- Made positional arguments optional (`nargs='?'`) in affected verbs
- Added fzf as an execution dependency in `ros2cli/package.xml`
- Maintains 100% backward compatibility - all existing command usage still works

## Is this user-facing behavior change?

**Yes.** Users can now omit arguments and use fuzzy search instead:

**Before:**
```bash
ros2 topic echo /my_topic_name  # Must know exact topic name
```

**After:**
```bash
ros2 topic echo                  # Launches fzf to select from available topics
ros2 topic echo /my_topic_name   # Still works as before
```

If fzf is not installed, users get a clear error message directing them to install via rosdep.

## Did you use Generative AI?

Yes, GitHub Copilot was used

## Additional Information

- **Requires fzf**: Added as dependency in package.xml but dependent on https://github.com/ros/rosdistro/pull/49148
- **No breaking changes**: All existing CLI usage remains unchanged
- **Graceful degradation**: Clear error message if fzf is missing


tonynajjar:
This PR still needs some testing and refinement but before I iterate further I'd like to know if you would merge this idea in. Thanks
@ahcorde @fujitatomoya 
You can see a gif demo in https://github.com/tonynajjar/ros2-aliases

ros2cli/ros2cli/helpers.py
```diff
@@ -133,3 +136,61 @@ def get_rmw_additional_env(rmw_implementation: str) -> Dict[str, str]:
         return {
             'RMW_IMPLEMENTATION': rmw_implementation,
         }
+
+
+def interactive_select(
+    items: List[str],
+    prompt: str = 'Select an item:'
+) -> Optional[str]:
```
fujitatomoya, line 144: does it miss tty detection? no check for whether stdin/stdout is a tty? i think this will cause scripts and CI/CD pipelines to hang, if it meets this case. related to this issue, fzf reads keyboard input from /dev/tty, not stdin. however, when ros2cli commands are piped or redirected, fzf may not be able to access the tty properly?

the followings are pseudo code,

```python
    # Check if we're in an interactive terminal
    if not sys.stdin.isatty() or not sys.stdout.isatty():
        print('Error: Interactive selection requires a TTY terminal.', 
              file=sys.stderr)
        return None
```

and

```python
try:
    # fzf needs direct access to /dev/tty for keyboard input
    tty = open('/dev/tty', 'r')
    process = subprocess.Popen(
        ['fzf', '--prompt', prompt + ' ', '--height', '40%', '--reverse'],
        stdin=subprocess.PIPE,
        stdout=subprocess.PIPE,
        stderr=tty,  # fzf uses stderr for display
        text=True
    )
    # ...
finally:
    tty.close()
```

ros2cli/ros2cli/helpers.py
```diff
@@ -133,3 +136,61 @@ def get_rmw_additional_env(rmw_implementation: str) -> Dict[str, str]:
...
+        # Launch fzf with items as input - using direct TTY access
+        process = subprocess.Popen(
+            ['fzf', '--prompt', prompt + ' ', '--height', '40%', '--reverse'],
+            stdin=subprocess.PIPE,
+            stdout=subprocess.PIPE,
+            text=True
+        )
+
+        # Send items to fzf
+        stdout, _ = process.communicate(input='\n'.join(items))
```
fujitatomoya: We do not need to have any signal control here when opening the child process? If the user presses Ctrl+C, this may leave the terminal in a bad state or not clean up properly... maybe we can add `try/except` statement to catch the KeyboardInterrupt to call `process.terminate()`?

```python
import signal

try:
    stdout, _ = process.communicate(input='\n'.join(items))
except KeyboardInterrupt:
    process.terminate()
    process.wait()
    return None
```

ros2cli/ros2cli/helpers.py
```diff
@@ -133,3 +136,61 @@ def get_rmw_additional_env(rmw_implementation: str) -> Dict[str, str]:
...
+            timeout=1
+        )
+        if result.returncode != 0:
+            raise FileNotFoundError()
+    except (FileNotFoundError, subprocess.TimeoutExpired):
+        print(
+            'Error: fzf is not installed but is a dependency for this package. You can install it with rosdep',
+            file=sys.stderr
+        )
+        return None
```
fujitatomoya: i believe this is redundant. i would do the following instead to check `fzf` availability.

```python
import shutil

if shutil.which('fzf') is None:
    print('Error: fzf is not installed...', file=sys.stderr)
    return None
```

ros2cli/ros2cli/helpers.py
```diff
@@ -133,3 +136,61 @@ def get_rmw_additional_env(rmw_implementation: str) -> Dict[str, str]:
...
+        if process.returncode != 0:
+            return None
+
+        # Return selected item (strip newline)
+        selected = stdout.strip()
+        return selected if selected else None
+
+    except Exception as e:
+        print(f'Error during interactive selection: {e}', file=sys.stderr)
+        return None
```
fujitatomoya: Catching all exceptions hides bugs and makes debugging harder.


```suggestion
    except (OSError, subprocess.SubprocessError) as e:
        print(f'Error during interactive selection: {e}', file=sys.stderr)
        return None
```

ros2cli/ros2cli/helpers.py
fujitatomoya: besides that, i would add the timeout just in case if fzf hangs for some reason, the entire command hangs indefinitely.

fujitatomoya (commented):
@tonynajjar generally i like this enhancement, thanks 🚀 

there are several comments that i think we need to address but overall i think this is gonna be better user-experience for ROS 2 users and developers.

we already have type completion via shell command prompt but tab completion requires us to know the prefix and type sequentially from the beginning, while fzf lets us match any part of the name in any order (e.g., typing "wrist depth" finds  /arm/wrist_camera/depth/image_raw). fzf also provides visual browsing of all available options with real-time filtering, making it ideal for exploring unfamiliar systems or when you only remember fragments of a name.

before fixing up my comments, let's hear out more feedback on this 👍

tonynajjar:
Thanks for the feedback @fujitatomoya . Like you said I will start working on your comments once I get another opinion on the general idea, to make sure this can ultimately get merged in

tonynajjar:
Maybe @ahcorde you can give your general opinion on the feature and I can continue to iterate with @fujitatomoya? Thanks!

tonynajjar:
friendly ping @ahcorde

ros2cli/ros2cli/helpers.py
tonynajjar: > We do not need to have any signal control here when opening the child process? If the user presses Ctrl+C, this may leave the terminal in a bad state or not clean up properly... 

```
root@dexory:/ws# ros2 topic echo
No topic selected
```
This is what it looks like currently after CTRL-C, there doesn't seem to be an issue.


>  i would add the timeout

hmm, what would you set the timeout value to? I feel like it might be bad user experience if after e.g. 30 seconds, the command is just killed. I don't see many reasons for a hang to happen but in the rare occasion that it does, one can probably still workaround it with CTRL-D or frankly just kill the terminal session and retry.

ros2cli/ros2cli/helpers.py
tonynajjar: Theoretically this shouldn't be a problem if fzf is not called in the CI right? I can't immediately think of why it would but anyway I don't mind the protection; I added the "Check if we're in an interactive terminal" part but the second part that opens `/dev/tty` caused it to hang.

tonynajjar: > but the second part that opens /dev/tty caused it to hang.

ok but something like that seems to be necessary indeed because while testing I got an error and it did leave the terminal in a "bad state" :smile: . I look into it

tonynajjar:
@fujitatomoya I did a second round of fixes and am pretty happy with it so far but I invite you to test it out and find some bugs.

I'd like to note that there are probably more verbs that can benefit from an fzf integration like `ros2 param set` or `ros2 topic pub` but these require an additional input from the user in the end so slightly more complication (still manageable) - I would however get this first version in and I'm sure if people find it useful then they (or me) will be quick to contribute more fzf integrations

ros2cli/ros2cli/helpers.py
tonynajjar: Okay I found a way to recover safely

```
        except KeyboardInterrupt:
            # Handle Ctrl+C gracefully to avoid leaving terminal in bad state
            process.terminate()
            process.wait()
            # Reset terminal to normal mode after fzf interruption
            subprocess.run(['stty', 'sane'], check=False)
            return None
        finally:
            # Ensure terminal is restored even if an exception occurs
            if process.poll() is None:
                process.terminate()
                process.wait()
            subprocess.run(['stty', 'sane'], check=False)
```

ros2interface/ros2interface/verb/show.py
```diff
@@ -167,8 +169,8 @@ def __call__(self, parser, namespace, values, option_string=None):
             if sys.stdin.isatty():
                 parser.error('expected stdin pipe')
             values = sys.stdin.readline().strip()
-        if not values:
-            parser.error('the passed value is empty')
+            if not values:
+                parser.error('the passed value is empty')
```
fujitatomoya, line 173: this changes the behavior in this if statement? is this intentional change?

fujitatomoya (approved):
lgtm with green CI. just checking one last comment.

fujitatomoya:
@ahcorde @InvincibleRMC i would like to have another approval on this before merge, can you take a look at this?

ros2topic/ros2topic/verb/pub.py
```diff
@@ -28,6 +28,7 @@
 from ros2cli.qos import add_qos_arguments
 from ros2cli.qos import profile_configure_short_keys
 
+from ros2topic.api import get_topic_names
```
InvincibleRMC: Is this an unused import?

ros2interface/ros2interface/api/__init__.py
```diff
@@ -52,3 +52,20 @@ def interface_to_yaml(identifier):
         instance = interface()
 
     return message_to_yaml(instance)
+
+
+def get_all_interface_names():
```
InvincibleRMC:
```suggestion
def get_all_interface_names() -> list[str]:
```

ros2cli/ros2cli/helpers.py
```diff
@@ -133,3 +137,70 @@ def get_rmw_additional_env(rmw_implementation: str) -> Dict[str, str]:
         return {
             'RMW_IMPLEMENTATION': rmw_implementation,
         }
+
+
+def interactive_select(
+    items: List[str],
```
InvincibleRMC:
```suggestion
    items: list[str],
```

ros2cli/ros2cli/helpers.py
```diff
@@ -16,10 +16,14 @@
 import functools
 import inspect
 import os
+import shutil
+import subprocess
 import sys
 import time
 
 from typing import Dict
+from typing import List
```
InvincibleRMC:
```suggestion
```

InvincibleRMC (dismissed):
Some small questions

ros2interface/ros2interface/verb/show.py
tonynajjar: yeah it's intentional, it's to allow passing nothing to `ros2 interface show` without an error - so that fzf is triggered

tonynajjar:
Thanks for the second review. Some tests are failing because before this PR, verbs are expected to fail if no argument is given but now it's different. I will take a look at this next, hopefully pretty soon

tonynajjar:
> Some tests are failing because before this PR, verbs are expected to fail if no argument is given

Well, I really just removed these 2 tests because they are simply invalid at this point. Tests pass now :D

kscottz:
This is super cool. THANKS!

Once this gets merged I want to share it on social media. Also consider doing a ROS Discourse post.

tonynajjar:
I'm glad you like it! Don't worry, I never intended it to go unnoticed ;)

tonynajjar:
I see there is a conflict now - I'll fix it soon but please after that if there is nothing else blocking let's merge it before I get more conflicts caused by other newer PRs merged first (first come first served :wink:)

fujitatomoya:
@tonynajjar yeah it is likely to happen... 😓 after resolving conflicts, i will start the CI. please let me know.

mjcarroll:
Just another comment that this is great! I use `fzf` pretty extensively, so it would be nearly-muscle-memory to have it in the ros2 command line tools as well.

tonynajjar:
- I fixed the conflict. @fujitatomoya Previously `ros2 param get <some-string>` launched fzf to select the parameter but since your PR, this syntax is used to get the parameter from all the nodes. I left your implementation intact so now only `ros2 param get` will launch fzf to select both the node and the param.

- I added fzf-support for `ros2 run` - not sure why I had skipped it before. There are more opportunities to use fzf but I think I caught the most important ones

Let's see what CI has to say

tonynajjar:
CI passed @fujitatomoya @mjcarroll

fujitatomoya:
Pulls: ros2/ros2cli#1151
Gist: https://gist.githubusercontent.com/fujitatomoya/dcfb635a10e6856b1fb5453efb8d880c/raw/744f7f32509a1b04fc6822cf969c6d659d7eee6c/ros2.repos
BUILD args:  --packages-above-and-dependencies ros2cli ros2interface ros2param ros2run ros2service ros2topic
TEST args:  --packages-above ros2cli ros2interface ros2param ros2run ros2service ros2topic
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/18036
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/27206/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/20548/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/6976/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26500/)

fujitatomoya:
Pulls: ros2/ros2cli#1151
Gist: https://gist.githubusercontent.com/fujitatomoya/100b911488a5723f3aa42adc5448f806/raw/744f7f32509a1b04fc6822cf969c6d659d7eee6c/ros2.repos
BUILD args:  --packages-above-and-dependencies ros2cli ros2interface ros2param ros2run ros2service ros2topic ros2node
TEST args:  --packages-above ros2cli ros2interface ros2param ros2run ros2service ros2topic ros2node
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/18037
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/27207/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/20549/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/6977/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26501/)

tonynajjar:
Let me know if there is anything to do about the rhel build failing

fujitatomoya:
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/6979/)

fujitatomoya:
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/6981/)

tonynajjar:
Green!

tonynajjar:
@kscottz Here is a updated demo video

https://github.com/user-attachments/assets/558ec912-36c7-497e-9888-b0e38af822ca

ros-discourse:
This pull request has been mentioned on **Open Robotics Discourse**. There might be relevant details there:

https://discourse.openrobotics.org/t/ros2cli-meet-fzf/52097/1

ros-discourse:
This pull request has been mentioned on **Open Robotics Discourse**. There might be relevant details there:

https://discourse.openrobotics.org/t/ros2cli-meet-fzf/52097/3

ros-discourse:
This pull request has been mentioned on **Open Robotics Discourse**. There might be relevant details there:

https://discourse.openrobotics.org/t/ros2cli-meet-fzf/52097/5

szobov:
Hey @fujitatomoya,
Can we trigger back-merge into Jazzy?

tonynajjar:
@szobov unfortunately backporting this is not so trivial, even to Kilted. It will have to be done manually as it has quite a few conflicts. It might not be possible at all because of API changes but someone would have to check.

PR #1161, opened 2026-01-09
ros2doctor: add node and parameter info to report

### LINKED ISSUE #1090: Add node and parameter information to doctor report.
### Description

Currently Node and Parameter information is just missing from the `ros2doctor report`.
Since those information is also important to know the ROS 2 system configuration, it would be nice to add those information as well as endpoints.

### Motivation

Providing more information for `ros2doctor report` can be useful to understand the entire ROS 2 network.

_No response_

### Additional Information

_No response_

BhuvanB404:
## Description

Add NodeCheck/NodeReport and ParameterCheck/ParameterReport so the ros2doctor report shows active nodes and their parameters.

- Add `ros2doctor/api/node.py` (node discovery, NodeCheck, NodeReport)
- Add `ros2doctor/api/parameter.py` (parameter discovery, ParameterCheck, ParameterReport)
- Update  entry points in `ros2doctor/setup.py`:
  - `ros2doctor.checks: NodeCheck` and `ParameterCheck`
  - `ros2doctor.report: NodeReport` and `ParameterReport`

Why:
This enhancement provides valuable insights into the active nodes and their parameters within a ROS 2 system. By including this information in the `ros2 doctor` report sections in the `ros2 doctor --report` output, giving users better visibility into the runtime ROS 2 system configuration.

Fixes: ros2/ros2cli#1090

Yes

### DIFF
--- a/ros2doctor/ros2doctor/api/node.py
+++ b/ros2doctor/ros2doctor/api/node.py
@@ -1,0 +1,69 @@
+# Copyright 2025 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from collections import Counter
+
+from ros2node.api import get_node_names
+from ros2cli.node.strategy import NodeStrategy
+from ros2doctor.api import DoctorCheck
+from ros2doctor.api import DoctorReport
+from ros2doctor.api import Report
+from ros2doctor.api import Result
+from ros2doctor.api.format import doctor_warn
+
+
+def find_duplicates(values):
+    """Return values that appear more than once."""
+    counts = Counter(values)
+    return [v for v, c in counts.items() if c > 1]
+
+
+class NodeCheck(DoctorCheck):
+    """Check for duplicate node names."""
+
+    def category(self):
+        return 'node'
+
+    def check(self):
+        result = Result()
+        with NodeStrategy(None) as node:
+            node_list = get_node_names(node=node, include_hidden_nodes=True)
+            node_names = [n.full_name for n in node_list]
+            duplicates = find_duplicates(node_names)
+            for duplicate in duplicates:
+                doctor_warn(f'Duplicate node name: {duplicate}')
+                result.add_warning()
+        return result
+
+
+class NodeReport(DoctorReport):
+    """Report node related information."""
+
+    def category(self):
+        return 'node'
+
+    def report(self):
+        report = Report('NODE LIST')
+        with NodeStrategy(None) as node:
+            node_list = get_node_names(node=node, include_hidden_nodes=True)
+            node_names = [n.full_name for n in node_list]
+            if not node_names:
+                report.add_to_report('node count', 0)
+                report.add_to_report('node', 'none')
+            else:
+                report.add_to_report('node count', len(node_names))
+                for node_name in sorted(node_names):
+                    report.add_to_report('node', node_name)
+        return report
--- a/ros2doctor/ros2doctor/api/parameter.py
+++ b/ros2doctor/ros2doctor/api/parameter.py
@@ -1,0 +1,104 @@
+# Copyright 2025 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import rclpy
+from rclpy.parameter import parameter_value_to_python
+from ros2param.api import call_get_parameters
+from ros2param.api import call_list_parameters
+
+from ros2cli.node.direct import DirectNode
+from ros2cli.node.strategy import NodeStrategy
+from ros2doctor.api import DoctorCheck
+from ros2doctor.api import DoctorReport
+from ros2doctor.api import Report
+from ros2doctor.api import Result
+from ros2doctor.api.format import doctor_warn
+
+
+class ParameterReport(DoctorReport):
+    """Report parameter related information."""
+
+    def category(self):
+        return 'parameter'
+
+    def report(self):
+        report = Report('PARAMETER LIST')
+        with NodeStrategy(None) as node:
+            try:
+                node_names_and_namespaces = \
+                    node.get_node_names_and_namespaces()
+            except Exception:
+                node_names_and_namespaces = []
+            if not node_names_and_namespaces:
+                report.add_to_report('total nodes checked', 0)
+                report.add_to_report('total parameter count', 0)
+                report.add_to_report('parameter', 'none')
+                return report
+
+            param_count = 0
+            nodes_checked = 0
+
+            with DirectNode(None) as param_node:
+                for node_name, namespace in sorted(node_names_and_namespaces):
+                    nodes_checked += 1
+                    full_name = f"{namespace.rstrip('/')}/{node_name}"
+                    try:
+                        response = call_list_parameters(
+                            node=param_node.node, node_name=full_name)
+                        if response is None:
+                            continue
+                        elif response.result() is None:
+                            continue
+
+                        param_names = response.result().result.names
+                        if param_names:
+                            param_count += len(param_names)
+                            report.add_to_report('node', full_name)
+                            try:
+                                param_response = call_get_parameters(
+                                    node=param_node.node,
+                                    node_name=full_name,
+                                    parameter_names=param_names
+                                )
+                                param_values = None
+                                if param_response:
+                                    param_values = param_response.values
+                                if param_values and len(param_values) == len(
+                                    param_names
+                                ):
+                                    params_with_values = sorted(
+                                        zip(param_names, param_values)
+                                    )
+                                    for name, value_msg in params_with_values:
+                                        value = parameter_value_to_python(
+                                            value_msg
+                                        )
+                                        report.add_to_report(
+                                            'parameter', f'{name}: {value}')
+                                else:
+                                    for param_name in sorted(param_names):
+                                        report.add_to_report(
+                                            'parameter', param_name
+                                        )
+                            except RuntimeError:
+                                for param_name in sorted(param_names):
+                                    report.add_to_report(
+                                        'parameter', param_name
+                                    )
+                    except RuntimeError:
+                        pass
+
+            report.add_to_report('total nodes checked', nodes_checked)
+            report.add_to_report('total parameter count', param_count)
+        return report
--- a/ros2doctor/test/fixtures/complex_node.py
+++ b/ros2doctor/test/fixtures/complex_node.py
@@ -1,0 +1,69 @@
+# Copyright 2026 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import rclpy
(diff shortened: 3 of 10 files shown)


BhuvanB404:
@fujitatomoya  this is the approach I have used with less dependencies outside of ros2cli. 

Do provide your suggestions and review

fujitatomoya:
there are some flake8 warnings https://build.ros2.org/job/Rpr__ros2cli__ubuntu_noble_amd64/346/testReport/ros2doctor.ros2doctor.test/test_flake8/test_flake8/, besides that we need to add the tests to make sure those sub-commands are working as expected.

BhuvanB404:
This is the output for `ros2 doctor --report` with  1 talker and listener running.

```
  NODE LIST
node count    : 3
node          : /_ros2cli_387960
node          : /listener
node          : /talker

   PACKAGE VERSIONS
ros2doctor                                : latest=0.40.4, local=0.40.4
ros_workspace                             : latest=1.0.3, local=1.0.3
ament_package                             : latest=0.18.2, local=0.18.1
ament_cmake_core                          : latest=2.8.4, local=2.8.4
ament_cmake_export_definitions            : latest=2.8.4, local=2.8.4
ament_cmake_libraries                     : latest=2.8.4, local=2.8.4
ament_cmake_export_dependencies           : latest=2.8.4, local=2.8.4
ament_cmake_export_include_directories    : latest=2.8.4, local=2.8.4
ament_cmake_export_libraries              : latest=2.8.4, local=2.8.4
ament_cmake_export_link_flags             : latest=2.8.4, local=2.8.4
ament_cmake_export_targets                : latest=2.8.4, local=2.8.4
ament_cmake_gen_version_h                 : latest=2.8.4, local=2.8.4
ament_cmake_python                        : latest=2.8.4, local=2.8.4
ament_cmake_include_directories           : latest=2.8.4, local=2.8.4
ament_cmake_target_dependencies           : latest=2.8.4, local=2.8.4
ament_cmake_test                          : latest=2.8.4, local=2.8.4
ament_cmake_version                       : latest=2.8.4, local=2.8.4
ament_cmake                               : latest=2.8.4, local=2.8.4
ament_cppcheck                            : latest=0.20.3, local=0.20.3
ament_cmake_cppcheck                      : latest=0.20.3, local=0.20.3
ament_cpplint                             : latest=0.20.3, local=0.20.3
ament_cmake_cpplint                       : latest=0.20.3, local=0.20.3
ament_lint                                : latest=0.20.3, local=0.20.3
ament_flake8                              : latest=0.20.3, local=0.20.3
ament_cmake_flake8                        : latest=0.20.3, local=0.20.3
ament_mypy                                : latest=0.20.3, local=0.20.3
ament_cmake_mypy                          : latest=0.20.3, local=0.20.3
ament_pep257                              : latest=0.20.3, local=0.20.3
ament_cmake_pep257                        : latest=0.20.3, local=0.20.3
uncrustify_vendor                         : latest=3.2.0, local=3.2.0
ament_uncrustify                          : latest=0.20.3, local=0.20.3
ament_cmake_uncrustify                    : latest=0.20.3, local=0.20.3
ament_index_python                        : latest=1.13.1, local=1.12.1
rcutils                                   : latest=7.0.5, local=7.0.4
rosidl_typesupport_interface              : latest=5.1.2, local=5.1.1
rosidl_runtime_c                          : latest=5.1.2, local=5.1.1
rosidl_dynamic_typesupport                : latest=0.4.0, local=0.4.0
rmw                                       : latest=7.9.1, local=7.9.1
rosidl_cli                                : latest=5.1.2, local=5.1.1
rosidl_adapter                            : latest=5.1.2, local=5.1.1
rosidl_parser                             : latest=5.1.2, local=5.1.1
rosidl_pycommon                           : latest=5.1.2, local=5.1.1
rosidl_cmake                              : latest=5.1.2, local=5.1.1
rosidl_generator_type_description         : latest=5.1.2, local=5.1.1
--
   PARAMETER LIST
node                     : /listener
parameter                : start_type_description_service
parameter                : use_sim_time
node                     : /talker
parameter                : qos_overrides./parameter_events.publisher.depth
parameter                : qos_overrides./parameter_events.publisher.durability
parameter                : qos_overrides./parameter_events.publisher.history
parameter                : qos_overrides./parameter_events.publisher.reliability
parameter                : start_type_description_service
parameter                : use_sim_time
total nodes checked      : 3
total parameter count    : 8

   PLATFORM INFORMATION
system           : Linux
platform info    : Linux-6.12.65-1-lts-x86_64-with-glibc2.39
release          : 6.12.65-1-lts
processor        : x86_64

   QOS COMPATIBILITY LIST
topic [type]            : /chatter [std_msgs/msg/String]
publisher node          : talker
subscriber node         : listener
compatibility status    : OK

   RMW MIDDLEWARE
middleware name    : rmw_fastrtps_cpp

   ROS 2 INFORMATION
distribution name      : rolling
distribution type      : ros2
distribution status    : rolling
release platforms      : {'debian': ['bookworm'], 'rhel': ['9'], 'ubuntu': ['noble']}

   SERVICE LIST ```

**Comment by BhuvanB404** [comment]
@fujitatomoya i have added tests for these. If any suggestions, do provide  : )

File: ros2doctor/ros2doctor/api/node.py [hunk 1]
```diff
@@ -0,0 +1,73 @@
...
+
+    def category(self):
+        return 'node'
+
+    def check(self):
+        result = Result()
+        with NodeStrategy(None) as node:
+            node_list = get_node_names(node=node, include_hidden_nodes=True)
+            node_names = [n.full_name for n in node_list]
+            if has_duplicates(node_names):
```
fujitatomoya: does this make more sense and useful?

```python
from collections import Counter
   
   def find_duplicates(values):
       """Return values that appear more than once."""
       counts = Counter(values)
       return [v for v, c in counts.items() if c > 1]
```

ros2doctor/ros2doctor/api/parameter.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+def call_list_parameters(node, node_name):
+    """Call the list_parameters service for a specific node using AsyncParameterClient."""
+    try:
+        client = AsyncParameterClient(node, node_name)
+        ready = client.wait_for_services(timeout_sec=2.0)
+        if not ready:
+            return None
+        future = client.list_parameters()
```
fujitatomoya: i am not sure if listing parameters are sufficient for ros2doctor.

maybe it should print the parameter values that are set as well? for example, `use_sim_time` parameter can be list but we are not able to see if that is set or not. i think for the ros2doctor would be useful to see the parameter variables for this case? what do you think?

ros2doctor/test/fixtures/complex_node.py
```diff
@@ -0,0 +1,67 @@
+# Copyright 2019 Open Source Robotics Foundation, Inc.
```
fujitatomoya: ??? did you copy this fixture from somewhere?

```suggestion
# Copyright 2026 Open Source Robotics Foundation, Inc.
```

fujitatomoya (commented):
overall lgtm with green CI. i got a few comments.

ros2doctor/ros2doctor/api/node.py
BhuvanB404: Ya , makes sense, removes my useless double check. Thank for the suggestion :)

ros2doctor/ros2doctor/api/parameter.py
BhuvanB404: providing state information for the parameters sounds more useful than current implementation . I'll add it :+1:

ros2doctor/test/fixtures/complex_node.py
BhuvanB404: For tests, I mainly referred to  `ros2node` and `ros2param` . Since the testing suite would be similar. So I copied the complex_node.py from there. It already had a launcher for  different node types, so i copied the fixtures files from there. If its not how it should , ill change it, let me know.

BhuvanB404:
@fujitatomoya sry for pinging again. I have pushed the changes, pls review.

BhuvanB404:
@fujitatomoya please take a look

BhuvanB404:
@fujitatomoya  . Sorry for pinging  again, could you take a look at this when you are free.

ros2doctor/ros2doctor/api/node.py
```diff
@@ -0,0 +1,86 @@
...
+    node_names_and_namespaces = node.get_node_names_and_namespaces()
+    return [
+        NodeName(
+            name=t[0],
+            namespace=t[1],
+            full_name=t[1] + ('' if t[1].endswith('/') else '/') + t[0],
+        )
+        for t in node_names_and_namespaces
+        if (include_hidden_nodes or (t[0] and not t[0].startswith(HIDDEN_NODE_PREFIX)))
+    ]
```
fujitatomoya: why do we not import this from `ros2node.api.get_node_names()`? i think this is pure duplicate.

ros2doctor/ros2doctor/api/parameter.py
```diff
@@ -0,0 +1,147 @@
...
+        rclpy.spin_until_future_complete(node, future, timeout_sec=2.0)
+        if future.done() and future.result() is not None:
+            # Return the parameter names list
+            return future.result().result.names
+        return None
+    except Exception:
+        return None
```
fujitatomoya: same here, can be imported from `from ros2param.api import call_list_parameters` to avoid duplicating this?

ros2doctor/ros2doctor/api/parameter.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+def call_get_parameters(node, node_name, parameter_names):
```
fujitatomoya: replaces with `from ros2param.api import call_get_parameters`?

ros2doctor/ros2doctor/api/parameter.py
```diff
@@ -0,0 +1,147 @@
...
+        if future.done() and future.result() is not None:
+            # Return the parameter values list
+            return future.result().values
+        return None
+    except Exception:
+        return None
+
+
+class ParameterCheck(DoctorCheck):
+    """Check for nodes without parameter services."""
```
fujitatomoya: why this can be the check for ros2doctor? i mean, user application can start the nodes without having the parameter service with `NodeOptions` by `start_parameter_services = false`, so this is expected configuration for the nodes. if we are expecting this check to confirm the parameter availability of the node, how about checking the `use_sim_time` that nodes always possess.

fujitatomoya (changes requested):

fujitatomoya:
@BhuvanB404 sorry for being late to get back to you. i have some more comments, can you check them?

ros2doctor/ros2doctor/api/node.py
BhuvanB404: I had it be imported at first, but then build was failing. I got stuck in the "ig need to decrease dependancies" mindset so copied them from it. I will fix it. thx

BhuvanB404:
@fujitatomoya  uses `use_sim_time` now . added a warnings for better debugging in case there are multiple nodes using different time variable. Thought would be better. But could be redundant if they just see them in the report itself. Refactered the functions as to better match coding style in other files. 

Now it checks and skips nodes without parameters  quietly . 

Any changes?

BhuvanB404:
@fujitatomoya , please review

ros2doctor/ros2doctor/api/parameter.py
```diff
@@ -0,0 +1,172 @@
...
+                doctor_warn('Inconsistent use_sim_time parameter detected:')
+                doctor_warn(
+                    '  Nodes with use_sim_time=True: '
+                    f'{", ".join(nodes_with_true)}'
+                )
+                doctor_warn(
+                    '  Nodes with use_sim_time=False: '
+                    f'{", ".join(nodes_with_false)}'
+                )
+                result.add_warning()
```
fujitatomoya: can you explain why this is warning for checking? i mean it can be expected that some nodes have use_sim_time true and some do not with false. (e.g user is doing some debugging on some nodes with simulation time, but other non-related nodes are running in system time?)

IMO, what it needs to check is that, to make sure all nodes have `use_sim_time` parameter no matter the value.

BhuvanB404: My thought process was that when multiple nodes are there with some real systems and some simulation systems based on presence and absence of the utilities, if their is a function that works based on timing of its node and other nodes, it would fail due to one having sim_time and the other real time. so i added this warning so that they can find root cause easier. Sry if that is not a general use case.

fujitatomoya: thanks for explanation. i think there would be the case that some nodes use simulation time, some do not. it is up to user configuration, at least this is not error or warning state.

i reconsidered what needs to be done here. how about listing the parameters from all nodes? that aligns with other doctor subcommand. maybe we do not need to get the parameter values to show, just listing all the available parameter of nodes?

BhuvanB404: @fujitatomoya . But wouldn't that reduce efficiency of the report? Like the param subcommand can show only list or values along with it based on needs. It has both options. How about adding a flag to show values if needed like a verbose mode?   like  `ros2 doctor --report-values/ -rv`

But if the goal is keeping report simple and act as  an overview, ig simplifying it is better. 
 For users who need parameter values, they can use the 
dedicated ros2 param commands.

Will update the PR shortly.

BhuvanB404:
I am unsure about removing values for parameters. It feels kind of incomplete somehow. Please suggest if values need to be removed

fujitatomoya:
> I am unsure about removing values for parameters. It feels kind of incomplete somehow. Please suggest if values need to be removed

i am okay with that, let's keep your suggestion here! we can keep the parameter values in the report. my concern was that if the ROS 2 network is huge like 300 nodes, this report is gonna take some time to list and get the parameter values.

PR #1162, opened 2026-01-14
Output node parameters upon each receipt

### LINKED ISSUE #1158: Better ros2cli UX for responsiveness and progress
### Description

For example, current `ros2 param list` collects all parameter list service responses from all nodes before printing any output. This causes the command to appear unresponsive when any node is slow to respond (or any of them are not responding), as the user sees no output until all nodes have replied.

the same behavior can be found on else where.

### Motivation

From a user experience perspective, seeing progressive output provides:

- Immediate feedback - Users know the command is working
- Partial results - Users can see available parameters while waiting for slow nodes
- Better debugging - Easier to identify which specific node is slow or unresponsive

This is especially important in large systems with many nodes, where one misbehaving node can make the entire command appear frozen.

- Sort the discovered node names alphabetically before iterating
- For each node (in sorted order), call the parameter list service and print the result immediately
- Remove the response caching mechanism

Trade-off: The current implementation sorts responses after all services have replied, which could result in different ordering if parameter names within a node were being sorted across all nodes. However, since parameters are printed per-node and sorted within each node, sorting nodes first and processing sequentially should produce identical output while improving responsiveness.

### Additional Information

_No response_

Barry-Xu-2018:
## Description

Improve the user experience of `ros2 param list`.   
As soon as the parameter information for a node is received, output it immediately instead of waiting to collect the parameters for all nodes before outputting them together.

Fixes #1158 

No

No

### DIFF
--- a/ros2param/ros2param/verb/list.py
+++ b/ros2param/ros2param/verb/list.py
@@ -75,7 +75,7 @@ class ListVerb(VerbExtension):
                     node=node,
                     node_name=node_name.full_name,
                     prefixes=args.param_prefixes)
-                # print responses
+                # print response
                 if response is None:
                     print(
                         'Wait for service timed out waiting for '
--- a/ros2param/ros2param/verb/list.py
+++ b/ros2param/ros2param/verb/list.py
@@ -68,15 +68,14 @@ class ListVerb(VerbExtension):
             regex_filter = re.compile(regex_filter[0])
 
         with DirectNode(args) as node:
-            responses = {}
+            # Sort node_names alphabetically
+            node_names = sorted(node_names, key=lambda n: n.full_name)
             for node_name in node_names:
-                responses[node_name] = call_list_parameters(
+                response = call_list_parameters(
                     node=node,
                     node_name=node_name.full_name,
                     prefixes=args.param_prefixes)
-            # print responses
-            for node_name in sorted(responses.keys()):
-                response = responses[node_name]
+                # print responses
                 if response is None:
                     print(
                         'Wait for service timed out waiting for '


fujitatomoya (approved):
lgtm with green CI.

Barry-Xu-2018:
Pulls: https://github.com/ros2/ros2cli/pull/1162
Gist: https://gist.githubusercontent.com/Barry-Xu-2018/b7bf6ed7353257611b68a0995631452b/raw/ddaacc46d02cdb80b2769297e8bee5257e4bffd0/ros2_rolling.repos
BUILD args: --packages-up-to ros2param
TEST args: --packages-select ros2param
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/17944
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/26937/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/20449/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/6692/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26404/)

ahcorde:
@Barry-Xu-2018 should we backport this ?

fujitatomoya:
i think this is better user experience but we need to see this is slight behavior change for `ros2 param list`. i would like to see this behavior is available in all downstream distros, what do you think?

Barry-Xu-2018:
> i think this is better user experience but we need to see this is slight behavior change for `ros2 param list`. i would like to see this behavior is available in all downstream distros, what do you think?

Yes. I also think so.

Barry-Xu-2018:
@ahcorde please backport to Kilted and Jazzy. (The Humble version’s code is incompatible with this PR, so it won’t be ported to this version.)

ahcorde:
https://github.com/Mergifyio backport kilted jazzy

PR #1170, opened 2026-01-19
Add per-node timeout option to ros2 param list

### LINKED ISSUE #1159: Add configurable timeout for service calls to prevent command from hanging indefinitely
### Description

For a instance, current `ros2 param list` can hang indefinitely when a node's parameter service is unresponsive.

https://github.com/ros2/ros2cli/blob/7a1ea4bc486b8e5d047762af2c3f59c79100f1ce/ros2param/ros2param/api/__init__.py#L112-L113

Note that wait_for_services() has a 5 second timeout for service discovery, but the actual service call via spin_until_future_complete() has no timeout. This means if a node's parameter service exists but is unresponsive (e.g., node is deadlocked, overloaded, or stuck), the command will block forever.

Proposed behavior would be to add the timeout argument option for the CLI, so that it can give up on unresponsive service to proceed the next request. at the end, this is going to be better user experience. i would say this could also be security risk.

### Motivation

- Predictable behavior: Users should expect the command to complete in a reasonable time
- Robustness: One misbehaving node should not block the entire command
- Debugging: Clear timeout messages help identify problematic nodes

Add `--timeout` command line argument with a reasonable default (e.g., 3.0 seconds), and pass it to the spin methods to wait. if that is timed out, handle timeout case gracefully with an informative message.

### Additional Information

_No response_

araitaiga:
## Description

This PR adds a `--per-node-timeout` option to the `ros2 param list` command.

Currently, when listing parameters from multiple nodes, the command can hang indefinitely if a node does not respond to the `list_parameters` service call. This change introduces a configurable timeout (default: 5.0 seconds) per node, allowing the command to skip unresponsive nodes and continue processing the remaining ones.

Fixes #1159

Yes.   
Users can use `--per-node-timeout <seconds>` option with `ros2 param list` to set how long to wait for each node's response. 

Example:
ros2 param list --per-node-timeout 2.0

Claude Opus 4.5

### Additional Information

- The `call_list_parameters()` function can be called with `spin_timeout_sec=None`, in which case it will wait indefinitely as before.

- The `call_list_parameters()` is also called from the `dump` verb and `ParameterNameCompleter`. In particular, `ParameterNameCompleter` may hang under the same conditions as `ros2 param list` when attempting tab completion during commands like `ros2 param get /node`. This may require a separate fix after further discussion.

### DIFF
--- a/ros2param/ros2param/api/__init__.py
+++ b/ros2param/ros2param/api/__init__.py
@@ -104,13 +104,13 @@ def call_set_parameters(*, node, node_name, parameters):
     return response
 
 
-def call_list_parameters(*, node, node_name, prefixes=None):
+def call_list_parameters(*, node, node_name, prefixes=None, spin_timeout_sec=None):
     client = AsyncParameterClient(node, node_name)
     ready = client.wait_for_services(timeout_sec=5.0)
     if not ready:
         return None
     future = client.list_parameters(prefixes=prefixes)
-    rclpy.spin_until_future_complete(node, future)
+    rclpy.spin_until_future_complete(node, future, timeout_sec=spin_timeout_sec)
     return future
 
 
--- a/ros2param/ros2param/verb/list.py
+++ b/ros2param/ros2param/verb/list.py
@@ -91,7 +91,8 @@ class ListVerb(VerbExtension):
                     # Future did not complete within timeout
                     print(
                         'Timed out waiting for list_parameters response '
-                        f'from node {node_name}')
+                        f'from node {node_name} '
+                        f'(timeout: {args.per_node_timeout}s)')
                     continue
                 elif response.result() is None:
                     e = response.exception()
--- a/ros2param/ros2param/verb/list.py
+++ b/ros2param/ros2param/verb/list.py
@@ -50,6 +50,11 @@ class ListVerb(VerbExtension):
         parser.add_argument(
             '--param-type', action='store_true',
             help='Print parameter types with parameter names')
+        parser.add_argument(
+            '--per-node-timeout', metavar='N', type=float, default=5.0,
+            help=(
+                'Maximum wait time per node for list_parameters service call '
+                'in seconds (default: %(default)s)'))
 
     def main(self, *, args):  # noqa: D102
         with NodeStrategy(args) as node:
@@ -74,12 +79,19 @@ class ListVerb(VerbExtension):
                 response = call_list_parameters(
                     node=node,
                     node_name=node_name.full_name,
-                    prefixes=args.param_prefixes)
+                    prefixes=args.param_prefixes,
+                    spin_timeout_sec=args.per_node_timeout)
                 # print response
                 if response is None:
                     print(
                         'Wait for service timed out waiting for '
                         f'parameter services for node {node_name}')
+                    continue
+                elif not response.done():
+                    # Future did not complete within timeout
+                    print(
+                        'Timed out waiting for list_parameters response '
+                        f'from node {node_name}')
                     continue
                 elif response.result() is None:
                     e = response.exception()
--- a/ros2param/test/fixtures/param_list_hang_node.py
+++ b/ros2param/test/fixtures/param_list_hang_node.py
@@ -1,0 +1,43 @@
+# Copyright 2026 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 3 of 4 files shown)


ros2param/ros2param/verb/list.py
```diff
@@ -50,6 +50,11 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
         parser.add_argument(
             '--param-type', action='store_true',
             help='Print parameter types with parameter names')
+        parser.add_argument(
+            '--per-node-timeout', metavar='N', type=float, default=5.0,
+            help=(
+                'Maximum wait time per node for list_parameters service call '
+                'in seconds (default: %(default)s)'))
```
fujitatomoya, line 57: @araitaiga 

```console
  --spin-time SPIN_TIME
                        Spin time in seconds to wait for discovery (only applies when not using an already running
                        daemon)
```

(as we talked in ROS JP Dev meeting) i was thinking we could use above parameter which already exists for the NodeStrategy. but according to the description, i do not think this is appropriate to be used for our use case here. sorry, i was wrong on this. probably having the dedicated parameter like this would be more reasonable.

ros2param/ros2param/api/__init__.py
```diff
@@ -104,13 +104,13 @@ def call_set_parameters(*, node, node_name, parameters):
     return response
 
 
-def call_list_parameters(*, node, node_name, prefixes=None):
+def call_list_parameters(*, node, node_name, prefixes=None, spin_timeout_sec=None):
```
fujitatomoya, line 107: i think you are trying to keep the default behavior for other use cases.
but this is something we need to consider to change? i mean, blocking forever on command line interfaces (any other command from ros2cli) is not good behavior, since it appears no response from CLI. and this is easy to happen in production grade distributed system with many remote endpoints. if any of nodes in ROS 2 network (could be user app, could be 3rd party app) is not responding, this situation can happen so easy.

i think default behavior is with timeout to give up. what do you think?

ros2param/ros2param/verb/list.py
```diff
@@ -74,13 +79,20 @@ def main(self, *, args):  # noqa: D102
...
                 if response is None:
                     print(
                         'Wait for service timed out waiting for '
                         f'parameter services for node {node_name}')
                     continue
+                elif not response.done():
+                    # Future did not complete within timeout
+                    print(
+                        'Timed out waiting for list_parameters response '
+                        f'from node {node_name}')
```
fujitatomoya: can we also print the timeout duration so that user can know that is not long enough for next time?

the other thing we could consider here is that, we should fail if any of nodes is responding? i think that depends on the use case if it should fail the command or ignore the non-responsive nodes?

araitaiga: > we should fail if any of nodes is responding? 

I agree that failing the entire command when any node fails would make exception handling easier in scripts. However, I believe the primary use case for param list is debugging, where users want to understand the system state. If the command stops at the first failure, users must fix those nodes before they can run param list again—even if those nodes are not relevant to them, which can be inconvenient.

Continuing past failures also allows users to identify all problematic nodes in a single run, which might be more useful for debugging.

ros2param/ros2param/api/__init__.py
araitaiga: Yes, I agree. Ideally, all applications using call_list_parameters should handle timeouts properly.
However, this requires significant changes and discussion, so I tried to keep the existing behavior in my initial PR. Currently, call_list_parameters is used by `ParameterNameCompleter` and `DumpVerb`, and adding proper timeout handling to both is somewhat complex.

For `ParameterNameCompleter`: It runs implicitly during tab completion, so adding a timeout option might confuse users. Using a fixed timeout (e.g., 5s) could break completion entirely in some environments.

For `DumpVerb`: Adding a timeout option seems reasonable, but it already has a --timeout option. We need to decide whether to reuse it, create a new one, and how to name it.

Despite these concerns, a 5-second timeout for call_list_parameters should be sufficient for most environments. Setting this as the default value could be a good approach.

fujitatomoya (approved):

fujitatomoya:
@Barry-Xu-2018 could you have a 2nd review on this?

fujitatomoya:
Pulls: ros2/ros2cli#1170
Gist: https://gist.githubusercontent.com/fujitatomoya/72681487d2c59025948504d3e0adcd78/raw/ee563dbf9245a41b6da5a11fbc7bfc2edc01c6f6/ros2.repos
BUILD args:  --packages-above-and-dependencies ros2param
TEST args:  --packages-above ros2param
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/18117
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/27536/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/20633/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/7409/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26590/)

Barry-Xu-2018 (approved):
LGTM.

fujitatomoya:
@araitaiga since this is new optional argument, we are not going to backport this to the downstream distros.

PR #1191, opened 2026-02-17
feat: fzf inline editing completion

### LINKED ISSUE #1184: cannot configure optional argument after fzf-based interactive selection
@tonynajjar 

i do not think this is issue, should be more like discussion. this just came up in ROS Japan Developer Meeting.
looks like this is constrain compared to tab completion.

```console
$ ros2 topic echo <enter>
```

goes in interactive selection mode, and once we pick the topic, it will execute the command right away.
but it is likely that user wants to add the command line argument after the select.

such as,

```console
$ ros2 topic echo /rosout --use-sim-time --clear ...
```

if we want to add any optional arguments with interactive selection mode, we need to make sure all the optional arguments set before <topic name>.

nice to have here is that, interactive selection mode just place the selected `<topic name>` in the command line, and lets the user to continue to assign the optional argument. and finally user is responsible to execute the command?

ArkhamKnight25:
@fujitatomoya san, Please review the changes. Thank you.

## Description

Added shell scripts (Bash + Zsh) that lets us pick topics, services and nodes with fzf, then drop us back on the command line so we can add flags before running. If you cancel fzf, it falls back to ros2cli's default picker. Should I keep this fallback or remove it?

Fixes #1184 

Yes, now we can easily hit up and down arrow keys to get the executed commands from the terminal.

Please let me know of any changes that are needed. It was interesting to learn about cli workflows. Thanks.

### DIFF
--- a/ros2cli/completion/ros2-fzf-completion.bash
+++ b/ros2cli/completion/ros2-fzf-completion.bash
@@ -1,0 +1,73 @@
+if [ -z "$BASH_VERSION" ]; then
+  return 0 2>/dev/null || exit 0
+fi
+
+ros2() {
+  local list_cmd=""
+  local fzf_prompt=""
+  local preview_cmd=""
+
+  case "$*" in
+    "topic echo" | "topic info" | "topic hz" | \
+    "topic bw"   | "topic pub"  | "topic type" | \
+    "topic delay")
+      list_cmd="ros2 topic list"
+      fzf_prompt="topic> "
+      preview_cmd="ros2 topic info {} 2>/dev/null"
+      ;;
+    "service call" | "service type" | "service find")
+      list_cmd="ros2 service list"
+      fzf_prompt="service> "
+      preview_cmd="ros2 service type {} 2>/dev/null"
+      ;;
+    "node info")
+      list_cmd="ros2 node list"
+      fzf_prompt="node> "
+      preview_cmd="ros2 node info {} 2>/dev/null"
+      ;;
+    "param get"  | "param set"  | "param list" | \
+    "param describe" | "param delete" | \
+    "param dump" | "param load")
+      list_cmd="ros2 node list"
+      fzf_prompt="node> "
+      preview_cmd="ros2 node info {} 2>/dev/null"
+      ;;
+    *)
+      command ros2 "$@"
+      return $?
+      ;;
+  esac
+
+  if ! command -v fzf &> /dev/null; then
+    echo "fzf is not installed. Install with: sudo apt install fzf" >&2
+    command ros2 "$@"
+    return $?
+  fi
+
+  local selected
+  selected=$(
+    command $list_cmd 2>/dev/null | fzf \
+      --height 40% --reverse --border \
+      --prompt "$fzf_prompt" \
+      --preview "$preview_cmd" \
+      --preview-window "right:50%:wrap"
+  )
+
+  if [ -z "$selected" ]; then
+    command ros2 "$@"
+    return $?
+  fi
+
+  printf '\e[A\r\e[2K'
+
+  local prefilled="ros2 $* ${selected} "
+  local full_cmd
+  read -e -p "${PS1@P}" -i "$prefilled" full_cmd
+
+  if [ -z "$full_cmd" ]; then
+    return 0
+  fi
+
+  history -s -- "$full_cmd"
+  eval -- "$full_cmd"
+}
--- a/ros2cli/completion/ros2-zle-completion.zsh
+++ b/ros2cli/completion/ros2-zle-completion.zsh
@@ -1,0 +1,87 @@
+if [ -z "$ZSH_VERSION" ]; then
+  return 0
+fi
+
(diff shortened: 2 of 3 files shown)


ros2cli/completion/ros2-fzf-completion.bash
```diff
@@ -0,0 +1,73 @@
...
+  local prefilled="ros2 $* ${selected} "
+  local full_cmd
+  read -e -p "${PS1@P}" -i "$prefilled" full_cmd
+
+  if [ -z "$full_cmd" ]; then
+    return 0
+  fi
+
+  history -s -- "$full_cmd"
+  eval -- "$full_cmd"
```
fujitatomoya, line 72: is this a security concern? the user is presented with a pre-filled prompt they can edit freely, and the result is passed to eval?

ros2cli/completion/ros2-fzf-completion.bash
```diff
@@ -0,0 +1,73 @@
...
+      ;;
+    *)
+      command ros2 "$@"
+      return $?
+      ;;
+  esac
+
+  if ! command -v fzf &> /dev/null; then
+    echo "fzf is not installed. Install with: sudo apt install fzf" >&2
+    command ros2 "$@"
```
fujitatomoya, line 43: it replaces the ros2 binary with a shell function for the entire session. that said, this will intercept all ros2 calls, including from other scripts. it changes behavior for every user who sources this file, not just in interactive completion contexts.

ros2cli/completion/ros2-fzf-completion.bash
```diff
@@ -0,0 +1,73 @@
...
+
+ros2() {
+  local list_cmd=""
+  local fzf_prompt=""
+  local preview_cmd=""
+
+  case "$*" in
+    "topic echo" | "topic info" | "topic hz" | \
+    "topic bw"   | "topic pub"  | "topic type" | \
+    "topic delay")
```
fujitatomoya, line 13: this matches "$*" (all args joined by space) against exact strings. i think there are some problems here,

- `ros2 topic  echo` (double space) does not.
- `ros2 --log-level debug topic echo` does not match.
- any future subcommands (e.g., ros2 topic inspect) won't be caught without updating these scripts.

fujitatomoya (commented):
@ArkhamKnight25 thanks for creating PR.

IMO, i am not inclined to take this approach, there could be many downsides with this PR because those shell scripts are not small shim... originally those shim scripts are meant to register ros2cli python application with our shell’s completion framework. (https://kislyuk.github.io/argcomplete/)

i believe the fundamental issue is that the scripts shadow the entire ros2 command with a shell function....

ros2cli/completion/ros2-fzf-completion.bash
ArkhamKnight25: you are right, using eval will introduce a security risk, allowing execution of additional commands, will not use it.

ros2cli/completion/ros2-fzf-completion.bash
ArkhamKnight25: yes, it is fragile, will remove the pattern matching.

ros2cli/completion/ros2-fzf-completion.bash
ArkhamKnight25: yes shadowing ros2() will indeed start intercepting everything

ArkhamKnight25:
@fujitatomoya Thank you for reviewing it. 
Okay so after hitting enter on `ros2 topic echo`, and then again selecting the method, to add the additional arguments,
Im thinking of fixing this with some python side changes then, the idea is to add an `interactive_select_with_edit()` helper in `helpers.py`. After the fzf selection, it would show a readline prompt pre-filled with the selected topic. The user could then type extra flags and hit enter, or just hit enter to run it.

So the flow would look something like this:

```
amrit@system25:~/ros2_ws$ ros2 topic echo
Select topic to echo:
> /chatter

```

And once the user selects it, it would look like this:

```
amrit@system25:~/ros2_ws$ ros2 topic echo
ros2 topic echo > /chatter <args>
```
Or 

can make it look better with just
```
amrit@system25:~/ros2_ws$ ros2 topic echo
> /chatter <args>
```
User can type the additional arguments after` ros2 topic echo > /chatter `or `> /chatter` and hit `enter` to execute.

Then I’d update the five topic verbs (echo, info, hz, bw, type). Each verb would store a reference to its parser in `add_arguments()` using `self._parser = parser`, use the new helper instead of `interactive_select`, and then reparse the user input with `self._parser.parse_args(tokens, namespace=args)` so that any extra flags get handled exactly the same way as if they were typed on the original command line.


And
* This will solve the problem of adding the arguments in the interactive, but the shell still will record only the `ros2 topic echo` in the history. How should I approach it?
* Is storing `self._parser = parser` in `add_arguments()` okay? and  using `shlex.split()` to tokenize the user input before re-parsing?



I did think of this approach earlier, but it won't store it in the history, so dropped the idea.

happy to adjust based on feedback.

PR #1185, opened 2026-02-05
add timeout arguments elsewhere

fujitatomoya:
## Description

Follow up of https://github.com/ros2/ros2cli/pull/1170, so that we can avoid the hang-up situation described in https://github.com/ros2/ros2cli/issues/1159

Fixes # (issue)

### Is this user-facing behavior change?

No, the additional timeout falls back to the current behavior.

### Did you use Generative AI?

Yes, Claude Sonnet 4.5

PR #1193, opened 2026-02-18
Implement ros2 param check_sim_time verb

### LINKED ISSUE #1192: ros2 param check_sim_time verb
### Description

I propose a new verb, `ros2 param check_sim_time`, that checks the `use_sim_time` value for every node currently available.

```
> ros2 param check_sim_time
use_sim_time:
/test/test_node1: True
/test/test_node2: False
/test/test_node3: <error: Exception while calling service of node /test/test_node3: None>
/test/test_node4: <not declared>
```

### Motivation

When working with real robots alongside simulation, it’s easy for a node to be launched with `use_sim_time` misconfigured, which can throw off timestamps across the system. Tracking down the culprit is often tedious.

_No response_

### Additional Information

_No response_

techtasie:
## Description

Fixes #1192

ros2 param check_sim_time is a new verb.
```
usage: ros2 param check_sim_time [-h] [--spin-time SPIN_TIME] [-s] [--no-daemon] [--include-hidden-nodes] [--per-node-timeout N]

Check for all node the use_sim_time parameter

options:
  --spin-time SPIN_TIME
  --include-hidden-nodes
                        Consider hidden nodes as well
  --per-node-timeout N  Maximum wait time per node for list_parameters service call in seconds (default: 5.0)
```

GPT-5.2 was used to format the output and help generate the test.

### DIFF
--- a/ros2param/ros2param/api/__init__.py
+++ b/ros2param/ros2param/api/__init__.py
@@ -74,14 +74,14 @@ def call_describe_parameters(*, node, node_name, parameter_names=None):
     return response
 
 
-def call_get_parameters(*, node, node_name, parameter_names):
+def call_get_parameters(*, node, node_name, parameter_names, spin_timeout_sec=None):
     client = AsyncParameterClient(node, node_name)
     ready = client.wait_for_services(timeout_sec=5.0)
     if not ready:
         raise RuntimeError('Wait for service timed out waiting for '
                            f'parameter services for node {node_name}')
     future = client.get_parameters(parameter_names)
-    rclpy.spin_until_future_complete(node, future)
+    rclpy.spin_until_future_complete(node, future, timeout_sec=spin_timeout_sec)
     response = future.result()
     if response is None:
         raise RuntimeError('Exception while calling service of node '
--- a/ros2param/ros2param/verb/check_sim_time.py
+++ b/ros2param/ros2param/verb/check_sim_time.py
@@ -1,0 +1,94 @@
+# Copyright 2026 Tim Wendt
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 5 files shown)


fujitatomoya (commented):
@techtasie thanks for bringing up the real use case and PR !

do we already have this in rolling source build? which can be applied to any parameter instead of `use_sim_time`, that is much better.

see also https://docs.ros.org/en/rolling/Tutorials/Beginner-CLI-Tools/Understanding-ROS2-Parameters/Understanding-ROS2-Parameters.html#ros2-param-get

```console
root@tomoyafujita-B760M-Pro-RS-D4:~/ros2_ws/colcon_ws# ros2 param get use_sim_time
/launch_ros_720881:
  Boolean value is: False
/lc_client:
  Boolean value is: False
/lc_listener:
  Boolean value is: False
/lc_talker:
  Boolean value is: False
```

besides, i think this PR is not good precedent. we might end up having the other verbs like `check_qos_override` and so on including test cases. i do not think this is gonna scale that way... this eventually becomes the maintenance burden...

techtasie:
@fujitatomoya Oh, I didn’t know about `ros2 param get use_sim_time`. Good to know!

But at least I found a bug in the current get-parameter helper:

```diff
--- a/ros2param/ros2param/api/__init__.py
+++ b/ros2param/ros2param/api/__init__.py
@@ -74,14 +74,14 @@ def call_describe_parameters(*, node, node_name, parameter_names=None):
     return response

-def call_get_parameters(*, node, node_name, parameter_names):
+def call_get_parameters(*, node, node_name, parameter_names, spin_timeout_sec=None):
     client = AsyncParameterClient(node, node_name)
     ready = client.wait_for_services(timeout_sec=5.0)
     if not ready:
         raise RuntimeError('Wait for service timed out waiting for '
                            f'parameter services for node {node_name}')
     future = client.get_parameters(parameter_names)
-    rclpy.spin_until_future_complete(node, future)
+    rclpy.spin_until_future_complete(node, future, timeout_sec=spin_timeout_sec)
     response = future.result()
     if response is None:
         raise RuntimeError('Exception while calling service of node '
```
There should be a timeout for the service request; otherwise, hung/unresponsive nodes can prevent this from completing, this is only a real problem if you want to request the config from multiple nodes.. I will write up a patch for that.

fujitatomoya:
> There should be a timeout for the service request; otherwise, hung/unresponsive nodes can prevent this from completing, this is only a real problem if you want to request the config from multiple nodes

that i am trying to address with https://github.com/ros2/ros2cli/pull/1185?

techtasie:
@fujitatomoya thanks!

Issue #1210, opened 2026-03-12
`ros2 action info` output is not consistent with `ros2 topic/service info`


fujitatomoya:
- `ros2 topic info`

```console
root@tomoyafujita-B760M-Pro-RS-D4:~/ros2_ws/colcon_ws# ros2 topic info /parameter_events
Type: rcl_interfaces/msg/ParameterEvent
Publisher count: 1
Subscription count: 0

root@tomoyafujita-B760M-Pro-RS-D4:~/ros2_ws/colcon_ws# ros2 topic info /parameter_events -v
Type: rcl_interfaces/msg/ParameterEvent

Publisher count: 1

Node name: _ros2cli_daemon_0_745ef2761fee4e5eadb327ec21e771c7
Node namespace: /
Topic type: rcl_interfaces/msg/ParameterEvent
Topic type hash: RIHS01_043e627780fcad87a22d225bc2a037361dba713fca6a6b9f4b869a5aa0393204
Endpoint type: PUBLISHER
GID: 01.0f.ac.7e.39.26.2e.73.00.00.00.00.00.00.04.03
QoS profile:
  Reliability: RELIABLE
  History (Depth): KEEP_LAST (1000)
  Durability: VOLATILE
  Lifespan: Infinite
  Deadline: Infinite
  Liveliness: AUTOMATIC
  Liveliness lease duration: Infinite

Subscription count: 0

```

- `ros2 service info`

```console
root@tomoyafujita-B760M-Pro-RS-D4:~/ros2_ws/colcon_ws# ros2 service info /fibonacci_action_server/get_type_description
Type: type_description_interfaces/srv/GetTypeDescription
Clients count: 0
Services count: 1

root@tomoyafujita-B760M-Pro-RS-D4:~/ros2_ws/colcon_ws# ros2 service info /fibonacci_action_server/get_type_description -v
Type: type_description_interfaces/srv/GetTypeDescription
Clients count: 0
Services count: 1
Node name: fibonacci_action_server
Node namespace: /
Service type: type_description_interfaces/srv/GetTypeDescription
Service type hash: RIHS01_69b9c19c1021405984cc60dbbb1edceb147a6538b411d812ba6afabeed962cd5
Endpoint type: SERVER
Endpoint count: 2
GIDs:
 - Request Reader : 01.0f.ac.7e.af.25.d7.58.00.00.00.00.00.00.11.04
[...]

PR #1220, opened 2026-04-18
Make ros2 action info --verbose consistent with topic/service info

catplotlib:
## Description

Adds a `--verbose` / `-v` option to `ros2 action info`, matching the output style already provided by `ros2 topic info -v` and `ros2 service info -v`.

Default output is unchanged. With `--verbose`, each action client / server is printed as a block containing:

- Node name
- Node namespace
- Action type
- Endpoint type (`CLIENT` / `SERVER`)

QoS and type hash are intentionally omitted since they are not exposed for action endpoints (per the linked issue).
Addresses the second TODO in [#1210](https://github.com/ros2/ros2cli/issues/1210). 

Yes. `ros2 action info` now accepts `--verbose` / `-v`. Previously this flag was rejected as an unknown argument.

Example (new behavior):

    $ ros2 action info -v /fibonacci
    Action: /fibonacci
    Action clients: 0
    Action servers: 1

    Node name: fibonacci_action_server
    Node namespace: /
    Action type: test_msgs/action/Fibonacci
    Endpoint type: SERVER

Without `-v`, output is identical to before.

Yes — Claude Opus 4.7

### Additional Information

- Added two CLI tests in `ros2action/test/test_cli.py`:
  - `test_fibonacci_info_verbose` — asserts the verbose block fields
  - `test_fibonacci_info_verbose_count` — asserts `-v -c` suppresses the per-endpoint block
- No changes to the default (non-verbose) output, so existing users / scripts are unaffected.
- `--verbose` combined with `--count` prints only the counts, matching the behavior of `ros2 topic info -v -c`.

### DIFF
--- a/ros2action/ros2action/verb/info.py
+++ b/ros2action/ros2action/verb/info.py
@@ -19,6 +19,25 @@ from ros2action.verb import VerbExtension
 from ros2cli.node.strategy import NodeStrategy
 
 
+def _split_fully_qualified_name(fqn):
+    if not fqn.startswith('/'):
+        return '/', fqn
+    idx = fqn.rfind('/')
+    if idx == 0:
+        return '/', fqn[1:]
+    return fqn[:idx], fqn[idx + 1:]
+
+
+def _format_action_endpoint_block(node_fqn, types, endpoint_label):
+    node_namespace, node_name = _split_fully_qualified_name(node_fqn)
+    return (
+        f'Node name: {node_name}\n'
+        f'Node namespace: {node_namespace}\n'
+        f'Action type: {", ".join(types)}\n'
+        f'Endpoint type: {endpoint_label}'
+    )
+
+
 class InfoVerb(VerbExtension):
     """Print information about an action."""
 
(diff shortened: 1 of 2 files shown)


catplotlib:
@fujitatomoya can i get a review on this please?

fujitatomoya (changes requested):
this PR does not provide consistent format without `--verbose` option to print the action type just like `ros2 topic info` and `ros2 service info` do. i think this is format breaking change, but it would be nice to be consistent over the endpoint types.

how about introducing `ActionEndpointInfo` and call the corresponding rcl interface to get those information and client libraries such as rclpy and rclcpp call this API, including ros2action via rclpy. this is more consistent with other endpoint implementation, and all the client libraries can take advantage of this API?

fujitatomoya:
@catplotlib are you still working on this?

fujitatomoya:
@catplotlib looks like you are not working on this anymore. i just opened the several PRs to replace this on https://github.com/ros2/ros2cli/issues/1210, it would be nice if you can review. i will go ahead to close this PR.

PR #1223, opened 2026-04-20
feat: add optional colorized output support

### LINKED ISSUE #1222: Add optional colorized output support
### Description

This issue adds optional colorized terminal output to `ros2cli`. Colorization is off by default and can be enabled in two ways:

- Set the environment variable `ROS_COLORIZED_OUTPUT=1` (requires stdout to be a TTY)
- Pass the `--color` flag on the command line (e.g. `ros2 --color topic echo /chatter`)

Image

### Motivation

`ros2cli` output — especially from commands like `ros2 topic echo`, `ros2 interface show`, and `ros2 param dump` — is often dense, structured text (YAML, IDL, key-value pairs). Without color, it is difficult to quickly distinguish field names from values, ROS type strings from plain text, or section headers from content. Colorized output is a standard ergonomic feature in modern CLI tools and has been a recurring community request.

**Modified: `ros2cli/cli.py`**
- Adds `--color` CLI argument
- Activates `ColoringStdout` wrapping when color is enabled, selecting the appropriate colorizer for the active command/verb

**New file: `ros2cli/color.py`**
- Introduces a `Colorizer` base class and an extensible registry keyed by `(command, verb)` pairs
- `DefaultColorizer` applies generic ANSI heuristics (ROS paths, type strings, YAML keys) as a fallback for unregistered commands
- `ColoringStdout` wraps `sys.stdout` and colorizes output line-by-line
- Built-in colorizers registered for: `ros2` (help), `topic echo`, `topic hz`, `topic bw`, `topic info`, `param get`, `param dump`, `interface show`, `component list`
- Third-party packages can register custom colorizers via `ros2cli.color.register(command, verb, colorizer)`

### Additional Information

_No response_

penww:
## Description

Fixes #1222 

Add ros2cli/color.py with an extensible colorizer framework that wraps sys.stdout and applies ANSI colors line-by-line. Colorizers are registered per (command, verb) pair; unregistered commands fall back to DefaultColorizer.

Colorized output is activated when ROS_COLORIZED_OUTPUT=1 and stdout is a TTY, or via the new --color CLI flag.

ros2color

No. 
- No behavioral change when `--color` is not passed and `ROS_COLORIZED_OUTPUT` is unset. 
- It only takes effect when stdout is a TTY, so it does not break pipe commands.

Yes, partially with Claude Sonnet 4.6

### Additional Information

### DIFF
--- a/ros2cli/ros2cli/cli.py
+++ b/ros2cli/ros2cli/cli.py
@@ -16,11 +16,15 @@
 import argparse
 import builtins
 import functools
+import os
 import signal
 import sys
 
 from rclpy.executors import ExternalShutdownException
 
+from ros2cli.color import ColoringStdout
+from ros2cli.color import get_colorizer
+from ros2cli.color import is_color_enabled
 from ros2cli.command import add_subparsers_on_demand
 
 
@@ -42,6 +46,11 @@ def main(*, script_name='ros2', argv=None, description=None, extension=None):
             'Do not force line buffering in stdout and instead use the python default buffering, '
             'which might be affected by PYTHONUNBUFFERED/-u and depends on whatever stdout is '
             'interactive or not'))
+    parser.add_argument(
+        '--color',
+        action='store_true',
+        default=False,
+        help='Force colorized output (also enabled by ROS_COLORIZED_OUTPUT=1)')
 
     # add arguments for command extension(s)
     if extension:
@@ -80,6 +89,15 @@ def main(*, script_name='ros2', argv=None, description=None, extension=None):
     if extension is None:
         # get extension identified by the passed command (if available)
         extension = getattr(args, selected_extension_key, None)
+
+    # activate colorized output if requested
+    if args.color:
+        os.environ['ROS_COLORIZED_OUTPUT'] = '1'
+    if is_color_enabled():
+        command_name = getattr(extension, 'NAME', '') if extension else ''
+        verb_name = getattr(getattr(args, '_verb', None), 'NAME', '') or ''
+        colorizer = get_colorizer(command_name, verb_name)
+        sys.stdout = ColoringStdout(sys.stdout, colorizer)
 
     # handle the case that no command was passed
     if extension is None:
--- a/ros2cli/ros2cli/color.py
+++ b/ros2cli/ros2cli/color.py
@@ -1,0 +1,364 @@
+# Copyright 2026 Peng Wang
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Colorized output support for ros2cli.
+
+Colorization is disabled by default.  It is enabled when the environment
+variable ROS_COLORIZED_OUTPUT=1 is set and stdout is a tty, or when the
+--color flag is passed on the command line.
+
+Architecture
+------------
+cli.py wraps sys.stdout with ColoringStdout before calling extension.main().
+ColoringStdout splits each write() call into lines and passes each non-empty
+line through a Colorizer.colorize() method before forwarding to the real
+stdout.
+
+The colorizer to use is looked up from a registry keyed by (command, verb)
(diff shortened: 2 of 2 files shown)


StefanFabian:
This parses the output from existing commands to colorize them. Why did you choose this over adding the color output directly to the commands?
In my opinion, this would be less fragile, since any change to the output formatting could break your colorizers, if I'm not mistaken.

kscottz:
@StefanFabian that's an astute observation. It feels like the right solution here is a standard base package that can be used by all of the CLI commands that then bubbles up to the final CLI. That's a big re-factor but probably worth it. It would make a fantastic student project for next year.

penww:
@StefanFabian @kscottz Thanks for your feedback. I fully agree with your observation that a colorizer is easy to break when output formatting changes. A more robust solution would be to update the output of all commands directly, but that would be a significant effort, as many commands currently rely on extensive use of raw print() and string concatenation.

https://github.com/ros2/ros2cli/blob/3b57dd3de6edbf3c7c5e04cffd34628a029b770d/ros2node/ros2node/verb/info.py#L78-L85

https://github.com/ros2/ros2cli/blob/3b57dd3de6edbf3c7c5e04cffd34628a029b770d/ros2node/ros2node/verb/info.py#L32-L33

I will continue tracking this feature and am willing to commit the effort required for a larger refactor if it becomes necessary.

fujitatomoya (commented):
I am not intend to take this PR as it is. i do agree with @StefanFabian 's comment, this is so easy to break because it couples a side-module to the exact text format of every command's output.
what's worse with this PR is, every output-format change becomes a silent breakage vector and the coupling is invisible from the command's side. this will be hard to maintain, i believe.

> A more robust solution would be to update the output of all commands directly, but that would be a significant effort, as many commands currently rely on extensive use of raw print() and string concatenation.

what i would do is, (as @kscottz  gestured at it) formatting/printing utility that individual commands opt into, so that color is produced at the source where the semantic meaning of each token is known such as a node name, a topic type, a warning and so on. this utility can be applied to each command and sub-command with token attribution. this seems to be straight-forward change for me but this would be much bigger change, we probably need to have more discussion on the basic design 1st.

kscottz:
This might be a good use case for Github's DRAFT PR feature. 

@penww you're more than welcome to drop by [Zulip / #ROS General](https://openrobotics.zulipchat.com/) to discuss more. This would be a heavy lift but it would also have super high impact and visibility.

StefanFabian:
Agree with @kscottz and @fujitatomoya, but with all the critical feedback on how this has to be implemented differently, which may be perceived as negativity, I want to remark that I also do think this is a great idea, and would be a huge improvement for the CLI tools.

fujitatomoya:
yeah, i believe this is good visual aid for all ros2cli users. just saying we could do this better to avoid maintainability issue.

penww:
Hi @fujitatomoya, fully understood — and thank you for confirming that this feature is valuable. I also agree that enabling it properly would require a fairly large refactor, which addresses my earlier concern about review complexity caused by large-scale changes.

Based on your suggestions, a model like `print( colorizer.green(node.name) )` (or similar APIs) makes much more sense. I can draft an initial version along these lines and bring it up for discussion.

@kscottz, I’ve opened a discussion thread on Zulip at [ROS General>ROS2 command line color support](https://openrobotics.zulipchat.com/#narrow/channel/526027-ROS-General/topic/ROS2.20command.20line.20color.20support/with/588578287). and I’ll post updates there as I have new progress.

penww:
Thanks for the discussion and guidance so far.

I opened #1248 to continue this work with a smaller utility-based approach: a shared color output utility, tests, and one command-level integration example.

One note about Lyrical: `Python/argparse` now provides **built-in** color support by default, but that only applies to argparse-generated help and usage text. It does **NOT** colorize actual ROS command output such as topic data, node info, interface definitions, or parameters.

image


So this still seems useful as a separate command-output colorization effort. Feedback on #1248 would be appreciated.

penww:
Following @kscottz 's suggestion, I will convert this PR to a draft PR.

kscottz:
Just for clarification here, you want us to review #1248? Can we close this PR in favor of #1248 or do you want to stage this as a series of PRs using this PR as part of that process?

This would be a great feature to get landed, but I want to make sure we understand your intention on how you want to stage the changes so we make sure we have someone ready to review it.

penww:
@kscottz Yes, this PR is no longer needed and can be closed.

I plan to use #1248 as the new foundation and incrementally add color support across the CLI commands.

At the moment, #1248 introduces the shared utility functions for colorized output and uses `ros2 topic info` as an example to demonstrate how these utilities can be integrated into a command. Please help to review it.

Once this approach is reviewed and accepted, I can start extending color support to additional commands. The work will be staged as a series of follow-up PRs, each covering a subset of commands, until color support is available more broadly across the CLI.

PR #1233, opened 2026-05-15
add ament_cmake_python build type to ros2 pkg create

darshmenon:
Closes #739

Adds `ament_cmake_python` as a valid `--build-type` option for `ros2 pkg create`, enabling users to scaffold a hybrid C++/Python package without having to manually write boilerplate.

**What changes**

`verb/create.py`
- Added `ament_cmake_python` to the `--build-type` choices
- Sets `buildtool_depends` to `['ament_cmake', 'ament_cmake_python']`
- Reuses `ament_cmake` test dependencies (lint tooling)
- Exports `build_type` as `ament_cmake` in `package.xml` (correct — colcon treats it as a CMake package)
- Calls `populate_ament_cmake_python` and creates a `scripts/` node when `--node-name` is given

`api/create.py`
- Added `populate_ament_cmake_python` — generates `CMakeLists.txt` and seeds the Python package directory with `__init__.py` and `py.typed`
- Extended `create_package_environment` to create the Python package source folder for `ament_cmake_python`

`resource/ament_cmake_python/CMakeLists.txt.em`
- New template: calls `find_package(ament_cmake_python)`, `ament_python_install_package(${PROJECT_NAME})`, and optionally installs a node script

**Usage**
```bash
ros2 pkg create my_pkg --build-type ament_cmake_python --node-name my_node --dependencies rclpy
```

Generates a package with:
- `CMakeLists.txt` using `ament_cmake_python`
- `my_pkg/__init__.py` Python package
- `scripts/my_node` executable node
- Correct `package.xml` with both `ament_cmake` and `ament_cmake_python` buildtool deps

PR #1234, opened 2026-05-15
fix(ros2pkg): replace deprecated importlib.resources.path with files/as_file

darshmenon:
## Problem

`importlib.resources.path()` has been deprecated since Python 3.11 and emits `DeprecationWarning` on Python 3.12, which is the interpreter used by ROS 2 Jazzy and Rolling. This means every `ros2 pkg create` invocation currently prints a deprecation warning on current Ubuntu 24.04 systems.

There is also a latent correctness bug: the existing code exits the `with resources.path(...)` context manager before using `template_path`, then calls `os.path.exists()` and `_expand_template()` on a path that may already be invalid. For resources packaged inside a zip (e.g. installed wheels), the temporary extracted file is deleted when the context manager exits, making the subsequent `os.path.exists()` check and `_expand_template()` call operate on a stale path.

## Fix

Replace `resources.path()` with `resources.files().joinpath()` + `resources.as_file()`, the API recommended since Python 3.9. Move all path-dependent calls inside the `as_file()` context manager so the extracted file remains valid for the duration of template expansion.

```python
# before
with resources.path(full_package, template_file_name) as path:
    template_path = str(path)
if not os.path.exists(template_path):          # path may be stale here
    raise FileNotFoundError(...)
_expand_template(template_path, ...)           # and here

# after
ref = resources.files(full_package).joinpath(template_file_name)
with resources.as_file(ref) as path:
    template_path = str(path)
    if not os.path.exists(template_path):      # path is valid inside the block
        raise FileNotFoundError(...)
    _expand_template(template_path, ...)       # and here
```

## Test Plan

- [ ] `ros2 pkg create --build-type ament_cmake my_pkg` — no `DeprecationWarning` on Python 3.12
- [ ] `ros2 pkg create --build-type ament_python my_pkg` — all template files generated correctly
- [ ] `ros2 pkg create --build-type ament_cmake --node-name my_node my_pkg` — node stub created correctly
- [ ] Verify no regression on Python 3.10 (Humble)

### DIFF
--- a/ros2pkg/ros2pkg/api/create.py
+++ b/ros2pkg/ros2pkg/api/create.py
@@ -83,15 +83,16 @@ def _create_template_file(
         template_subdir, template_file_name, output_directory, output_file_name, template_config
         ):
     full_package = 'ros2pkg.resource.' + template_subdir
-    with resources.path(full_package, template_file_name) as path:
+    ref = resources.files(full_package).joinpath(template_file_name)
+    with resources.as_file(ref) as path:
         template_path = str(path)
-    if not os.path.exists(template_path):
-        raise FileNotFoundError('template not found:', template_path)
+        if not os.path.exists(template_path):
+            raise FileNotFoundError('template not found:', template_path)
 
-    output_file_path = os.path.join(output_directory, output_file_name)
+        output_file_path = os.path.join(output_directory, output_file_name)
 
-    print('creating', output_file_path)
-    _expand_template(template_path, template_config, output_file_path)
+        print('creating', output_file_path)
+        _expand_template(template_path, template_config, output_file_path)
 
 
 def create_package_environment(package, destination_directory):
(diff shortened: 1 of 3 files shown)


fujitatomoya (commented):
same here, AI usage needs to be disclosed, please use our PR template.
besides, this PR includes unrelated difference from https://github.com/ros2/ros2cli/pull/1233. please include only code and changes related to the PR. do not include unrelated changes.

and DCO is missing.

christophebedard:
> importlib.resources.path() has been deprecated since Python 3.11 and emits DeprecationWarning on Python 3.12, which is the interpreter used by ROS 2 Jazzy and Rolling. This means every ros2 pkg create invocation currently prints a deprecation warning on current Ubuntu 24.04 systems.

>  Verify no regression on Python 3.10 (Humble)

This doesn't make sense. There's Kilted in between Jazzy and Rolling, and Rolling isn't on Ubuntu 24.04 anymore, it's on 26.04. And this targets Rolling, so I don't see why you'd need to check for regressions on Humble.

christophebedard:
> This means every ros2 pkg create invocation currently prints a deprecation warning on current Ubuntu 24.04 systems.

I sourced Jazzy and then ran `ros2 pkg create mypkg` and I don't see any `DeprecationWarning`. Can you provide a `ros2 pkg create` command that results in a warning?

PR #1263, opened 2026-07-29
ros2index new package

tomkimsour:
## Description

I have been working a lot with [ament_index](https://github.com/ament/ament_index 
) as well as my colleagues however there is a threshold where it becomes confusing what exist or not on the system. Therefore I wanted a tool to discover the ament ressources on the system.

The package is not doing much more than calling functions from ament_index_python and formatting it.

Yes, Claude Sonnet 5

### DIFF
--- a/ros2index/README.md
+++ b/ros2index/README.md
@@ -1,0 +1,51 @@
+# ros2index
+
+A `ros2cli` command to list and inspect entries in the [ament resource index](https://github.com/ament/ament_cmake/blob/master/ament_cmake_core/doc/resource_index.md) — the same index queried by `ament_index_python` — exposed as `ros2 index` verbs instead of a standalone script.
+
+## Usage
+
+```sh
+$ ros2 index --help
+usage: ros2 index [-h] ...
+
+Commands:
+  get    Output the content of a specific resource in the ament resource index
+  list   Output the resources of a given type in the ament resource index
+  types  Output the resource types in the ament resource index
+```
+
+### `types` — list resource types
+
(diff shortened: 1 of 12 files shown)


tomkimsour:
I just discovered ament_index already provide a cli entrypoint https://github.com/ament/ament_index/blob/rolling/ament_index_python/setup.py#L46

fujitatomoya (commented):
the motivation makes sense.
once we're juggling overlays and plugin registrations, the ament index is effectively invisible, and having a first-class way to inspect it would be genuinely useful.

i think there are two questions worth settling before the code review, and
they're really maintainer/governance questions rather than implementation ones:

how about considering the improvement of ros2pkg command?
there's real overlap in the most common use case: `ros2 index list packages` is
essentially `ros2 pkg list` plus a prefix column. e.g. `ros2 pkg list --prefixes`, since that's where users already look.
on the other hand, the generic index inspection (`types`, `list <type>` for things like
`rclcpp_components`, `get <type> <name>`) doesn't fit naturally under `ros2 pkg`, i think. the index isn't only about packages.

since a new `ros2 <command>` is effectively forever once it ships in a distro, it might
be worth a short ROS Discourse post laying out the motivation and the design options
(new command vs. `ros2 pkg` verbs), to get feedback beyond the people who happen to see this PR. an alternative path that has worked for other tools: release `ros2index` as a
standalone community package first, the plugin architecture means it works identically
from outside this repo and then revisit core inclusion once there's usage to point at.

Issue #44, opened 2017-08-21
Add CLI tool to get  a package's dependencies


jpsamper2009:
----
#### Feature description

It could be nice to have a tool to get the dependencies of a given package. Something like:
```
ros2 pkg depends <package name>
```
If we want to get fancier, we could also add options for different types of dependencies (build, exec, test, etc.).
And if we want to get even fancier, we could add an option to get sub-dependencies (essentially calling ros2 pkg depends recursively). This could be a separate feature request, though.

#### Implementation considerations

My first instinct would be to use the information encoded in the package.xml file. This would require two steps: 
1) Finding the package.xml for the desired package
  -  `ros2 pkg prefix <package_name>` is supposed to output the prefix path of a package. However, it seems to alway return the path to the install folder (not sure if this is a bug).
2) Parsing the file to extract the relevant information.
  - There's python libraries to parse xml files (e.g. xml.etree.ElementTree). I don't know ROS has a prefered library.
  - If we want to give options for the different types of dependencies, we would just have a filter during parsing.

mikaelarguedas:
It looks like the feature requested is very similar to:
`ament list_dependencies`

Would you mind trying it out and comment here if any of you use cases is not covered by it?

Thanks!

dirk-thomas:
I think having a command like `ros2 pkg depends <package name>` would be good. While `ament_tools` implements something similar a user might not even have that build tool installed. Also `ament_tools` is focusing on the packages in a workspace while this new command would cover all packages installed on a system.

> `ros2 pkg prefix <package_name>` is supposed to output the prefix path of a package. However, it seems to alway return the path to the install folder (not sure if this is a bug).

Returning the install prefix of the package is not a bug but intentional. It allows you to construct the path to any artifact provided by that package. By appending `share/<pkgname>/package.xml` you will get the path to the manifest.

You can call the same functionality through the Python API directly: https://github.com/ros2/ros2cli/blob/5137887fa95bab1323f81a3ae6b1cd670e1f7fbd/ros2pkg/ros2pkg/api/__init__.py#L32

> There's python libraries to parse xml files (e.g. xml.etree.ElementTree). I don't know ROS has a prefered library.
> If we want to give options for the different types of dependencies, we would just have a filter during parsing.

[...]

PR #1268, opened 2026-08-13
Add ros2 pkg depends verb

Miko997:
## Description

Add `ros2 pkg depends <package_name>` to print the package's direct declared dependencies.

The verb reads the installed `package.xml`, evaluates dependency conditions against the current environment, and prints unique dependency names in sorted order. Recursive lookup and dependency-type filters are outside this PR.

Addresses #44

Yes. `ros2 pkg` now provides a `depends` verb for installed packages.

### Additional Information

Validation:

- focused unit and CLI tests: 2 passed
- `ros2pkg` tests excluding unavailable `xmllint`: 6 passed, 1 deselected
- copyright, flake8, and pep257: 3 passed
- `git diff --check`

`xmllint` was not available in the test environment. This PR does not change XML files.

### DIFF
--- a/ros2pkg/ros2pkg/verb/depends.py
+++ b/ros2pkg/ros2pkg/verb/depends.py
@@ -1,0 +1,68 @@
+# Copyright 2026 Miko Parkkinen.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+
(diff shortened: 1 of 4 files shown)


fujitatomoya (commented):
what is the difference between this and `ros2 pkg xml | grep`? for maintainability, we do not want to maintain the overlapped features for sure.

besides, https://github.com/ros2/ros2cli/issues/44 is originally created almost 10 years ago. i am not sure if there are any requests from community... at least, it's been 10 years...

Timple:
Would the outcome be different than 
```
colcon list --names-only --packages-up-to <package_name>
```
?

fujitatomoya:
@Miko997 a couple of questions here before starting review, please have them checked out.

Miko997:
The outcome is different. `ros2 pkg xml | grep` returns raw XML lines, so it can include duplicates and dependencies whose conditions are false. `colcon list --packages-up-to` returns the recursive build set from a source workspace, including the target package.

This verb reads an installed package from the ament index and prints its unique direct dependencies after evaluating conditions. It does not require a source workspace or colcon.

I have not found newer demand beyond #44. If that installed-package use case does not justify another verb, I am happy to close the PR.

Timple:
Thanks for the clarification!

Issue #1077, opened 2025-07-16
A few of `ros2 doctor hello` enhancement.


fujitatomoya:
### Description

The following enhancement can be added to `ros2 doctor hello` command.

- include localhost connectivity check for the topic and multicast. currently it excludes the localhost (same hostname) connectivity check intentionally, but i do not find any reason to exclude this localhost connection check.
- UDP multicast group/port should be configurable via command line argument. Usually in practical environment, users would have specified multicast group and port number to be configured for their requirements, having those optional argument available, `ros2 doctor hello` can help the multicast connectivity check without code change.
- we could add service and action connectivity check aligned with topics.

### Motivation

motivation are described above.

### Design / Implementation Considerations

- localhost inclusion: we can simply remove the following if statement to include the same hostname information in the summary table.

```patch
diff --git a/ros2doctor/ros2doctor/verb/hello.py b/ros2doctor/ros2doctor/verb/hello.py
index 04fd666..cf5d483 100644
--- a/ros2doctor/ros2doctor/verb/hello.py
+++ b/ros2doctor/ros2doctor/verb/hello.py
@@ -160,8 +160,7 @@ class HelloSubscriber:
     def _callback(self, msg):
         msg_data = msg.data.split()
         pub_hostname = msg_data[-1]
-        if pub_hostname != socket.gethostname():
-            self._summary_table.increment_sub(pub_hostname)
+        self._summary_table.increment_sub(pub_hostname)


 class HelloMulticastUDPSender:
@@ -225,8 +224,7 @@ class HelloMulticastUDPReceiver:
                 data, _ = self._socket.recvfrom(4096)
                 data = data.decode('utf-8')
                 sender_hostname = data.split()[-1]
-                if sender_hostname != socket.gethostname():
-                    self._summary_table.increment_receive(sender_hostname)
+                self._summary_table.increment_receive(sender_hostname)
         except socket.timeout:
             pass

```

- Adding service and action check: we need to check if there is service server and action server is online in the ROS 2 network 1st. and if there isn't, we can start the service and action server on that host, otherwise skip creating servers because they will conflict. it would be probably nice to run the `MultiThreadedExecutor`, so that it can process the requests concurrently.

fujitatomoya:
CC: @InvincibleRMC

PR #1269, opened 2026-08-20
Include localhost in ros2 doctor hello connectivity checks

sylvesterkaczmarek:
## Summary

Addresses the localhost-connectivity part of #1077.

`ros2 doctor hello` currently excludes messages originating from the same hostname, so it cannot validate local topic delivery or multicast loopback. Count those messages as well so the command reports single-host connectivity.

The UDP receiver shutdown path now uses a per-instance socket pair instead of sending a wake-up datagram to the shared UDP port, so one `ros2 doctor hello` instance cannot wake or contaminate another instance on the same host. In `--once` mode, the command waits with a bounded timeout for the local topic and multicast round-trip instead of assuming both complete within one 0.1-second emit period.

The existing single-host CLI test now expects one local topic subscription and one local multicast receive.

## Testing

- Updated the existing single-host test expectations
- Branch is one signed commit and two files ahead of Rolling
- Full platform/RMW integration testing is left to project CI

### DIFF
--- a/ros2doctor/ros2doctor/verb/hello.py
+++ b/ros2doctor/ros2doctor/verb/hello.py
@@ -160,8 +160,7 @@ class HelloSubscriber:
     def _callback(self, msg):
         msg_data = msg.data.split()
         pub_hostname = msg_data[-1]
-        if pub_hostname != socket.gethostname():
-            self._summary_table.increment_sub(pub_hostname)
+        self._summary_table.increment_sub(pub_hostname)
 
 
 class HelloMulticastUDPSender:
@@ -223,10 +222,11 @@ class HelloMulticastUDPReceiver:
         try:
             while not self._is_shutdown:
                 data, _ = self._socket.recvfrom(4096)
+                if self._is_shutdown:
+                    break
                 data = data.decode('utf-8')
                 sender_hostname = data.split()[-1]
-                if sender_hostname != socket.gethostname():
-                    self._summary_table.increment_receive(sender_hostname)
+                self._summary_table.increment_receive(sender_hostname)
         except socket.timeout:
             pass
 
@@ -267,7 +267,7 @@ class SummaryTable:
             self._pub += 1
 
     def increment_sub(self, hostname):
-        """Increment subscribed msg count from different host(s)."""
+        """Increment subscribed msg count from host(s)."""
         with self.lock:
             if hostname not in self._sub:
                 self._sub[hostname] = 1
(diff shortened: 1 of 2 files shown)


ros2doctor/ros2doctor/verb/hello.py
```diff
@@ -223,10 +222,11 @@ def recv(self):
         try:
             while not self._is_shutdown:
                 data, _ = self._socket.recvfrom(4096)
+                if self._is_shutdown:
+                    break
```
fujitatomoya: how does it guarantee that is come from the same instance shutdown though? guessing this is flaky and can leak into other instances on the same host?

ros2doctor/test/test_cli.py
```diff
@@ -61,11 +63,12 @@ def test_hello_single_host(self):
...
         args.ttl = None
         args.once = True
-        with mock.patch('socket.gethostname', return_value='!nv@lid-n*de-n4me'):
+        hostname = '!nv@lid-n*de-n4me'
+        with mock.patch('socket.gethostname', return_value=hostname):
             summary = SummaryTable()
             hello_verb = HelloVerb()
             hello_verb.main(args=args, summary_table=summary)
-            expected_summary = _generate_expected_summary_table()
+            expected_summary = _generate_expected_summary_table(hostname)
```
fujitatomoya, line 71: with PR, the following assertion is only correct if two independent round-trips both complete within roughly that one 0.1-second window, i suggest that is tough and this is gonna be flaky test.

fujitatomoya (changes requested):
@sylvesterkaczmarek thanks for creating PR.

i assume you are using AI agent to create this PR, that is fine. but we have AI guideline: https://discourse.openrobotics.org/t/guidelines-for-ai-assisted-contributions-in-the-ros-project/55903 and please use our PR template to disclose the AI tools.

ros2doctor/ros2doctor/verb/hello.py
sylvesterkaczmarek: Good point. The shared UDP wake-up could be delivered to another receiver bound to the same port. I replaced it with a per-instance socket.socketpair() used only to wake that receiver thread, so shutdown signaling can no longer leak into another ros2 doctor hello instance.

ros2doctor/test/test_cli.py
sylvesterkaczmarek: Agreed. I removed the 0.1-second timing assumption. In --once mode the command now waits, with a bounded 5-second timeout, for both the local topic subscription and multicast receive before producing the summary.

fujitatomoya:
@sylvesterkaczmarek thanks for fixing up, out of curiosity what kind of AI models and tools are used to generate this code?

ahcorde:
> @sylvesterkaczmarek thanks for fixing up, out of curiosity what kind of AI models and tools are used to generate this code?

@sylvesterkaczmarek friendly ping

Issue #1214, opened 2026-03-23
topic delay performance overhead


DanBmh:
When trying to measure the topic delay of camera images for a real time setup, I found that `ros2 topic delay` is not very well suited for this:

- too few digits and metrics to measure millisecond delays
- using _python_ to calculate the delay introduces a notable overhead

This can be well seen if comparing it against a custom _c++_ tooling i've created for those reasons:

```bash
ros2 topic delay /camera02/pylon_ros2_camera_node/image_raw 
average delay: 0.014 	min: 0.014s max: 0.015s std dev: 0.00016s window: 131
average delay: 0.014	min: 0.014s max: 0.015s std dev: 0.00015s window: 181
average delay: 0.014	min: 0.014s max: 0.015s std dev: 0.00015s window: 231
average delay: 0.014	min: 0.014s max: 0.015s std dev: 0.00014s window: 281
average delay: 0.014	min: 0.014s max: 0.015s std dev: 0.00014s window: 331
```

```bash
ros2 run micro_delay micro_delay --ros-args -p topic:=/camera02/pylon_ros2_camera_node/image_raw
Avg: 0.012300s | Min: 0.012173s | P5: 0.012185s | P50: 0.012308s | P95: 0.012422s | Max: 0.012436s
Avg: 0.012321s | Min: 0.012186s | P5: 0.012194s | P50: 0.012323s | P95: 0.012513s | Max: 0.012531s
Avg: 0.012347s | Min: 0.012180s | P5: 0.012197s | P50: 0.012352s | P95: 0.012506s | Max: 0.012522s
Avg: 0.012358s | Min: 0.012172s | P5: 0.012194s | P50: 0.012374s | P95: 0.012473s | Max: 0.012477s
Avg: 0.012375s | Min: 0.012251s | P5: 0.012266s | P50: 0.012403s | P95: 0.012460s | Max: 0.012463s
```

The delay measured by my custom _c++_ tool is around _2ms_ less than the one reported by _ros2 topic delay_. I assume the main difference is caused by the better optimized programming language.

One important advantage of the _python_ implementation here is that it works with _arbitrary_ topics, while my _c++_ version only with _sensor_msgs::msg::Image_. So I can't provide an adequate replacement, but only inform about this possible performance problem.

The code for the `micro_delay` tool can be found here: https://gitlab.com/Percipiote/CamStreamer/-/blob/master/extras/delay/micro_delay/src/micro_delay.cpp

[...]

PR #1273, opened 2026-08-20
Make ros2 topic delay precision configurable

sylvesterkaczmarek:
## Summary

Addresses the output-precision part of #1214.

Add `--precision` to `ros2 topic delay` so users measuring millisecond-scale delays can request more decimal places for the average, minimum, and maximum values.

The default remains 3 decimal places, preserving existing output. Standard deviation keeps its existing 5-decimal formatting.

Example:

`ros2 topic delay /camera/image --precision 6`

## Testing

Added focused coverage for default and custom precision formatting.

### DIFF
--- a/ros2topic/ros2topic/verb/delay.py
+++ b/ros2topic/ros2topic/verb/delay.py
@@ -35,6 +35,7 @@ import rclpy
 
 from rclpy.time import Time
 
+from ros2cli.helpers import unsigned_int
 from ros2cli.node.direct import add_arguments as add_direct_node_arguments
 from ros2cli.node.direct import DirectNode
 from ros2cli.qos import add_qos_arguments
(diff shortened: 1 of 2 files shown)


fujitatomoya (commented):
the fix looks good as a cosmetic improvement, but #1214 stays open. adding precision length was never the original problem but the performance and accuracy of rclpy.

something i can think of is that we could get the clock from message info? so that we do not need to get the clock in the callback calculate the delay, possibly avoid the deserialization... but i have never tried, so just idea.

sylvesterkaczmarek:
> the fix looks good as a cosmetic improvement, but #1214 stays open. adding precision length was never the original problem but the performance and accuracy of rclpy.
> 
> something i can think of is that we could get the clock from message info? so that we do not need to get the clock in the callback calculate the delay, possibly avoid the deserialization... but i have never tried, so just idea.

Thanks, agreed. This PR is intentionally limited to the output-precision part of #1214, so #1214 should remain open for the underlying performance and accuracy issue.

Using the subscription `MessageInfo` receive timestamp is an interesting direction for that separate problem. It may avoid calling `node.get_clock().now()` for every message, although we'd need to check the clock semantics carefully and it would not by itself avoid deserializing the message because `topic delay` still needs the header timestamp.

Issue #1229, opened 2026-05-13
:farmer: ros2cli_daemon tests failing in win rel because it can't spawn daemons


miguelgonrod:
### Operating System:

Windows 11

### ROS version or commit hash:

Rolling

### Steps to reproduce issue

1. Go to https://ci.ros2.org/job/nightly_win_rel
2. Build a job
3. See test failing

### Expected behavior

Build without test regressions

### Actual behavior

Building with lots of test regressions because `assert spawn_daemon()` isn't spawning the daemon

### Reference Builds
- [nightly_win_rel#3794](https://ci.ros2.org/job/nightly_win_rel/3794)

### Failing tests
ros2cli.ros2cli.test.test_ros2cli_daemon.test_get_name
ros2cli.ros2cli.test.test_ros2cli_daemon.test_get_namespace
ros2cli.ros2cli.test.test_ros2cli_daemon.test_get_node_names_and_namespaces
ros2cli.ros2cli.test.test_ros2cli_daemon.test_get_node_names_and_namespaces_with_enclaves
ros2cli.ros2cli.test.test_ros2cli_daemon.test_get_topic_names_and_types
ros2cli.ros2cli.test.test_ros2cli_daemon.test_get_service_names_and_types
ros2cli.ros2cli.test.test_ros2cli_daemon.test_get_action_names_and_types
ros2cli.ros2cli.test.test_ros2cli_daemon.test_get_publisher_names_and_types_by_node
ros2cli.ros2cli.test.test_ros2cli_daemon.test_get_subscriber_names_and_types_by_node
ros2cli.ros2cli.test.test_ros2cli_daemon.test_get_service_names_and_types_by_node
ros2cli.ros2cli.test.test_ros2cli_daemon.test_get_client_names_and_types_by_node
ros2cli.ros2cli.test.test_ros2cli_daemon.test_get_action_server_names_and_types_by_node
ros2cli.ros2cli.test.test_ros2cli_daemon.test_get_action_client_names_and_types_by_node
ros2cli.ros2cli.test.test_ros2cli_daemon.test_get_publishers_info_by_topic
ros2cli.ros2cli.test.test_ros2cli_daemon.test_get_subscriptions_info_by_topic
ros2cli.ros2cli.test.test_ros2cli_daemon.test_get_clients_info_by_service
ros2cli.ros2cli.test.test_ros2cli_daemon.test_get_servers_info_by_service
[...]

PR #1274, opened 2026-08-20
Retry daemon socket acquisition after shutdown tail

sylvesterkaczmarek:
## Summary

Addresses the remaining failure mode in #1229.

`shutdown_daemon()` now waits for the XML-RPC address to become reusable, but a daemon that exits through its inactivity timeout can still stop serving before the process releases the listening socket. During that window, `spawn_daemon()` currently interprets `EADDRINUSE` as an already-running daemon and returns `False`.

When a spawn timeout is provided, distinguish that shutdown-tail case from a live daemon, wait for the address to become reusable, and retry the bind once.

Existing behavior is preserved when:
- a live daemon is actually running
- no spawn timeout was requested
- another process wins the bind race

## Testing

Added focused unit tests for:
- an address owned by a live daemon
- a temporarily busy shutdown-tail address
- existing non-blocking behavior without a timeout

### DIFF
--- a/ros2cli/ros2cli/node/daemon.py
+++ b/ros2cli/ros2cli/node/daemon.py
@@ -29,6 +29,12 @@ from ros2cli.helpers import wait_for
 from ros2cli.xmlrpc.client import ServerProxy
 
 
+# The daemon server polls for shutdown every 0.2 seconds. Allow several such
+# intervals for Windows process teardown without letting a foreign port owner
+# consume the caller's full (or indefinite) daemon startup timeout.
+_DAEMON_SOCKET_RELEASE_GRACE_PERIOD = 1.0
+
+
 class DaemonNode:
 
     def __init__(self, args):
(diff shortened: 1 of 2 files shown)


ros2cli/ros2cli/node/daemon.py
```diff
@@ -90,6 +90,32 @@ def _is_daemon_address_free():
...
+
+    # A live daemon owns the address legitimately. If no wait was requested,
+    # preserve the existing non-blocking behavior for any occupied address.
+    if is_daemon_running(args) or timeout is None:
+        return None
+
+    # On Windows a daemon can stop serving before its process releases the
+    # listening socket. Give that shutdown tail time to finish, then retry the
+    # bind. Another process may win the race, in which case it owns the socket.
+    if not wait_for(_is_daemon_address_free, timeout):
```
fujitatomoya: full-timeout stall (and possible infinite wait) when a foreign process holds the daemon port?
the daemon address is a fixed loopback port derived from ROS_DOMAIN_ID; any unrelated process can hold it.
if that happens, is_daemon_running() is false, and wait_for(_is_daemon_address_free, timeout) polls a predicate that can never become true.
the CLI blocks for the entire spawn timeout, then returns the same false the pre-PR code returned immediately though. this could be worse for UX, the module's documented timeout convention

fujitatomoya (changes requested):
@sylvesterkaczmarek did you actually try this PR with windows that reproduce the issue?

sylvesterkaczmarek:
> @sylvesterkaczmarek did you actually try this PR with windows that reproduce the issue?

Thanks. I addressed the port-owner concern by making the retry Windows-only and bounding it to a 1-second socket-release grace period rather than the caller's full or indefinite spawn timeout. I also added regression coverage for a foreign Windows port owner.

I have not reproduced the original failure on a Windows host myself, so I am not claiming that verification. I would still want Windows CI/build-farm validation before merge.

sylvesterkaczmarek:
The requested port-owner concern is addressed on the current head: the retry is Windows-only, bounded to a 1-second socket-release grace period, and there is regression coverage for a foreign process holding the daemon port. The original inline thread is now outdated and resolved. The PR is currently clean/mergeable and both DCO and Summary are green. Could you re-review the current head when convenient?

PR #1275, opened 2026-08-20
Support 64-bit integers in daemon XML-RPC

### LINKED ISSUE #629: ros2 daemon XML-RPC can't represent 64-bit integers
While investigating https://github.com/ros2/ros2cli/issues/603 , we found out that XML-RPC only supports 32-bit signed integers: http://xmlrpc.com/spec.md .  Since this is used as the communication mechanism between the ros2 daemon and clients, this could be a problem in the future.  We worked around that problem for the above issue by forcing the depth number in question to 0 (see https://github.com/ros2/rmw_cyclonedds/pull/305), but that won't always work.

The potential solutions are:
1. Switch to using strings for all values in XML-RPC.  We'd then have to marshal back-and-forth between integers and strings as needed.
2. Switch to a different protocol that allows more modern types.  The XML-RPC spec seems to have been last updated in 2003, so may not be up to the task anymore.

There may be other options here, but we should figure out a real solution.

sylvesterkaczmarek:
## Summary

Fixes #629.

Python's standard XML-RPC marshaller rejects integers outside the signed 32-bit range even though its unmarshaller supports the standard `i8` extension.

Register a custom integer marshaller that:
- preserves the normal `int` tag for signed 32-bit values
- uses `i8` for signed 64-bit values
- continues rejecting values outside signed 64-bit range

This allows daemon RPC payloads to carry 64-bit integer values without changing existing 32-bit output.

## Testing

Added focused round-trip coverage for:
- signed 32-bit values using `int`
- positive and negative values outside 32-bit using `i8`
- signed 64-bit boundaries
- rejection outside the signed 64-bit range

### DIFF
--- a/ros2cli/test/test_xmlrpc_int64.py
+++ b/ros2cli/test/test_xmlrpc_int64.py
@@ -12,12 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import threading
+from unittest.mock import patch
 from xmlrpc.client import dumps
 from xmlrpc.client import loads
 
 import pytest
 
+import ros2cli.daemon as daemon
+from ros2cli.node.daemon import DaemonNode
 import ros2cli.xmlrpc  # noqa: F401
+from ros2cli.xmlrpc.local_server import LocalXMLRPCServer
 
 
 @pytest.mark.parametrize('value', [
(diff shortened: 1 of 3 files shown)


ros2cli/test/test_xmlrpc_int64.py
```diff
@@ -0,0 +1,46 @@
+# Copyright 2026 Open Source Robotics Foundation, Inc.
```
fujitatomoya, line 1: i think we need an actual test with xmlrlc client via ros2daemon.

fujitatomoya (changes requested):
the implementation to patch the mashaller looks good to work, but i would take the other path to address the root cause such as JSON-RPC 2.0 over the existing loopback HTTP transport. at least, this needs to be tested.

ros2cli/test/test_xmlrpc_int64.py
sylvesterkaczmarek: Thanks. I added transport-level coverage using ros2daemon's own `make_xmlrpc_server()` together with `DaemonNode`, so the test now sends and receives a 64-bit integer through the actual HTTP/XML-RPC client/server path.

I kept this PR scoped to compatibility with the existing XML-RPC transport. Moving the daemon protocol to something like JSON-RPC would be a substantially larger architectural change and can be considered separately.

sylvesterkaczmarek:
The requested transport-level coverage is now present on the current head: the test uses ros2daemon's own XML-RPC server/client path and exercises a 64-bit value through the actual transport. The review thread is resolved, the PR is clean/mergeable, and DCO plus Summary are green. Could you re-review the current head when convenient?

Issue #696, opened 2022-02-25
Suggestions for Improved Ergonomics on ros2 action send_goal. 


kscottz:
## Bug report

**Required Info:**

- Operating System:
Ubuntu 20.04 
- Installation type:
Binaries
- Version or commit hash:
Galactic
- DDS implementation:
 All 
 Client library (if applicable):
 CLI

## Feature request
This is a general request to improve the action send_goal CLI ergonomics. I was trying to put together a TB4 tutorial and I spent a good hour wrestling with the CLI. In particular, I was trying to call an action with an empty message type for it's input. The short of it is that there is no way that a new user would be able to easily and correctly use the action send_goal command. 

#### Feature description
`ros2 action send_goal` takes in three parameters: the action, the action type, and the goal value to send. The empty message case is both particularly common and confusing. 

#### Implementation considerations
Here are my recommendations:

For the second parameter, the action type:
* The action type should be implicit in the action name. For most cases, the CLI should look up the action type for the user. E.g. `ros2 action send_goal /simple_thing '{x:1,y:2}'` should just work.  
* When the user specifies a goal type the CLI should be forgiving. Things like leading slashes, `.action` at the end of the parameter, and other modifications should be ignored. Copying most path variants into send_goal should just work.
* When the send_goal action fails, the CLI should output an example of what a correctly formed action type looks like from the action file.  E.g. `"my_ws/src/my_package/action/my_action.action => my_package/my_action"`
* The default case should be assumed to be an empty message. This would make it trivial to call simple action servers that require no input. 

For the third parameter (the goal value):
* If the user does not specify the third parameter the command should exit and output the goal type and an example of the goal message. E.g. `'{foo:3,bar:1}'`
* If the action goal is of type Empty, the user should not have to specify it. Make the common case fast E.g. `ros2 action send_goal /simple_thing`
* Bonus feature, allowing the goal to be input from a yaml file with a flag (e.g. `ros2 action send_goal /simple_thing -f ./goal.yaml`)

PR #1276, opened 2026-08-20
Add goal file input to ros2 action send_goal

sylvesterkaczmarek:
## Summary

Addresses the goal-file input part of #696.

Add `--goal-file FILE` to `ros2 action send_goal` so goal request values can be read from a YAML file instead of being provided inline or through standard input.

The new option participates in the same mutually exclusive input group as the existing positional goal and `--stdin`.

Example:

`ros2 action send_goal /fibonacci example_interfaces/action/Fibonacci --goal-file goal.yaml`

`-f` is already used by `--feedback`, so the file input uses the unambiguous long option `--goal-file`.

## Testing

Added unit coverage for runtime file reading, empty files, and read failures, plus an end-to-end launch test that sends a goal from --goal-file to the Fibonacci action server and verifies successful completion.

### DIFF
--- a/ros2action/ros2action/verb/send_goal.py
+++ b/ros2action/ros2action/verb/send_goal.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 from argparse import ArgumentTypeError
+from argparse import FileType
 
 import signal
 
@@ -61,6 +62,9 @@ class SendGoalVerb(VerbExtension):
         group.add_argument(
             '--stdin', action='store_true',
             help='Read goal from standard input')
+        group.add_argument(
+            '--goal-file', metavar='FILE', type=FileType('r'),
+            help='Read goal request values from a YAML file')
         arg.completer = ActionGoalPrototypeCompleter(action_type_key='action_type')
         parser.add_argument(
             '-f', '--feedback', action='store_true',
@@ -81,6 +85,9 @@ class SendGoalVerb(VerbExtension):
 
         if args.stdin:
             goal = collect_stdin()
+        elif args.goal_file:
+            with args.goal_file:
+                goal = args.goal_file.read()
         else:
             goal = args.goal
 
--- a/ros2action/ros2action/verb/send_goal.py
+++ b/ros2action/ros2action/verb/send_goal.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 from argparse import ArgumentTypeError
-from argparse import FileType
 
 import signal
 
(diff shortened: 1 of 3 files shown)


ros2action/ros2action/verb/send_goal.py
```diff
@@ -61,6 +62,9 @@ def add_arguments(self, parser, cli_name):
         group.add_argument(
             '--stdin', action='store_true',
             help='Read goal from standard input')
+        group.add_argument(
+            '--goal-file', metavar='FILE', type=FileType('r'),
```
fujitatomoya: i think argparse.FileType('r') is the wrong mechanism.
the file is opened at parse time and the open handle is stored in the args namespace, coupling resource lifetime to argument parsing.

ros2action/test/test_send_goal_file.py
```diff
@@ -0,0 +1,43 @@
+# Copyright 2026 Open Source Robotics Foundation, Inc.
```
fujitatomoya, line 1: coverage is a single mocked happy path, this is not good enough.
the actual test should be added to make sure the goal request from yaml file can be requested to the action server.

ros2action/ros2action/verb/send_goal.py
```diff
@@ -81,6 +85,9 @@ def main(self, *, args):
 
         if args.stdin:
             goal = collect_stdin()
+        elif args.goal_file:
+            with args.goal_file:
+                goal = args.goal_file.read()
```
fujitatomoya: could be None if the file is empty. i think error handling should be added.

fujitatomoya (changes requested):
the PR quality is not good enough, requesting change.

ros2action/ros2action/verb/send_goal.py
sylvesterkaczmarek: Agreed. I removed `argparse.FileType`. `--goal-file` now stores only the path, and the file is opened with a context manager when the command actually executes. I also added explicit handling and coverage for file-read failures.

ros2action/test/test_send_goal_file.py
sylvesterkaczmarek: Agreed. I added an actual CLI launch test using the existing Fibonacci action server. It runs `ros2 action send_goal ... --goal-file <file>` through the real command path and verifies that the file-supplied goal is accepted and finishes successfully.

ros2action/ros2action/verb/send_goal.py
sylvesterkaczmarek: Fixed. Empty or whitespace-only goal files are now rejected before `send_goal()` is called. I added coverage for that case, and also added coverage for a missing/unreadable goal file.

sylvesterkaczmarek:
All requested changes are addressed on the current head: `--goal-file` stores a path rather than an open handle, empty/read-error cases are handled, and an end-to-end CLI launch test sends a YAML-file goal to the Fibonacci action server. All review threads are resolved, the PR is clean/mergeable, and DCO plus Summary are green. Could you re-review the current head when convenient?

Issue #1093, opened 2025-08-28
🧑‍🌾 `ros2topic.ros2topic.test.test_bw_delay_hz.test_bw_delay_hz` failing in Kilted Zenoh


Crola1702:
### Operating System:

Ubuntu Noble

### ROS version or commit hash:

Kilted

### RMW implementation (if applicable):

Zenoh

### Steps to reproduce issue

1. Run a build in Kci Zenoh
2. See test regression fail

### Expected behavior

Test should not fail

### Actual behavior

Test is failing

### Additional information

Reference build:
- [Kci__nightly-zenoh_ubuntu_noble_amd64#35](http://build.ros2.org/job/Kci__nightly-zenoh_ubuntu_noble_amd64/35)

Test regression:
- [ros2topic.ros2topic.test.test_bw_delay_hz.test_bw_delay_hz](https://build.ros2.org/job/Kci__nightly-zenoh_ubuntu_noble_amd64/35/testReport/ros2topic.ros2topic.test/test_bw_delay_hz/test_bw_delay_hz/)

Log output:
```
FAIL: test_bw_delay_hz.TestROS2TopicBwDelayHz.test_bw_basic (topic='/clitest/topic/bw_compatible_qos', provide_qos=True, compatible_qos=True)
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Traceback (most recent call last):
  File "/tmp/ws/src/ros2/ros2cli/ros2topic/test/test_bw_delay_hz.py", line 170, in helper_verb_basic
    assert re.search(success_regex, command.output, flags=re.MULTILINE), (
AssertionError: bw CLI did not print expected message
assert None
 +  where None = <function search at 0x79acf8335c60>('^[0-9]+ B/s from [0-9]+ messages$', 'WARNING: topic [/clitest/topic/bw_compatible_qos] does not appear to be published yet\n', flags=re.MULTILINE)
 +    where <function search at 0x79acf8335c60> = re.search
 +    and   'WARNING: topic [/clitest/topic/bw_compatible_qos] does not appear to be published yet\n' = <launch_testing.tools.process.ProcessProxy object at 0x79ace8107110>.output
 +    and   re.MULTILINE = re.MULTILINE
```

[...]

PR #1098, opened 2025-09-04
Fujitatomoya/clearup isolated ros2daemon

fujitatomoya:
## Description

Fixes https://github.com/ros2/ros2cli/issues/1097

Possible fix for https://github.com/ros2/ros2cli/issues/1088, at least my local environment test results are all green and stable.

### Is this user-facing behavior change?

No, only test cleanup process is fixed.

### Did you use Generative AI?

No

### Additional Information

This PR includes https://github.com/ros2/ros2cli/pull/1096

PR #1277, opened 2026-08-20
Make topic CLI tests wait on command output

sylvesterkaczmarek:
## Description

Address the timing failure tracked in #1093 by waiting on the actual `ros2 topic bw`, `delay`, or `hz` process instead of using the publishing test node as a readiness signal.

The previous test called `self.node.count_publishers()` on the same node that had just created the publisher, so that check could succeed immediately without saying anything about what the fresh `DirectNode` inside the CLI process had discovered. My first revision tried to use the daemon as an independent observer; review correctly pointed out that this still did not prove the CLI node's state and also introduced an unwanted daemon lifecycle dependency.

The revised change removes `NodeStrategy` entirely. For the basic bw/delay/hz cases, the test now launches the real CLI command, keeps spinning the publisher executor, and waits up to 15 seconds for the command itself to produce the expected success output or incompatible-QoS warning. The wait remains bounded and fails if the command under test never reaches the expected state.

I do not have a local Zenoh reproduction result for this revised head, so I am not claiming that the build-farm failure is confirmed fixed. This change replaces the invalid readiness proxy with a condition observed from the command under test; Zenoh/build-farm validation is still needed.

Addresses #1093.

No. This only changes integration-test synchronization.

### Additional Information

The change is now limited to the basic bw/delay/hz path where #1093 reported the failure. The existing multi-topic and `--all` tests are left unchanged.

### DIFF
--- a/ros2topic/test/test_bw_delay_hz.py
+++ b/ros2topic/test/test_bw_delay_hz.py
@@ -42,6 +42,7 @@ from rclpy.qos import QoSCompatibility
 from rclpy.qos import QoSProfile
 from rclpy.qos import ReliabilityPolicy
 from rclpy.utilities import get_rmw_implementation_identifier
+from ros2cli.node.strategy import NodeStrategy
 
 
 # Skip cli tests on Windows while they exhibit pathological behavior
@@ -102,6 +103,15 @@ class TestROS2TopicBwDelayHz(unittest.TestCase):
         self.node.destroy_node()
         rclpy.shutdown(context=self.context)
 
+    def _wait_for_daemon_publishers(self, topics):
+        """Wait until the CLI daemon has discovered all publishers under test."""
+        with NodeStrategy(None) as node:
+            for _ in range(30):
+                self.executor.spin_once(timeout_sec=0.1)
+                if all(node.count_publishers(topic) > 0 for topic in topics):
+                    return
+        self.fail('Publishers were not discovered by the CLI daemon')
+
     def helper_verb_basic(self, launch_service, proc_info, proc_output, verb, success_regex):
         params = [
             (f'/clitest/topic/{verb}_basic', False, True),
@@ -156,15 +166,7 @@ class TestROS2TopicBwDelayHz(unittest.TestCase):
                     publisher.publish(msg)
 
                 publish_timer = self.node.create_timer(0.5, publish_message)
-
-                # Wait for the publisher to be discovered
-                publisher_count = 0
-                timeout_count = 0
-                while publisher_count == 0 and timeout_count < 10:
-                    self.executor.spin_once(timeout_sec=0.1)
-                    publisher_count = self.node.count_publishers(topic)
-                    timeout_count += 1
-                assert publisher_count > 0, 'Publisher was not discovered'
+                self._wait_for_daemon_publishers([topic])
 
                 try:
                     command_action = ExecuteProcess(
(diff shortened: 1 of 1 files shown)


fujitatomoya (commented):
> Fix the race tracked in https://github.com/ros2/ros2cli/issues/1093 by waiting until the CLI daemon has discovered the test publishers before launching ros2 topic bw, delay, or hz.

this does not really make sense to me. can you explain why this can address the issue?

sylvesterkaczmarek:
> > Fix the race tracked in #1093 by waiting until the CLI daemon has discovered the test publishers before launching ros2 topic bw, delay, or hz.
> 
> this does not really make sense to me. can you explain why this can address the issue?

Hi @fujitatomoya, you're right, the wording in the description is misleading imo. `bw`/`delay`/`hz` use a fresh `DirectNode`; the daemon is only being used here as an independent discovery observer.

The existing check calls `self.node.count_publishers()` on the same node that just created the publisher, so it can succeed before that endpoint has been discovered by any other participant. That's the race I was trying to remove: in #1093 the newly started CLI process reports that the topic does not appear to be published yet.

Waiting until the already-running isolated daemon can see the publisher gives us evidence that discovery has propagated beyond the publishing node before we launch the fresh CLI process.

If you'd prefer, I can also change the readiness probe to use a separate direct node instead of the daemon so it more closely mirrors the command under test. Thanks!

fujitatomoya:
> Waiting until the already-running isolated daemon can see the publisher gives us evidence that discovery has propagated beyond the publishing node before we launch the fresh CLI process.

`bw` / `delay` / `hz` run on `DirectNode` and never use the daemon intentionally. i think this guarantees nothing for this test. if you believe this reliably fixes the Zenoh failure, please share reproduction evidence and improvement result?

besides that (this can be even worse...), `NodeStrategy(None)` is a problem. `None` is not a valid args namespace, and if no daemon is running on the coordinated test domain, `NodeStrategy` will spawn one inside the test and leave it running after the test exits (2h inactivity timeout, bound to the domain port). we cleaned up stray isolated daemons before (#1098), this reintroduces the same problem into a test that previously did not depend on the daemon at all.

> If you'd prefer, I can also change the readiness probe to use a separate direct node instead of the daemon

that removes the daemon lifecycle problem but still does not close the race, another observer node is still not the CLI's node.

btw, are you pasting AI output here to answer review questions? your disclosure covers tests only. per https://discourse.openrobotics.org/t/guidelines-for-ai-assisted-contributions-in-the-ros-project/55903 you are accountable for understanding and explaining your contribution yourself. please answer the technical questions above in your own words, with reproduction data. (this will apply to all of your PRs now.)

thanks,

sylvesterkaczmarek:
> > Waiting until the already-running isolated daemon can see the publisher gives us evidence that discovery has propagated beyond the publishing node before we launch the fresh CLI process.
> 
> `bw` / `delay` / `hz` run on `DirectNode` and never use the daemon intentionally. i think this guarantees nothing for this test. if you believe this reliably fixes the Zenoh failure, please share reproduction evidence and improvement result?
> 
> besides that (this can be even worse...), `NodeStrategy(None)` is a problem. `None` is not a valid args namespace, and if no daemon is running on the coordinated test domain, `NodeStrategy` will spawn one inside the test and leave it running after the test exits (2h inactivity timeout, bound to the domain port). we cleaned up stray isolated daemons before (#1098), this reintroduces the same problem into a test that previously did not depend on the daemon at all.
> 
> > If you'd prefer, I can also change the readiness probe to use a separate direct node instead of the daemon
> 
> that removes the daemon lifecycle problem but still does not close the race, another observer node is still not the CLI's node.
> 
> btw, are you pasting AI output here to answer review questions? your disclosure covers tests only. per https://discourse.openrobotics.org/t/guidelines-for-ai-assisted-contributions-in-the-ros-project/55903 you are accountable for understanding and explaining your contribution yourself. please answer the technical questions above in your own words, with reproduction data. (this will apply to all of your PRs now.)
> 
> thanks,

Hi @fujitatomoya, you're right about the daemon approach. `NodeStrategy` was the wrong readiness signal here, and it also introduced an unnecessary daemon lifecycle dependency. So I've removed it.

The revised test now starts the actual `ros2 topic bw` / `delay` / `hz` command and keeps spinning the publisher while waiting for that command itself to produce the expected output. So the test no longer infers readiness from another observer node.

I don't currently have a local Zenoh environment where I can reproduce the original build-farm failure, so I can't honestly claim reproduction/improvement numbers yet. This revision fixes the synchronization problem in the test design, but Zenoh CI/build-farm results are still needed to confirm that it resolves #1093. 

Also, yes, I used AI assistance for the tests. I did and reviewed the technical changes myself, and I'm responsible for them. Thanks!

PR #1278, opened 2026-08-20
Support loading parameter files for all matching nodes

### LINKED ISSUE #778: No way to load a ros param file for all nodes
----
#### Feature description
ROS2 has no functionality for loading a parameter file that contains parameters for multiple nodes. This was easily doable in ROS1 and there is no reason it should not be possible in ROS2. For example, parameter file named "params.yaml" contains the following:

```/my_node1:
  ros__parameters:
    param1: 1.0
    param2: "test"
/my_node2:
  ros__parameters:
    param3: 73
```

and then running `ros2 param load params.yaml` would apply the relevant parameters to the nodes named `/my_node1` and `/my_node2`

sylvesterkaczmarek:
## Description

Allow `ros2 param load PARAMETER_FILE` to apply a parameter file to all currently running nodes matched by that file.

The existing `ros2 param load NODE_NAME PARAMETER_FILE` form is unchanged.

System-wide loading reuses `rclpy.parameter.parameter_dict_from_yaml_file()` per discovered node, so exact node entries, namespaces, and wildcard rules use the same matching behavior as normal parameter loading. The command also excludes its own temporary direct discovery node from wildcard targets.

Fixes #778.

Yes. `ros2 param load` can now omit the node name to load a multi-node parameter file across matching running nodes.

### Additional Information

Added focused coverage for the new one-positional CLI form, multi-node matching, unmatched nodes, and the no-match error path.

### DIFF
--- a/ros2param/ros2param/verb/load.py
+++ b/ros2param/ros2param/verb/load.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 from rclpy.parameter import parameter_dict_from_yaml_file
+from ros2cli.helpers import wait_for
 from ros2cli.node.direct import DirectNode
 from ros2cli.node.strategy import add_arguments
 from ros2cli.node.strategy import NodeStrategy
@@ -24,18 +25,31 @@ from ros2param.api import load_parameter_file
 from ros2param.verb import VerbExtension
 
 
-_NO_PARAMETERS_ERROR = 'Param file does not contain any valid parameters'
-
-
 def _parameter_file_matches_node(parameter_file, node_name, use_wildcard):
     try:
         parameter_dict_from_yaml_file(
             parameter_file, use_wildcard, target_nodes=[node_name])
-    except RuntimeError as e:
-        if str(e) == _NO_PARAMETERS_ERROR:
-            return False
-        raise
+    except RuntimeError:
+        # The complete file is validated before this helper is used. For a
+        # validated file, rclpy raises RuntimeError here when the selected node
+        # contributes no parameters.
+        return False
     return True
+
+
+def _get_matching_node_names(node, parameter_file, use_wildcard, include_hidden_nodes):
+    discovered_nodes = get_node_names(
+        node=node, include_hidden_nodes=include_hidden_nodes)
+    local_node_name = None
+    if node.daemon_node is None:
+        local_node_name = node.direct_node.get_fully_qualified_name()
+
+    return [
+        node_name
+        for node_name in sorted({n.full_name for n in discovered_nodes})
+        if node_name != local_node_name
+        and _parameter_file_matches_node(parameter_file, node_name, use_wildcard)
+    ]
 
 
 class LoadVerb(VerbExtension):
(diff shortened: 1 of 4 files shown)


ros2param/ros2param/verb/load.py
```diff
@@ -12,23 +12,39 @@
...
 
+_NO_PARAMETERS_ERROR = 'Param file does not contain any valid parameters'
+
+
+def _parameter_file_matches_node(parameter_file, node_name, use_wildcard):
+    try:
+        parameter_dict_from_yaml_file(
+            parameter_file, use_wildcard, target_nodes=[node_name])
+    except RuntimeError as e:
+        if str(e) == _NO_PARAMETERS_ERROR:
```
fujitatomoya: this is so fragile coupling to an rclpy error string. matching is detected by catching RuntimeError and comparing str(e) to the hardcoded 'Param file does not contain any valid parameters'

ros2param/ros2param/verb/load.py
```diff
@@ -48,12 +64,44 @@ def add_arguments(self, parser, cli_name):  # noqa: D102
...
+                load_parameter_file(
+                    node=node, node_name=node_name, parameter_file=args.parameter_file,
+                    use_wildcard=use_wildcard, timeout=args.service_timeout)
+            return
+
+        # Validate the file before discovery so malformed files fail consistently.
+        parameter_dict_from_yaml_file(args.parameter_file, use_wildcard)
+
         with NodeStrategy(args) as node:
-            if not wait_for_node(node, node_name, args.include_hidden_nodes, args.timeout):
```
fujitatomoya, line 53: --timeout semantics diverge between the two paths.

ros2param/test/test_load_all_nodes.py
```diff
@@ -0,0 +1,116 @@
+# Copyright 2026 Open Source Robotics Foundation, Inc.
```
fujitatomoya, line 1: can we add the actual test that issues the subcommand that this PR adds in practice here?

fujitatomoya (changes requested):

ros2param/ros2param/verb/load.py
sylvesterkaczmarek: Good point. Thanks a lot! I removed the error-string comparison. The file is now validated once before discovery, so per-node matching no longer depends on the text of rclpy's RuntimeError. I also changed the unit test to use a different error message to make sure that coupling doesn't come back.

ros2param/ros2param/verb/load.py
sylvesterkaczmarek: Agreed. The all-nodes path now uses the same underlying wait helper and respects `--timeout`, waiting up to that limit for at least one matching node before returning `No matching nodes found`.

ros2param/test/test_load_all_nodes.py
sylvesterkaczmarek: Yes. I added a launch test that runs `ros2 param load <parameter_file> --timeout 3` through the actual CLI, then reads the parameter back with `ros2 param get` to verify it was applied.

fujitatomoya:
DCO is missing.

ros2param/test/test_verb_load_all_nodes.py
```diff
@@ -0,0 +1,178 @@
...
+TEST_NODE = 'test_node'
+TEST_NAMESPACE = '/foo'
+TEST_TIMEOUT = 20.0
+
+
+if sys.platform.startswith('win'):
+    pytest.skip(
+        'CLI tests can block for a pathological amount of time on Windows.',
+        allow_module_level=True,
+    )
```
fujitatomoya: why?

sylvesterkaczmarek:
> DCO is missing.

Thanks. Fixed now. I added matching `Signed-off-by` trailers to all follow-up commits, and the DCO check is green on the current head.

ros2param/test/test_verb_load_all_nodes.py
sylvesterkaczmarek: You're right. There wasn't a repository-backed reason to exclude Windows here, so I removed the Windows-only module skip. The launch test now runs on all supported platforms.

sylvesterkaczmarek:
All requested changes are addressed on the current head: the fragile rclpy error-string coupling was removed, `--timeout` semantics were aligned, an actual CLI launch test was added, the Windows-only skip was removed, and DCO is green. The remaining outdated thread is now resolved. The PR is clean/mergeable with Summary green. Could you re-review the current head when convenient?

PR #1293, opened 2026-09-07
Match complete interface filenames in ros2 interface show

vpfkfl753:
## Description

After #1283, interface lookup uses `endswith('Foo.msg')`, so it can pick `OtherFoo.msg` before `Foo.msg`. It also accepts `Foo` when only `OtherFoo` exists.

Prefer the canonical `{type}/{name}.{type}` path, then fall back to a full filename match for nonstandard installation directories.

`ros2 interface show` displays the requested interface, or reports that it is missing, when another interface name has the same suffix.

Yes, OpenAI Codex (GPT-6), for the code change and tests.

### Additional Information

Added 12 regression cases using a temporary ament resource index. They cover messages, services and actions in standard and nested directories, with and without the requested interface. All 12 fail before the fix and pass in a ROS 2 Rolling container after it.

Added nine more cases for duplicate basenames, placing the canonical entry first, middle and last. Six fail before the follow-up change. All 21 lookup cases pass.

Built the package and its test interfaces with colcon. `colcon test --packages-select ros2interface` passes, including the existing CLI tests and lint checks.


ros2interface/ros2interface/verb/show.py
```diff
@@ -124,7 +124,7 @@ def _get_interface_lines(interface_identifier: str) -> typing.Iterable[Interface
     interfaces, _ = get_resource('rosidl_interfaces', pkg_name)
     interfaces = interfaces.splitlines()
 
-    interface = [f for f in interfaces if f.endswith(msg_name + '.' + msg_type)]
+    interface = [f for f in interfaces if os.path.basename(f) == msg_name + '.' + msg_type]
```
fujitatomoya: when more than one entry shares the basename (e.g. msg/Foo.msg and nested/msg/Foo.msg), interface[0] still picks silently by index order? probably we can consider preferring the canonical {type}/{name}.{type} path first?

vpfkfl753: You're right. e0dc20b now prefers `{type}/{name}.{type}`, with the basename match kept as a fallback for nonstandard paths. Added msg/srv/action tests with the canonical entry first, middle and last. Six cases fail before the change; all 21 lookup cases and the package's CLI/lint checks pass on Rolling.

kscottz:
@aarushjain355

